PR #7, opened 2024-04-11
Add Seasonal Statistic Example

dallasfoster:
# Earth2Studio Pull Request

## Description
This MR adds an example demonstrating how to define a custom statistic and run a seasonal simulation.

## Checklist

### DIFF
--- a/examples/06_seasonal_statistics.py
+++ b/examples/06_seasonal_statistics.py
@@ -165,10 +165,10 @@ def run_stats(
 #
 # We need the following:
 #
-# - Prognostic Model: Use the built in FourCastNet model :py:class:`earth2studio.models.px.FCN`.
+# - Prognostic Model: Use the built in Pangu 24 hour model :py:class:`earth2studio.models.px.Pangu24`.
 # - statistic: We define our own statistic: the Southern Oscillation Index (SOI).
 # - Datasource: Pull data from the GFS data api :py:class:`earth2studio.data.GFS`.
-# - IO Backend: Lets save the outputs into a Zarr store :py:class:`earth2studio.io.ZarrBackend`.
+# - IO Backend: Lets save the outputs into a NetCDF4 store :py:class:`earth2studio.io.NetCDF4Backend`.
 #
 # %%
 import numpy as np
@@ -228,7 +228,7 @@ class SOI:
         ds["date"] = dates
         ds = ds.set_index("date")
         ds = ds.drop(["Year", "Day", "SOI"], axis=1)
-        ds = ds.rolling(14, min_periods=1).mean().dropna()
+        ds = ds.rolling(30, min_periods=1).mean().dropna()
 
         self.climatological_means = torch.tensor(
             ds.groupby(ds.index.month).mean().to_numpy(), dtype=torch.float32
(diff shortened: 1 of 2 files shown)


examples/06_seasonal_statistics.py
```diff
@@ -0,0 +1,365 @@
...
+
+
+# %%
+# Set Up
+# ------
+# With the statistic workflow defined, we now need to create the indivdual components.
+#
+# We need the following:
+#
+# - Prognostic Model: Use the built in FourCastNet model :py:class:`earth2studio.models.px.FCN`.
```
NickGeneva: - Prognostic Model: Use the built in Pangu Weather 24 hour model :py:class:`earth2studio.models.px.Pangu24`.

examples/06_seasonal_statistics.py
```diff
@@ -0,0 +1,365 @@
...
+# Create the IO handler, store in memory
+io = NetCDF4Backend(
+    file_name="outputs/soi.nc",
+)
+
+# Create the custom statistic
+class SOI:
+    """Custom metric calculation the Southern Oscillation Index.
+
+    SOI = ( standardized_tahiti_slp - standardized_darwin_slp ) / soi_normalization
```
NickGeneva, line 199: Is there a difference between `slp` and `msl` in this instance?

dallasfoster: `slp` refers to sea level pressure, which is only defined at sea level, where `msl` refers to mean sea level pressure, which is interpolated in locations that are not at sea level. In this case, both Darwin and Tahiti are at sea level so `msl` should match `slp`.

examples/06_seasonal_statistics.py
```diff
@@ -0,0 +1,365 @@
...
+
+nsteps = 60
+nensemble = 1
+io = run_stats(["2022-01-01"], nsteps, nensemble, model, soi, data, io)
+
+
+# %%
+# Post Processing
+# ---------------
+# The last step is to post process our results. Cartopy is a greate library for plotting
```
NickGeneva, line 347: Just a line plot here

examples/06_seasonal_statistics.py
```diff
@@ -0,0 +1,365 @@
...
+
+        file_path = _download_cached(
+            "http://data.longpaddock.qld.gov.au/SeasonalClimateOutlook/SouthernOscillationIndex/SOIDataFiles/DailySOI1933-1992Base.txt"
+        )
+        ds = pd.read_csv(file_path, sep=r"\s+")
+        dates = pd.date_range("1999-01-01", freq="d", periods=len(ds))
+        ds["date"] = dates
+        ds = ds.set_index("date")
+        ds = ds.drop(["Year", "Day", "SOI"], axis=1)
+        ds = ds.rolling(14, min_periods=1).mean().dropna()
```
NickGeneva: Why rolling window of 14 here?

dallasfoster: The data source has noisy, daily SLP for each location but SOI is typically calculated on monthly climatologies. The window of 14 is a quasi-standard 2 week smoothing filter. I think the exact number itself is debatable and could be either 14 or 28.

PR #16, opened 2024-04-16
Add Ensemble Workflow

dallasfoster:
# Earth2Studio Pull Request

## Description
This MR implements an ensemble version of the `run.py:deterministic` workflow. The ensemble workflow supports a passed perturbation method and batching of ensemble members. 

This MR also updates examples 02, 04, and 05 to use this new ensemble workflow.
## Checklist

### DIFF
--- a/examples/03_corrdiff_inference.py
+++ b/examples/03_corrdiff_inference.py
@@ -16,14 +16,19 @@
 
 # %%
 """
-Generative Down Sampling
+Generative Downscaling
 ========================
 
-Generative down-sampling over Taiwan using CorrDiff diffusion model.
+Generative downscaling over Taiwan using CorrDiff diffusion model.
 
 This example will demonstrate how to user Nvidia's CorrDiff model, trained for
-predicting weather over Taiwan, to perform generative downsampling from quater degree
+predicting weather over Taiwan, to perform generative downscaling from quater degree
 global forcast data to ~3km.
+
+This checkpoint was trained on ERA5 data and WRF data that spans 2018-2021 at one hour
+time resolution. In this example, we demonstrate an application to GFS data for a typhoon
+super-resolution from 2023. The model's performance on GFS data and on data from this year
+has not been evaluated.
 
 In this example you will learn:
 


earth2studio/run.py
```diff
@@ -108,3 +111,129 @@ def deterministic(
...
+    )
+    logger.success(f"Fetched data from {data.__class__.__name__}")
+
+    # Set up IO backend with information from output_coords (if applicable).
+    total_coords = {"ensemble": np.arange(nensemble)} | coords0.copy()
+    total_coords["lead_time"] = np.asarray(
+        [prognostic.output_coords["lead_time"] * i for i in range(nsteps + 1)]
+    ).flatten()
+    for key, value in total_coords.items():
+        total_coords[key] = output_coords.get(key, value)
```
NickGeneva, line 185: should this be in an if statement `if key in output_coords`?

earth2studio/run.py
```diff
@@ -108,3 +111,129 @@ def deterministic(
...
+
+def ensemble(
+    time: list[str] | list[datetime] | list[np.datetime64],
+    nsteps: int,
+    nensemble: int,
+    prognostic: PrognosticModel,
+    data: DataSource,
+    io: IOBackend,
+    batch_size: Optional[int] = None,
+    perturbation_method: PerturbationMethod = Zero(),
```
NickGeneva: I would make this a required property for this workflow so its clear this is intended to be used with a perturbation method.

dallasfoster: Why do you think it should be a required property?

earth2studio/run.py
```diff
@@ -108,3 +111,129 @@ def deterministic(
...
+        Number of forecast steps
+    nensemble : int
+        Number of ensemble members to run inference for.
+    batch_size: int
+        Number of ensemble members to run in a single batch.
+    prognostic : PrognosticModel
+        Prognostic models
+    data : DataSource
+        Data source
+    io : IOBackend
```
NickGeneva, line 142: doc string needs updated params

earth2studio/run.py
dallasfoster: Is there a downside to the current approach aside from a modest performance hit?

earth2studio/run.py
```diff
@@ -108,3 +111,129 @@ def deterministic(
...
+def ensemble(
+    time: list[str] | list[datetime] | list[np.datetime64],
+    nsteps: int,
+    nensemble: int,
+    prognostic: PrognosticModel,
+    data: DataSource,
+    io: IOBackend,
+    batch_size: Optional[int] = None,
+    perturbation_method: PerturbationMethod = Zero(),
+    output_coords: CoordSystem = OrderedDict({}),
```
NickGeneva, line 125: Not against this, also not against making the built in workflows as simple as possible. I.e. you could just not allow users to subselect outputs but for ensemble, I see the need a lot more. Definitely optional.

earth2studio/run.py
```diff
@@ -108,3 +111,129 @@ def deterministic(
...
+
+        # Create prognostic iterator
+        model = prognostic.create_iterator(x, coords)
+
+        with tqdm(
+            total=nsteps + 1, desc=f"Running batch {batch_id} inference", leave=False
+        ) as pbar:
+            for step, (x, coords) in enumerate(model):
+                # Subselect domain/variables as indicated in output_coords
+                x, coords = map_coords(x, coords, output_coords)
```
NickGeneva, line 234: This function actually works surprisingly well here for controlling output, Nice.

examples/02_ensemble_workflow.py
```diff
@@ -202,13 +100,25 @@ def run_ensemble(
...
+    ["2024-01-01"],
+    nsteps,
+    nensemble,
+    model,
+    data,
+    io,
+    batch_size=batch_size,
+    perturbation_method=sg,
+    output_coords={"variable": np.array(["t2m", "tcwv"])},
+    device="cuda:0",
```
NickGeneva: I dont think we need to specify the device in the examples

NickGeneva: workflow should figure it out, I hope

earth2studio/run.py
dallasfoster: I think the motivation comes from wanting to reduce the amount of data that could possibly be generated. Since it is surprisingly easy to implement with `map_coords`, I think it is worth inclusion here.

examples/02_ensemble_workflow.py
```diff
@@ -202,13 +100,25 @@ def run_ensemble(
...
+io = ensemble(
+    ["2024-01-01"],
+    nsteps,
+    nensemble,
+    model,
+    data,
+    io,
+    batch_size=batch_size,
+    perturbation_method=sg,
+    output_coords={"variable": np.array(["t2m", "tcwv"])},
```
NickGeneva, line 119: Does the example explicitly state that we only want these two fields? I think some small mention of what this is getting added would be good.

examples/02_ensemble_workflow.py
dallasfoster: So the answer here is to allow the device selection occur under the hood?

examples/02_ensemble_workflow.py
dallasfoster: Good point, I will include.

earth2studio/run.py
NickGeneva: Because if theres not perturbation, one should be using the deterministic workflow right?

examples/02_ensemble_workflow.py
NickGeneva: It does do this already right with the device check?
https://github.com/NVIDIA/earth2studio/blob/dallasf/run_ensemble/earth2studio/run.py#L156

I think just let the example use the method under the hood is best to make it one less param

earth2studio/run.py
NickGeneva: makes sense to me

earth2studio/run.py
NickGeneva: Oh sorry I missed the default thing, nvm

examples/02_ensemble_workflow.py
```diff
@@ -223,51 +133,55 @@ def run_ensemble(
...
-    io[variable][1, 0, step],
-    f"{forecast} - Lead time: {6*step}hrs - Member: {1}",
-)
-plot_(
-    ax3,
-    np.std(io[variable][:, 0, step], axis=0),
-    f"{forecast} - Lead time: {6*step}hrs - Std",
-)
+for variable, cmap in zip(["t2m", "tcwv"], ["coolwarm", "Blues"]):
+    step = 0  # lead time = 24 hrs
```
NickGeneva: should this be 4?

NickGeneva: image
presently 0 hours

NickGeneva: Similar question for Single Variable Perturbation Method

earth2studio/run.py
dallasfoster: One can use model noise, of course. But for the sake of clarity to the user I will make this a required argument. The `model_perturbation_hook` example will show how to pass the `Zero` method.

examples/02_ensemble_workflow.py
dallasfoster: Fixed.

PR #33, opened 2024-05-03
Adding distributed manager example

NickGeneva:
# Earth2Studio Pull Request

## Description

Adding an example that demo's how to use the Distributed manager in modulus to help run inference workflows in parallel.

## Checklist


examples/08_distributed_manager.py
```diff
@@ -0,0 +1,220 @@
...
+# means that multiple jobs will download the same date-time if needed.
+
+# %%
+from typing import Any
+
+from earth2studio.utils.type import CoordSystem
+
+
+# Create the IO handler, just save two variables
+class CustomIO(ZarrBackend):
```
NickGeneva: Thoughts about this?

I think its an interesting solution but I'm not sure I entirely like it.

examples/08_distributed_manager.py
```diff
@@ -0,0 +1,220 @@
...
+
+chunks = {"time": 1}
+io = CustomIO(file_name=f"outputs/08_output_{dist.rank}.zarr", chunks=chunks)
+
+# %%
+# Earth2Studio does not provide distributed IO support. The recommendation is to always
+# output data for each process, then aggregate the data during post processing.
+#
+# In this example, only the total column water vapor and surface temperature are of
+# interest. We can easily filter out just the surface winds while using the built-in
```
NickGeneva: "surface winds" fix this

examples/.env
```diff
@@ -1,2 +1,6 @@
 LOGURU_LEVEL="INFO"
+RANK=0
```
NickGeneva, line 2: Presently example will just run serial, I checked it with multiple GPUs with mpirun manually. So spoof a  parallel environment here

examples/08_distributed_manager.py
dallasfoster: I do not believe we should support this type of user behavior. For this particular functionality I think it is preferable to ammend `deterministic` with the `ensemble` behavior of allowing passed `output_coords` and using `map_coords` internally.

examples/.env
dallasfoster: Can we also check manually with `torchrun`?

NickGeneva: Can you elaborate? I'm not trying add additional checks inside the script that would not be usually done so we dont mislead readers.

dallasfoster: Just for our own sanity do we know that this example runs with `mpirun` and `torchrun`? I have just seen occasional different behavior with the distributed manager and I would like to ensure that it would run under both conditions.

examples/08_distributed_manager.py
NickGeneva: Makes sense to me.

NickGeneva: Okay added the logic from the ensemble workflow to the others for output coords.
Also added dedicated tests for this functionality for all three.

examples/.env
NickGeneva: Yep! I checked mpi and torch run on the DGX box. Runs the same.

Added torch run command too and a comment about the .env file.

examples/08_distributed_manager.py
```diff
@@ -0,0 +1,190 @@
...
+times = np.array([f"200{i:d}-06-01T00:00:00" for i in range(0, 6)])
+assert (  # noqa: S101
+    len(times) > dist.world_size
+), "Inference runs should be greater than processes"
+time_shard = np.array_split(times, dist.world_size)[dist.rank]
+
+nsteps = 20
+output_coords = {"variable": np.array(["tcwv"])}
+io = run.deterministic(
+    time_shard, nsteps, model, data, io, output_coords, device=dist.device
```
dallasfoster: maybe just ensure that `output_coords = output_coords` for safety.

NickGeneva: Added

dallasfoster (approved):

PR #34, opened 2024-05-06
Perturbation API update

NickGeneva:
# Earth2Studio Pull Request

## Description

Updating the perturbations methods to now modify the tensor directly. Making the API the same as diagnostics.

- Update perturbation methods APIs
- Remove legacy perturbation class
- Updated ensemble workflow and examples
- Added userguide section plus some other doc updates

## Checklist


earth2studio/perturbation/bv.py
dallasfoster, line 1: This file needs to be updated. The perturbation method works by following the perturbation `dx`, which is now mislabeled by line 92. 

One solution would be to change lines 92 and 93 to the following:
```python
92: dx, coords = self.perturbation_method(x, coords)
93: dx -= x
...

File: examples/05_ensemble_workflow_extend.py [hunk 1]
```diff
@@ -97,20 +98,21 @@ def __call__(
...
-        # Find variable in data
+        # Apply perturbation on variable
+        x0, coords0 = map_coords(x, coords, {"variable": np.array(self.variable)})
+        x0, _ = self.pm(x0, coords0)
+        # Add perturbed slice back into original tensor
         ind = np.in1d(coords["variable"], self.variable)
-        dx[..., ~ind, :, :] = 0.0
-        return dx, coords
+        x[..., ind, :, :] = x0
+        return x, coords
```
dallasfoster: Consider that this solution will not work for the bred vector method, which would require passing the full tensor to the perturbation method.

dallasfoster (changes requested):

earth2studio/perturbation/bv.py
NickGeneva: Great catch

examples/05_ensemble_workflow_extend.py
NickGeneva: updated to just apply perturbation to whole input, thanks!

dallasfoster (approved):

PR #36, opened 2024-05-07
Adding Datasource to File Util

NickGeneva:
# Earth2Studio Pull Request

## Description

Adding a small util to export a datasource and coordinates from a model to a netCDF.
This is useful when one wants to create a offline netCDF from data source for a given model to use for inference multiple times since caching via fsspec can be a little on the slower side.

## Checklist


earth2studio/data/utils.py
```diff
@@ -124,3 +126,50 @@ def prep_data_inputs(
         time = timearray_to_datetime(time)
 
     return time, variable
+
+
+def datasource_to_netcdf(
```
dallasfoster: Can we make this more general so that users can save as Zarr too?

The main changes might be adding the following keyward arguments
```
...,
backend: Literal["netcdf", "zarr"] = "netcdf",
chunks: dict = {},
)
```

NickGeneva: Added!

docs/modules/datasources.rst
```diff
@@ -40,5 +40,6 @@ Functions
    :toctree: generated/data/
    :template: function.rst
 
+   data.datasource_to_netcdf
```
dallasfoster: This needs to be updated still?

dallasfoster (approved):

docs/modules/datasources.rst
NickGeneva: Yep. Good catch

PR #50, opened 2024-05-15
Add rank histogram

### LINKED ISSUE #49: 🐛[BUG]: ACC calculation uses wrong mean
### Version

v0.1.0

Pip, Source

### Describe the issue

The calculation of `x_bar` for ACC should subtract the mean of `x_hat` from `x_hat`, not the mean of `x`. Same for `y_bar`/`y_hat`/`y`. See [here](https://github.com/NVIDIA/earth2studio/blob/main/earth2studio/statistics/acc.py#L144) on lines 144 and 148.

dallasfoster:
# Earth2Studio Pull Request

## Description
This PR adds the rank histogram function.
This PR also fixes #49 
This PR makes `reduction_dimensions` as a property of statistics and metrics. This
is useful becomes some workflows need to keep track of how statistics/metrics manipulate
the input coordinates to produce output coordinates.
This PR fixes some typos in the documentation
This PR improves utility of __str__

## Checklist

### DIFF
--- a/earth2studio/statistics/base.py
+++ b/earth2studio/statistics/base.py
@@ -27,13 +27,16 @@ class Statistic(Protocol):
 
     @property
     def reduction_dimensions(self) -> list[str]:
+        """Gives the input dimensions of which the statistic performs a reduction
+        over. The is used to determine, a priori, the output dimensions of a statistic.
+        """
         pass
 
     def __call__(
         self, x: torch.Tensor, coords: CoordSystem
     ) -> tuple[torch.Tensor, CoordSystem]:
         """Apply statistic to data `x`, with coordinates `coords` and reduce
-        over dimensions `reduction_dims`.
+        over dimensions `reduction_dimensions`.
 
         Parameters
         ----------
@@ -41,7 +44,7 @@ class Statistic(Protocol):
             Input tensor intended to apply statistic to.
         coords : CoordSystem
             Ordered dict representing coordinate system that describes the tensor.
-            'reduction_dims' must be in coords.
+            `reduction_dimensions` must be in coords.
         """
         pass
 
@@ -62,7 +65,7 @@ class Metric(Protocol):
         y_coords: CoordSystem,
     ) -> tuple[torch.Tensor, CoordSystem]:
         """Apply metric to data `x` and `y`, checking that their coordinates
-        are broadcastable. While reducing over `reduction_dims`.
+        are broadcastable. While reducing over `reduction_dimensions`.
 
         Parameters
         ----------
@@ -70,11 +73,11 @@ class Metric(Protocol):
             Input tensor #1 intended to apply metric to.
         x_coords : CoordSystem
             Ordered dict representing coordinate system that describes the `x` tensor.
-            'reduction_dims' must be in coords.
+            `reduction_dimensions` must be in coords.
         y : torch.Tensor
             Input tensor #2 intended to apply statistic to.
         y_coords : CoordSystem
             Ordered dict representing coordinate system that describes the `y` tensor.
-            'reduction_dims' must be in coords.
+            `reduction_dimensions` must be in coords.
         """
         pass
--- a/earth2studio/statistics/base.py
+++ b/earth2studio/statistics/base.py
@@ -70,12 +70,14 @@ class Metric(Protocol):
         Parameters
         ----------
         x : torch.Tensor
-            Input tensor #1 intended to apply metric to.
+            Input tensor #1 intended to apply metric to. `x` is typically understood
+            to be the forecast or prediction tensor.
         x_coords : CoordSystem
             Ordered dict representing coordinate system that describes the `x` tensor.
             `reduction_dimensions` must be in coords.
         y : torch.Tensor
-            Input tensor #2 intended to apply statistic to.
+            Input tensor #2 intended to apply statistic to. `y` is typically the observation
+            or validation tensor.
         y_coords : CoordSystem
             Ordered dict representing coordinate system that describes the `y` tensor.
             `reduction_dimensions` must be in coords.
--- a/earth2studio/statistics/base.py
+++ b/earth2studio/statistics/base.py
@@ -25,6 +25,10 @@ from earth2studio.utils.type import CoordSystem
 class Statistic(Protocol):
     """Statistic interface."""
 
+    @property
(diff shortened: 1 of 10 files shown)


earth2studio/statistics/base.py
```diff
@@ -25,6 +25,10 @@
 class Statistic(Protocol):
     """Statistic interface."""
 
+    @property
+    def reduction_dimensions(self) -> list[str]:
```
NickGeneva, line 29: Nit: In the doc string below this is referred to as `reduction_dims`. Should update

earth2studio/statistics/rank.py
```diff
@@ -0,0 +1,157 @@
...
+
+import numpy as np
+import torch
+from modulus.metrics.general.histogram import _count_bins, linspace
+
+from earth2studio.utils.coords import handshake_coords, handshake_dim
+from earth2studio.utils.type import CoordSystem
+
+
+class rank_histogram:
```
NickGeneva, line 27: Add to statistics doc rst file

earth2studio/statistics/rank.py
```diff
@@ -0,0 +1,157 @@
...
+        y: torch.Tensor,
+        y_coords: CoordSystem,
+    ) -> tuple[torch.Tensor, CoordSystem]:
+        """
+        Apply metric to data `x` and `y`, checking that their coordinates
+        are broadcastable. While reducing over `reduction_dims`.
+
+        Parameters
+        ----------
+        x : torch.Tensor
```
NickGeneva, line 83: Nit: Are some better descriptions needed here? For example which one is the forecast vs the reference data?

earth2studio/statistics/rank.py
```diff
@@ -0,0 +1,157 @@
...
+        coord_count = 0
+        for c in x_coords:
+            if c != self.ensemble_dimension:
+                handshake_dim(y_coords, c, coord_count)
+                handshake_coords(y_coords, x_coords, c)
+                coord_count += 1
+
+        dim = list(x_coords).index(self.ensemble_dimension)
+
+        _ranks = torch.mean(1.0 * torch.ge(y, x), dim=dim)
```
NickGeneva, line 126: Bit confused by this here.

From my understanding, the ensemble elements should define the bin edges. The rank seems to be the average of the reduction dims not ensemble?

https://journals.ametsoc.org/view/journals/mwre/129/3/1520-0493_2001_129_0550_iorhfv_2.0.co_2.xml

earth2studio/statistics/rank.py
```diff
@@ -0,0 +1,157 @@
...
+        dims = [list(y_coords).index(rd) for rd in self._reduction_dimensions]
+        new_dims = list(range(len(dims)))
+        _ranks = torch.flatten(
+            torch.movedim(_ranks, dims, new_dims),
+            start_dim=new_dims[0],
+            end_dim=new_dims[-1],
+        )
+
+        # Compute histogram
+        _bin_edges = linspace(
```
NickGeneva, line 138: Similarly, bin edges should be the sorted values of the ensemble data. Does the above operation somehow transform this to bins between [0,1]?

earth2studio/statistics/rank.py
```diff
@@ -0,0 +1,158 @@
...
+            torch.zeros_like(_ranks[0]), torch.ones_like(_ranks[0]), self.number_of_bins
+        )
+        _rank_histogram = _count_bins(_ranks, _bin_edges)
```
NickGeneva, line 141: Whats the expected dims of this? My understanding is [n_ensemble, (input_dims-ensemble_dims)...]

dallasfoster: Let's assume that `ensemble_dimension` is `ensemble` and `reduction_dimensions` is `[lat, lon]`. Then `x` starts life as, for the sake of argument, `[ensemble, time, variable, lat, lon]` and `y` needs to be `[time, variable, lat, lon]`. The ranking procedure reduces `x` to the same coordinates as `y`. Then we take the reduction dimensions and merge them together to take the histogram, so the input to `_count_bins` is `[ latxlon, time, variable]` and we return `[data, time, variable]`.

earth2studio/statistics/rank.py
dallasfoster: In that paper they say, "A rank histogram is found by repeatedly tallying the rank of the truth relative to an actual distribution of sorted ensemble forecasts (which may or may not be calibrated)". Then we average over a large sample of statistically independent points.

So first, at a given point, we find the rank (between 0 and 1) of the observation with respect to the ensemble. Now we have reduced the ensemble dimension through the rank procedure (this is what occurs on line 125). Then, we take a histogram over all of the ranks over some set of statistically independent points (this is the reduction dimension).

earth2studio/statistics/rank.py
dallasfoster: Yes, the ranking process transforms the data into 0 - 1.

earth2studio/statistics/rank.py
NickGeneva: In this case is it assumed that the reduction points are statistically independent?

What is `data` size? (seems `self.number_of_bins`?)

NickGeneva: Oh I think its `bin` out from this function. Not data

test/statistics/test_ranks.py
```diff
@@ -0,0 +1,113 @@
...
+
+@pytest.mark.parametrize(
+    "ensemble_dimension",
+    [
+        "ensemble",
+        "time",
+    ],
+)
+@pytest.mark.parametrize("device", ["cpu", "cuda:0"])
+def test_rank_histogram(ensemble_dimension: str, device: str) -> None:
```
NickGeneva, line 35: how hard is it to make a test that checks the ranked histogram converges to a uniform distribution, say you make a bin count of 2 or 3. I imaging a few hundred samples would get you close?

earth2studio/statistics/rank.py
dallasfoster: I guess we collect the bin centers and bin counts into a coordinate called "data" (there is probably a better name for this) and the bins are in a coordinate called "bin".

test/statistics/test_ranks.py
dallasfoster: Done.

Issue #59, opened 2024-05-29
🐛[BUG]: SFNO caches into ~/.cache/modulus cache folder


NickGeneva:
### Version

0.1.0

### On which installation method(s) does this occur?

Source

### Describe the issue

See title

NickGeneva:
Related to: https://github.com/NVIDIA/earth2studio/issues/46

Issue #46, opened 2024-05-10
🚀[FEA]: Improve caching experience


NickGeneva:
### Is this a new feature, an improvement, or a change to existing functionality?

New Feature

### How would you describe the priority of this feature request

Critical (currently preventing usage)

### Please provide a clear description of problem you would like to solve.

Presently the model cache works fine, it operates in the .cache folder.

However all files are stored there, and because of this the file names are hashed to prevent conflicts.
This makes it hard to know what file is what and in some cases, like onnx checkpoints w/ separate weight files, can break things.

I think this can be improved by:

1. Make name hashing optional,  this is allowed in Fsspec and should be an option here
2. Make models default cache location its own subfolder
3. Models do not hash file names by default since we control the location, we can safely avoid conflicts

PR #67, opened 2024-06-10
Model Package Update to improve Model file caching

NickGeneva:
# Earth2Studio Pull Request

## Description

Refactoring of how the `earth2studio.models.px.auto.Package` class works to lean more into Fsspec to use more of their existing caching system as well as provide better downloading feedback for users via tqdm. Also adds better ability to customize caching directly in model, namely default can not hash file names to make cache readable. I made the call to decouple this from Modulus file system, which operates a little differently, because it takes too long to upstream.

**There is a slight "bug"/limitation in Fsspec which is the reason this took long to figure out related to the call back for caching**. Had to  extend the caching cache to fix.

Closes: https://github.com/NVIDIA/earth2studio/issues/46
Closes: https://github.com/NVIDIA/earth2studio/issues/59

## Checklist

### DIFF
--- a/earth2studio/models/auto.py
+++ b/earth2studio/models/auto.py
@@ -15,11 +15,262 @@
 # limitations under the License.
 
 import hashlib
+import io
 import os
+import re
 from typing import Any
 
+import fsspec
+import s3fs
+from fsspec.callbacks import Callback, TqdmCallback
+from fsspec.compression import compr
+from fsspec.core import BaseCache
+from fsspec.implementations.cached import LocalTempFile, WholeFileCacheFileSystem
+from fsspec.implementations.http import HTTPFileSystem
+from fsspec.spec import AbstractBufferedFile, AbstractFileSystem
+from fsspec.utils import infer_compression
 from huggingface_hub import HfFileSystem
+from loguru import logger
 from modulus.utils.filesystem import _download_cached
+from tqdm import tqdm
+
+# TODO: Make this package wide? Same as in run.py
+logger.remove()
+logger.add(lambda msg: tqdm.write(msg, end=""), colorize=True)
+
+
+class CallbackWholeFileCacheFileSystem(WholeFileCacheFileSystem):
+    """Extension of Fsspec WholeFileCacheFileSystem to include callback function when
+    downloading files to cache (progress bar).
+
+    See: https://github.com/fsspec/filesystem_spec/blob/8be9763e5f895073a9f46c8147aebbc64933e013/fsspec/implementations/cached.py#L651
+    """
+
+    def _open(self, path, mode="rb", **kwargs):  # type: ignore
+        path = self._strip_protocol(path)
+        if "r" not in mode:
+            hash = self._mapper(path)
+            fn = os.path.join(self.storage[-1], hash)
+            user_specified_kwargs = {
+                k: v
+                for k, v in kwargs.items()
+                # those kwargs were added by open(), we don't want them
+                if k not in ["autocommit", "block_size", "cache_options"]
+            }
+            return LocalTempFile(self, path, mode=mode, fn=fn, **user_specified_kwargs)
+        detail = self._check_file(path)
+        if detail:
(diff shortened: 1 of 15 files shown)


earth2studio/models/auto.py
```diff
@@ -14,85 +14,287 @@
...
+        """Child callback for recursive get"""
+        tqdm_kwargs = self._tqdm_kwargs
+        tqdm_kwargs["unit"] = "B"
+        tqdm_kwargs["unit_scale"] = True
+        tqdm_kwargs["unit_divisor"] = 1024
+        callback = TqdmCallback(
+            tqdm_kwargs=tqdm_kwargs,
+            tqdm_cls=self._tqdm_cls,
+        )
+        return callback
```
dallasfoster, line 128: This is going to sound very small but I think we can find a better place to put these utility classes.

earth2studio/models/auto.py
```diff
@@ -14,85 +14,287 @@
...
     Parameters
     ----------
     root : str
         Root directory for file system
-    cache_dir : str, optional
-        Cache directory, if remote path files will be downloaded here. If none is
-        provided then the path set in LOCAL_CACHE enviroment variable will be used.
-        If that is empty the path "~/.cache/earth2studio" is used, by default None
+    fs : AbstractFileSystem | None, optional
+        The target filesystem to run under neath. If none is provided one will get
```
dallasfoster: typo "under neath"

earth2studio/models/auto.py
```diff
@@ -14,85 +14,287 @@
...
+        initialized based on root url, by default None
```
dallasfoster: Can we be more specific about what gets initialized.

earth2studio/models/auto.py
```diff
@@ -14,85 +14,287 @@
...
+                self.root = f"https://api.ngc.nvidia.com/v2/models/{org}/{model}/versions/{version}/files/"
+            self.fs = HTTPFileSystem(block_size=2**10)
+        elif root.startswith("hf://"):
+            self.fs = HfFileSystem(target_options={"default_block_size": 2**20})
+        elif root.startswith("s3://"):
+            self.fs = s3fs.S3FileSystem(
+                anon=True,
+                client_kwargs={},
+                target_options={"default_block_size": 2**20},
+            )
```
dallasfoster, line 203: It has been my experience in the past that I have needed to pass `client_kwargs`. Not sure there is an elegant solution here, though.

dallasfoster (commented):
Looks beautiful.

earth2studio/models/auto.py
NickGeneva: Yeah this assumes that its a public bucket (options based on settings used for s3 based data sources).

I didn't want to have file system specific stuff in the constructor of the package, so my solution was to permit users to just provide a initialized filestore. So in the case of a private S3 you would provide fs in the constructor.

This is borrowed from how Fsspec deals with file stores with caching systems.

dallasfoster: Good point, I missed that initially.

earth2studio/models/auto.py
NickGeneva: I agree if this gets used in more spots. Would be up for moving it once another part of the code base uses it (like a file system util class). Similar to the onnx utils. I'm not sure this will be used in the data sources, with the current refactoring of them I'm even less sure...

earth2studio/models/auto.py
NickGeneva: Updated!

earth2studio/models/auto.py
NickGeneva: Thanks, fixed!

earth2studio/models/auto.py
NickGeneva: Honestly, I'm hoping a fix gets upstreams to fsspec and eventually the cache class can get removed.

dallasfoster (approved):

PR #92, opened 2024-07-08
ARCO Asyncio and Apache Beam Prep

NickGeneva:
# Earth2Studio Pull Request

## Description

Some prep work to refactor ARCO into a form that makes it better for Apache Beam pipelines as well as adding async IO to speed up. Some higher level design decisions made:

- The atomic function used inside data sources is now `fetch_array(time, variable) -> np.nd.array`. This allows this single line to be integrated into other pipelines instead of the asyncio one. This also encourages the building of the xarray object to live elsewhere.
- Number of async io tasks is limited by the property `async_process_limit`, this prevents people from hammering the API with 100s of requests at once. I used some code I found online and tested to act as a sort of un ordered map. I intentionally left this out of constructor because I don't want this too exposed to users because I do not want people increasing this a lot. Plus keeps the constructor clean.
- Placed the async code largely in the utils.py. I think multiple data sources can use this easily.
- There is a global time out for the entire data request controlled by the `async_timeout` limit. Its presently 100s. Similar to the previous parameter I didnt put this in the constructor to intentionally steer people away from using this to fetch massive in memory requests, rather make multiple data arrays.

Also:

- Remove tp06 from ARCO since its not officially in the data set, user WB2 instead. This was a temp fix to support the tp06 models.
 

## Checklist

### DIFF
--- a/earth2studio/data/arco.py
+++ b/earth2studio/data/arco.py
@@ -157,7 +157,7 @@ class ARCO:
         async def fetch_wrapper(
             e: tuple[datetime, int, str, int]
         ) -> tuple[int, int, np.ndarray]:
-            """Small wrapper to make"""
+            """Small wrapper that is awaitable for async generator"""
             return e[1], e[3], self.fetch_array(e[0], e[2])
 
         args = [
--- a/earth2studio/data/arco.py
+++ b/earth2studio/data/arco.py
@@ -14,10 +14,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import asyncio
 import os
 import pathlib
 import shutil
-from datetime import datetime, timedelta
+from datetime import datetime
 
 import fsspec
 import gcsfs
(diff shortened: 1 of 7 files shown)


earth2studio/data/arco.py
```diff
@@ -109,101 +112,115 @@ def __call__(
...
             shutil.rmtree(self.cache)
 
-        return xr.concat(data_arrays, dim="time")
+        return xr_array
 
-    def fetch_arco_dataarray(
-        self,
-        time: datetime,
-        variables: list[str],
+    async def create_data_array(
```
loliverhennigh, line 125: async code aways confuses me haha. Maybe I am missing something but does the create data array also need to be async. Could you not have the async part only be around the for loop for fetching the data.

loliverhennigh (commented):
Looks pretty good to me. Ill think about it over the weekend and no major comment. Bit nieve on python async though. I will have more free time next week and can help get all the other data sources using this :)

earth2studio/data/arco.py
NickGeneva: The trouble with async is that you cant really "await" inside a non-async function. The async generator here needs an await inside of it. So I find it easiest to rather enter entire function into async rather than do it through the loop. Rather than having a pool of tasks you have to collect, you can just have one async entry point here.

Perhaps there is a more parsimonious way to do this, but with the async generator its hard do something that looks more like a threaded solution. I would need to test things but suspect you would need another wrapper function inside this if you removed the async from the create data array

Note that the important function `fetch_array` isnt async in this case.

NickGeneva: Also this code is very confusing, because it has the throttling control to limit number of async processes. Its by no means a "strandard" async usage you would find else where.

PR #93, opened 2024-07-10
Add tensor noise amplitude parameters for perturbation methods

dallasfoster:
# Earth2Studio Pull Request

## Description
This is a backward compatible PR that adds the ability to pass broadcastable tensors as noise amplitudes in relevant perturbation methods.

## Checklist

### DIFF
--- a/earth2studio/models/dx/identity.py
+++ b/earth2studio/models/dx/identity.py
@@ -58,7 +58,6 @@ class Identity(torch.nn.Module):
         CoordSystem
             Coordinate system dictionary
         """
-
         return input_coords.copy()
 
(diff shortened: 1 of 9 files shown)


dallasfoster:
> Currently there's not mechanism to ensure the noise vector is on the same device as the input tensor.
> 
> I would maybe suggest converting floats to a tensor in the inits.
> 
> Then inside each call function, the amplitude is moved onto the same device as `x.device` is it does not match at the moment.
> 
> Because the workflow determines the device, we shouldn't assume that the user will know what the device is at the time of construction.

Thanks for the catch, just pushed an update to include these suggestions.

PR #108, opened 2024-07-29
add local option to WB2 climatology

dallasfoster:
# Earth2Studio Pull Request

## Description
Made it so that local copies of WB2 Climatology can be used with existing WB2 objects.

Changed the structure of `data.xr` objects so that keyword arguments can be passed in the object instantiate method.
 
## Checklist

### DIFF
--- a/earth2studio/data/xr.py
+++ b/earth2studio/data/xr.py
@@ -32,13 +32,9 @@ class DataArrayFile:
         Path to xarray data array compatible file.
     """
 
-    def __init__(self, file_path: str):
+    def __init__(self, file_path: str, **xr_args: Any):
         self.file_path = file_path
-
-    @property
-    def da(self, xr_args: dict[str, Any] = {}) -> xr.DataArray:
-        """Return the appropriate xarray DataArray."""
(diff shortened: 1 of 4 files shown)


earth2studio/data/xr.py
```diff
@@ -74,14 +70,9 @@ class DataSetFile:
         Data array name in xarray dataset
     """
 
-    def __init__(self, file_path: str, array_name: str):
+    def __init__(self, file_path: str, array_name: str, **xr_args: Any):
```
NickGeneva, line 73: could this be `backend_kwargs: dict{str, Any]` so it matches the other IO classes?

Also doc string needs a quick update.

NickGeneva (approved):
Approving to unblock

earth2studio/data/xr.py
dallasfoster: Yes! Good point.

dallasfoster: This is tricky because technically these keyword arguments go directly to xarray's `open_dataset` and among those keyword arguments is an argument explicitly labeled as `backend_kwargs`. I am worried about users misinterpreting the argument names here.

PR #164, opened 2024-12-13
Add StormCast model

pzharrington:
# Earth2Studio Pull Request

## Description
Adds the StormCast model as a prognostic model. Due to the use of curvilinear coordinates in HRRR data (lambert conformal projection) and conditioning data this incurs some minor changes elsewhere; mainly the ability to perform xarray interpolation to a desired set of coordinates has been added to `fetch_data`.

## Checklist

## Dependencies

The StormCast model depends on newly added material in modulus which has not yet been released, so the toml needs an update.


earth2studio/models/batch.py
```diff
@@ -164,15 +164,28 @@ def _batch_wrap(self, func: Callable) -> Callable:
...
+            if conditioning is not None:
+                out, out_coords = func(
+                    model,
+                    x,
+                    flatten_coords,
+                    conditioning=conditioning,
+                    conditioning_coords=conditioning_coords,
+                )
+            else:
+                out, out_coords = func(model, x, flatten_coords)
```
dallasfoster: @NickGeneva if we force users to use a `conditioning_data_source` in the model then we can avoid making changes here.

earth2studio/models/px/__init__.py
```diff
@@ -22,3 +22,4 @@
 from earth2studio.models.px.pangu import Pangu3, Pangu6, Pangu24
 from earth2studio.models.px.persistence import Persistence
 from earth2studio.models.px.sfno import SFNO
+from earth2studio.models.px.stormcast import StormCast
```
dallasfoster, line 25: it might be prescient of us to mark this as `StormCast-V1`

earth2studio/models/px/stormcast.py
```diff
@@ -0,0 +1,379 @@
...
+INVARIANTS = ["orography", "land_sea_mask"]
+
+
+class StormCast(torch.nn.Module, AutoModelMixin, PrognosticMixin):
+    """
+
+
+    Note
+    ----
+
```
dallasfoster: Add overview description.

dallasfoster: Link to paper

earth2studio/lexicon/hrrr.py
```diff
@@ -187,8 +188,8 @@ def build_vocab() -> dict[str, str]:
                 prs_variables[f"{id}{level:d}"] = f"sfc::fcst::{level} mb::{variable}"
 
         hybrid_levels = list(range(1, 51))
-        hybrid_names = ["UGRD", "VGRD", "HGT", "TMP", "SPFH", "PRES"]
-        e2s_id = ["u", "v", "z", "t", "q", "p"]
+        hybrid_names = ["UGRD", "VGRD", "HGT", "TMP", "SPFH", "PRES", "HGT"]
+        e2s_id = ["u", "v", "z", "t", "q", "p", "Z"]
```
dallasfoster, line 192: I think we need to also include `z`.

dallasfoster:
Looks really good, just the few minor comments.

earth2studio/models/batch.py
NickGeneva: I agree thats better to force/expect the use to use a certain data source of fetch method.

imo its better to have that an explicit requirement in the model and call it out in the docs than change this behavior. Low key the batching behavior is already not so clear to users so I dont think its a good idea to further complicate its behavior for one model.

pzharrington: Ok, so we want to remove the option entirely to pass the conditioning data at runtime to `StormCast` call? Upshot of doing that is it simplifies things and makes the call signature similar to other models (removing the need for the above changes); downside is probably performance, as then each `forward` will do it's own `fetch_data`

earth2studio/lexicon/hrrr.py
pzharrington: It's there, right? Both `z` and `Z` are in the list

earth2studio/data/utils.py
```diff
@@ -43,8 +44,12 @@ def fetch_data(
     variable: VariableArray,
     lead_time: LeadTimeArray = np.array([np.timedelta64(0, "h")]),
     device: torch.device = "cpu",
+    interp_to: CoordSystem = None,
+    interp_method: str = "nearest",
```
NickGeneva, line 48: Not the biggest fan of the increase of API parameters here. If these are specific for strom cast is this needed outside of call function of stormcast?

earth2studio/models/px/__init__.py
pzharrington: I'm sort of inclined to leave it as `StormCast` for consistency with the paper, and later versions can have a V2. The first GPT was only called GPT-1 after GPT-2 came out :)

earth2studio/data/utils.py
```diff
@@ -75,34 +85,103 @@ def fetch_data(
...
-
+    # Initialize the output CoordSystem
     out_coords = OrderedDict()
     for dim in da.coords.dims:
-        out_coords[dim] = np.array(da.coords[dim])
+        if dim in ["time", "lead_time", "variable"]:
+            out_coords[dim] = np.array(da.coords[dim])
+
+    # Fetch data and regrid if necessary
+    if interp_to is not None:
```
NickGeneva, line 133: Just noting that we need to figure out how to deal with interpolation better this now makes three spots we have such functionality: map_coords, interp.py and now here.

earth2studio/models/px/__init__.py
NickGeneva: StormCast is fine if it maps to the paper I imagine, we did this with FourCastNet :+1:

earth2studio/models/batch.py
NickGeneva: This would only occur at the start of the generator correct? So not every time-step just at the beginning?

pzharrington: No, `fetch_data` would be called from within each `__call__` when iterating over the generator. The model needs temporally-aligned conditioning data for each time-step

earth2studio/lexicon/hrrr.py
NickGeneva: Whats capital Z? is it for just hgt?

NickGeneva: If its geo potential height maybe it should be `gh` as the ID? we try to make to kinda consistent with ECMWF within reason.
https://codes.ecmwf.int/grib/param-db/?search=geo

earth2studio/data/utils.py
pzharrington: These should support interpolation for any data source onto any target coordinates; I think this use case will be increasingly common as more flavors of CorrDiff/StormCast emerge and as people play with their own regional datasets. The main usage pattern is intended to have this called at the start of a workflow, before running the model -- fetch data, interp, and move to GPU, then run inference. StormCast just makes use of this in the forward pass to do the regridding of conditioning data on the fly

NickGeneva: I see. Sounds reasonable, I'm not losing sleep over this so all good. Its okay if these util methods evolve over time to match more and more with whats needed. :+1:  Thanks for clarifying

earth2studio/models/batch.py
NickGeneva: Oh, I suppose I was thinking that the model would require all conditioning data to be fetched ahead of time as the input and then do the interpolation process at the beginning for all of it.

Yeah you could not run the generator forever like say FCN, but then you keep any downloading of data out of the model and all interpolation can occur at the beginning or even on the fly. But seems I'm miss understanding how this is working.

earth2studio/lexicon/hrrr.py
pzharrington: Yeah, geopotential height. I'm not opposed to swapping to `gh` instead of `Z` but it'll require updating the NGC package as well. @dallasfoster what do you think?

earth2studio/models/batch.py
NickGeneva: But now that I think about it, you would have to pull the data remotely anyway regardless. You could also pre pull all data and then put in a local data source for the model to use. So maybe not that big of a deal.

Tldr, I think thats okay what you mentioned Peter.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add StormCast model to prognostic models=
```
NickGeneva: Extra "=" sign here?

pzharrington: Oops, will fix

NickGeneva:
Oh also add this to the RST file:
https://github.com/NVIDIA/earth2studio/blob/main/docs/modules/models.rst

So it gets build into the docs :+1:

NickGeneva (approved):
Approving to unblock, looks really good.

earth2studio/models/batch.py
pzharrington: K, pushed an update that removes these changes to the batching and requires `conditioning_data_source` to be set in `StormCast` before running the model

earth2studio/lexicon/hrrr.py
dallasfoster: Let's keep `Z` for now. I am not convinced  `gh` is superior.

earth2studio/models/px/__init__.py
dallasfoster: Disagree but will approve.

dallasfoster (approved):

PR #204, opened 2025-03-18
Rename modulus dependencies to PhysicsNeMo

saikrishnanc-nv:
# Earth2Studio Pull Request

## Description
Earth2Studio depends on PhysicsNeMo (formerly Modulus).
Since that package was renamed, this PR updates the dependencies.

## Checklist

### DIFF
--- a/test/models/px/test_dlwp.py
+++ b/test/models/px/test_dlwp.py
@@ -247,6 +247,7 @@ def model(model_cache_context) -> DLWP:
         return p
 
 
+@pytest.mark.xfail  # TODO: REMOVE
 @pytest.mark.ci_cache
 @pytest.mark.timeout(360)
 @pytest.mark.parametrize("device", ["cuda:0"])


earth2studio/models/px/stormcast.py
```diff
@@ -238,15 +238,17 @@ def load_default_package(cls) -> Package:
     def load_model(cls, package: Package) -> DiagnosticModel:
         """Load StormCast model."""
 
-        # Require appropriate modulus version
-        installed_version = Version(modulus.__version__)
+        # Require appropriate physicsnemo version
+        # TODO (@saikrishnanc): How should we handle this?
```
saikrishnanc-nv: @loliverhennigh How do you think we should handle this?

NickGeneva: Physics Nemo 1.0.0 has Stormcast in it correct?
So as long as Physics nemo is installed should be good? (So we could remove)

This was because storm cast was integrated to main but not in the wheel originally, so we had this check in here to tell people to install from source

pyproject.toml
```diff
@@ -24,7 +24,7 @@ dependencies = [
     "netCDF4>=1.6.4",
     "ngcsdk>=3.48.0,<3.55.0",
     "numpy>=1.24.0",
-    "nvidia-modulus@git+https://github.com/NVIDIA/modulus.git",
+    "nvidia-physicsnemo@git+https://github.com/NVIDIA/physicsnemo.git",
```
NickGeneva: This should get changed to just the physics nemo package on pypi, was left over form Stormcast integraiton and will prevent the wheel getting uploaded current.

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -23,7 +23,7 @@ body:
   - type: markdown
     attributes:
       value: |
-        Thanks for taking the time to help Modulus and fill out this feature request!
+        Thanks for taking the time to help PhysicsNemo and fill out this feature request!
```
NickGeneva: Maybe we should change `Earth2Studio/PhysicsNemo`

Looks like I never changed it originally :)

pyproject.toml
saikrishnanc-nv: Gotcha, good callout, thanks!

earth2studio/models/px/stormcast.py
saikrishnanc-nv: Gotcha yep - That was my suggestion in the following comment lines too, thanks for confirming!

NickGeneva (approved):

NickGeneva (approved):

PR #237, opened 2025-04-07
Cyclone tracking

mariusaurus:
# Earth2Studio Pull Request

## Description
This PR copies cyclone trackers previously integrated in internal e2s.

## Checklist

- [X] The [CHANGELOG.md](https://github.com/NVIDIA/earth2studio/blob/main/CHANGELOG.md) is up to date with these changes.


earth2studio/models/dx/cyclone_tracking.py
```diff
@@ -0,0 +1,1063 @@
...
+
+            # Calculate vorticity at 850 hPa
+            vort850 = vorticity(u850, v850)
+            vort850[361:] *= -1
+
+            # Calculate wind speed at 10m height
+            w10m = torch.sqrt(torch.square(u10m) + torch.square(v10m))
+
+            # identify position of tropical storm centers
+            centers = find_centers_wuduan(lat, lon, vort850, w10m, msl)
```
NickGeneva: Is this function only expected to be used in this model?

If so then perhaps some of these should be moved into the class, even as a class function

earth2studio/models/dx/cyclone_tracking.py
```diff
@@ -0,0 +1,1063 @@
...
+OUT_VARIABLES = ["tc_lat", "tc_lon", "tc_msl", "tc_w10m"]
+
+
+class CycloneTrackingVorticity(torch.nn.Module, AutoModelMixin):
+    """
+
+    Finds a list of tropical cyclone centers using an adaption of the method described in the conditions
+    in Wu and Duan 2023 vort850
+
+    https://doi.org/10.1016/j.wace.2023.100626.
```
NickGeneva: Nit: Add these links in a Note block so they get called out better in the dos its excellent that these are included!

E.g. https://github.com/NVIDIA/earth2studio/blob/main/earth2studio/models/px/dlwp.py#L50

NickGeneva (commented):
This is looking good, I will need some more time to comb through it tomorrow but two items for discussion:

1) Are the functions in the tracker class ever really intended to be used on their own. I know they could be... but is the main mode of interaction ever stand alone. I would consider make them class functions

2) Is there a simple example that can be put together, for example... maybe use the GFS data source to pull Harvey and then just do some manual forward passes of the model

3) One of the unit tests seem like more of an example, pulling data and what not. Is there a synthetic case that could be used potentially? Not sure but a positive test is needed but at most maybe it pulls data from one source? I'll need to look at the coverage too

NickGeneva:
Add the model to the docs here: https://github.com/NVIDIA/earth2studio/blob/main/docs/modules/models.rst?plain=1#L61

Before I forget.

PR #238, opened 2025-04-07
HENS: example

mariusaurus:
# Earth2Studio Pull Request

## Description
This PR adds the HENS pipeline into the example folder

## Checklist


examples/11_hens/conf/config.yaml
```diff
@@ -0,0 +1,72 @@
...
+#ic_block_step: 24
+
+nsteps: 20          # number of forecasting steps
+nensemble: 2       # ensemble size per checkpoint
+batch_size: 2      # inference batch size
+torch_use_deterministic_algorithms: False # has impact on performace, in many cases results can be reproduced also if it is set to False
+#batch_ids_reproduce: [3] # select specifc batchs that you want to run to reproduce certain runs
+
+forecast_model:
+    architecture: earth2studio.models.px.SFNO   # forecast model class
```
NickGeneva: When would this ever be changed?

examples/11_hens/conf/helene.yaml
```diff
@@ -0,0 +1,48 @@
+project: 'helene'
```
NickGeneva: Seems to me that this could be a structured config? I.e. i uses the base config.yaml as the default then overwrites the changed values?

https://hydra.cc/docs/tutorials/structured_config/schema/#validating-against-a-schema-in-the-same-config-group

examples/11_hens/conf/helene.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+data_source:
+    _target_: earth2studio.data.GFS  # data source class
+
+cyclone_tracking:
+    out_dir: './outputs'
+
+file_output:
+    path: './outputs'       # directory to which outfiles are written
+    output_vars: ["t2m", 'u10m', 't850', 'q850', 'z500'] # , 'tp06']
```
NickGeneva: nit: single quotes on t2m

examples/11_hens/utilities.py
```diff
@@ -0,0 +1,1050 @@
...
+            perturbation = perturbation(model=model)
+        else:
+            raise ValueError(
+                f"perturbation method {perturbation.func.__name__} not implemented for partial instantiation"
+            )
+
+    return perturbation
+
+
+def build_package_list(cfg: DictConfig) -> list[str]:
```
NickGeneva: If I understand this correctly right now you have people wgetting the whole checkpoint repo, consider updating this to using the standard way to access the checkpoint on a as needed basis.

I.e. just manage a constant list of each one of these folders and then just build the package object based on the number of models requested

https://nvidia.github.io/earth2studio/examples/11_huge_ensembles.html#set-up

examples/11_hens/ensemble_utilities.py
```diff
@@ -0,0 +1,716 @@
...
+        track_coords_final = track_coords_list[-1]
+        member_ids = track_coords_final["ensemble"]
+        del track_coords_final["ensemble"]
+        track_coords_final["time"] = time_list
+        track_coords_final["point"] = np.sort(
+            np.unique(np.concatenate([x["point"] for x in track_coords_list]))
+        )
+        return out_tensor, track_coords_final, member_ids
+
+    def connect_centres_to_tracks(
```
NickGeneva: Moving a cyclone related items into a hens_cyclone_utils.py would perhaps be good

examples/11_hens/utilities.py
```diff
@@ -0,0 +1,1050 @@
...
+    ------
+    ValueError
+        If any of the cropbox settings are out of the valid range.
+
+    Notes
+    -----
+    - The function checks that `lat_min` is greater than or equal to -90 and `lat_max` is less than or equal to 90.
+    - It also ensures that `lat_min` is less than `lat_max`.
+    - For `lon_min` and `lon_max`, the function checks that `lon_min` is less than `lon_max`.
+    - Additionally, it verifies that `lon_min` and `lon_max` are within the valid range of -180 to 180 or 0 to 360, depending on whether `lon_min` is negative or positive.
```
NickGeneva: I would suggest only supporting 0,360. Its the preferred lon order in the package and its okay to force users to do this it it means removing some of these branches / complexity

examples/11_hens/utilities.py
```diff
@@ -0,0 +1,1050 @@
...
+    lon_range_model = (
+        360 if ((lon_coords.max() > 180) and (lon_coords.min() >= 0)) else 180
+    )
+    lon_range_out = (
+        360 if ((cfg_cropbox.lon_max > 180) and (cfg_cropbox.lon_min >= 0)) else 180
+    )
+    return lon_range_model, lon_range_out
+
+
+def crop_area(
```
NickGeneva: the map_coords function should basically have this functionality If im following this correctly

https://github.com/NVIDIA/earth2studio/blob/main/earth2studio/utils/coords.py#L250

examples/11_hens/utilities.py
```diff
@@ -0,0 +1,1050 @@
...
+        {"variable": np.array(output_vars)}
+        if "file_output" in cfg and "output_vars" in cfg.file_output
+        else {}
+    )
+    output_coords["lon"] = lon_coords
+    output_coords["lat"] = lat_coords
+    return output_coords
+
+
+def get_reproducibility_settings(cfg: DictConfig) -> tuple[str | int, list[int], bool]:
```
NickGeneva: move into repo utils perhaps

examples/11_hens/utilities.py
```diff
@@ -0,0 +1,1050 @@
...
+    ----------
+    cfg : DictConfig
+        Hydra configuration object containing the settings for diagnostic models.
+
+    Returns
+    -------
+    dx_model_dict : dict
+        A dictionary containing the diagnostic models, where the keys are the model names and the values are the models.
+    """
+    dx_model_dict = {}
```
NickGeneva: Just some general thoughts

Functions like this are a good demo of why you need to be careful not lean too hard into hydra, you start trying to config everything when all you needed was an API that accepts a list of diagnostics and let the user script do the initialization lift

NickGeneva: I'm speaking from experience with modulus-sym where you end up with these crazy utils to just handle basic instantiation of objects... I was the one the wrote them

examples/11_hens/utilities.py
```diff
@@ -0,0 +1,1050 @@
...
+                tmp.to_netcdf(**kw_args)
+            elif isinstance(io, KVBackend):
+                tmp = io.to_xarray()
+                tmp = extend_xarray_for_reproducibility(tmp, io, cfg, model_dict)
+                tmp.to_netcdf(**kw_args)
+
+    return writer_threads, writer_executor
+
+
+def extend_xarray_for_reproducibility(
```
NickGeneva: doesnt hydra save what the config is for some of this stuff?

examples/11_hens/hens.py
```diff
@@ -0,0 +1,144 @@
...
+    # run forecasts
+    then = datetime.now()
+    track_list = []
+
+    for pkg, ic, ens_idx, batch_ids_produce in ensemble_configs:
+        # TODO: add start time as optional call, so it works for run.ensemble and run_hens
+        #       without having to pass start_time in initialisation
+        base_seed_string = create_base_seed_string(pkg, ic, base_random_seed)
+        # load new weights if necessary
+        model_dict = update_model_dict(model_dict, pkg)
```
NickGeneva: where is the code that places the model on or off the GPU?  I wasnt able to find it

examples/11_hens/utilities.py
```diff
@@ -0,0 +1,1050 @@
...
+            model_dict["class"].load_default_package()
+            if package == "default"
+            else Package(package)
+        )
+        model_dict["model"] = model_dict["class"].load_model(package=package)
+
+    return model_dict
+
+
+def store_tracks(area_name: str, tracks: list[pd.DataFrame], cfg: DictConfig) -> None:
```
NickGeneva: My suggestions with the tracks is just convert the tensor and coords from the tracker into a xarray data array with the NaNs, then just save it all the netCDF. No need to convert into some data frame. Let the plotting functions deal with the filtering / mapping each point to a timestep.

Seems like a tough data structure to save... since theres potential holes in the paths.

examples/11_hens/utilities.py
```diff
@@ -0,0 +1,1050 @@
...
+            "random_seed",
+            "model_package",
+        ]
+    ]
+    cols_uint16 = ["point_number", "ens_member"]
+    tracks[cols_uint16] = tracks[cols_uint16].astype("uint16")
+    tracks = tracks.rename(columns={"track_id_global": "track_id"})
+    dir_path = os.path.join(cfg.cyclone_tracking.out_dir, area_name)
+    if DistributedManager().rank == 0:
+        os.makedirs(dir_path, exist_ok=True)
```
NickGeneva: potential issue is rank 1 or up hits here first?

examples/11_hens/hens.py
NickGeneva: NVM I found it

examples/11_hens/hens.py
```diff
@@ -0,0 +1,144 @@
...
+        tracks, io_dict = run_hens()
+        track_list.append(tracks)
+        # TODO collect cyclone tracks over all members
+
+        # for k, v in df_tracks_dict.items():
+        #     v["ic"] = pd.to_datetime(ic)
+        #     all_tracks_dict[k].append(v)
+
+        # if in-memory flavour of io backend was chosen, write content to disk now
+        if io_dict:
```
NickGeneva: Are there instances where this isnt true?

examples/11_hens/hens.py
```diff
@@ -0,0 +1,144 @@
...
+            writer_threads, writer_executor = write_to_disk(
+                cfg,
```
NickGeneva: bleh, we need to make some of these async eventually, sorry about that

examples/11_hens/hens_perturbation.py
```diff
@@ -0,0 +1,163 @@
...
+
+        This function creates a variable-specific noise vector based on forecast skill scores.
+        The noise amplitude for each variable is scaled according to its forecast skill at the
+        specified lead time, with variables not in the perturbation list set to zero.
+
+        Parameters
+        ----------
+        model : PrognosticModel
+            The prognostic model used for forecasting
+        skill_path : str, optional
```
NickGeneva: Consider defining a package that points to the hugging face repo... so it will auto download and cache for the user

examples/11_hens/utilities.py
```diff
@@ -0,0 +1,1050 @@
...
+            cfg["file_output"]["path"] = "outputs/"
+
+    pkg = model_dict["package"]
+    if pkg != "vanilla":
+        pkg = "_pkg_" + pkg.split("_")[-1]
+
+    file_name = cfg.project + "_" + str(time)[:13] + pkg
+
+    io_dict = {}
+    for k in output_coords_dict.keys():
```
NickGeneva: I'm trying to follow what the output_coords_dict? Why are multiple IO stores needed?

examples/11_hens/utilities.py
```diff
@@ -0,0 +1,1050 @@
...
+
+    for k, io in io_dict.items():
+        out_path = os.path.join(cfg.file_output.path, k, file_name)
+
+        kw_args = {"path": out_path + ".nc", "format": "NETCDF4"}
+
+        if writer_executor is not None:
+            if isinstance(io, XarrayBackend):
+                tmp = io.root
+            elif isinstance(io, KVBackend):
```
NickGeneva: Are there instances where other data stores are preferred. IMO pushing people to zarr store would be best future proofing for newer developments. Less may be more here

examples/11_hens/utilities.py
NickGeneva: Dont necessarily need to change anything, but seems this would be potentially more confusing to a user compared to just having them instantiate the objects pythonically if you think they would want to customize it... maybe this isnt intended. Something to consider

examples/11_hens/conf/config.yaml
mariusaurus: Had two customers who also tried the pipeline with other models to compare quality of results. Could imagine more need for that once GC, AIFS and other global models are added

examples/11_hens/conf/helene.yaml
mariusaurus: Given that the current config stays below 50 lines, ie fits on a screen, even with all options triggered, splitting it up would overcomplicate things imo.

examples/11_hens/utilities.py
mariusaurus: issue here are all our EU use cases, which typically span from neg to pos lon.

examples/11_hens/hens.py
mariusaurus: yeah when looking at full TC seasons storing even part of the fields becomes a problem. In that case users only store tracks and if they want to have a closer look at specific ensemble member with interesting tracks, they can reproduce.

examples/11_hens/hens.py
mariusaurus: no worries. Currently, this async implementation isn't stable either. Maybe should remove it for now and add proper support to standard backends. Not that urgent as nobody's currently using it afaik

examples/11_hens/utilities.py
mariusaurus: One customer wanted to use/are using the netcdf backend. Agree with Zarr, as also this would allow one store for all processes. Reason why I did not use it initially was that zarr slowed down inference speed by 1.5-2x when storing a handful of variables globally. Later found out that this might have been due to compression (which is switched on by default in zarr python library, should we maybe have it switched off by default in the zarr io backend?), but would need to re-evaluate.
If we can integrate a single zarr store to which all processes write, that is as fast as the other backends it would def be an option to limit to that one. 

For reference: the netcdf backend was pretty much on-par with the in-memory Xarray and KV backends. Only advantage with in memory was the option of async io, which gave around 10% speedup overall. But since it's not stable currently these backends are not that useful indeed anymore.

examples/11_hens/utilities.py
mariusaurus: but can have a look if this can be streamlined. Or could ask Georg to review who wrote that seciton

examples/11_hens/utilities.py
mariusaurus: will have a look at it and then comment or adapt. Not super familiar with that section as also written by Georg.

examples/11_hens/utilities.py
mariusaurus: Two customers were asking for support for multiple lat/lon boxes to output

gertln: Customers want to get the most out of their runs without using too much storage. They prefer not to save all the output, just the important parts. These areas of interest, like North Atlantic hurricanes and Pacific typhoons, are sometimes far apart.

examples/11_hens/utilities.py
gertln: If we only support 0-360 degrees, Europe will always be split into two sections, requiring extra postprocessing to combine them. This is very inconvenient for users.

examples/11_hens/utilities.py
gertln: Some of the data is stored by Hydra, but not everything needed for reproducibility. It's more convenient to have this information directly stored in the output files (NetCDF, CSV, etc.). This way, you can't lose the information, and it is all in one place. Easy to track how a specific output file was created

examples/11_hens/utilities.py
gertln: could we remove the if statement to avoid this potential issue? Should be fine with exist_ok=True

examples/11_hens/utilities.py
NickGeneva: I see, yeah that makes sense because of the crop region. I think its fine if the recipe supports it with the expectation that everything inside the package will be expecting 0,360 99% of the time.

The main concern is that being strict on 0,360 makes consistency a lot easier in the package.

examples/11_hens/utilities.py
NickGeneva: interesting, yeah I see this torch string thing going on... im assuming its related to that

examples/11_hens/utilities.py
NickGeneva: Sounds good, I Changed it

examples/11_hens/utilities.py
NickGeneva: Updated this myself

examples/11_hens/hens.py
NickGeneva: Got it, thanks for clarification.

examples/11_hens/utilities.py
NickGeneva: I see, makes sense. Thanks for the info

examples/11_hens/utilities.py
NickGeneva: I see, thanks a ton for this context. The IO in general needs a major overhaul.

In the future when you encounter stuff like this, please surface it and dont assume that it needs to be like that because this is very insightful. This type of info that IO is bottle necking an inference workflow is exactly the information we need to help guide engineering effort.

NickGeneva: Tracking in this issue: https://github.com/NVIDIA/earth2studio/issues/283

Will see what we can do.

examples/11_hens/hens_perturbation.py
NickGeneva: Will track here: https://github.com/NVIDIA/earth2studio/issues/286

examples/11_hens/hens.py
NickGeneva: Tracking here: https://github.com/NVIDIA/earth2studio/issues/286

examples/11_hens/utilities.py
NickGeneva: Ignore, the -180:180 requirement makes it not usable

examples/11_hens/conf/helene.yaml
NickGeneva: The benefit of that is so missing values are always populated.
Namely the user changes what they need in their yaml, and if they miss a stanza it gets filled with defaults.

Its like 2 lines at the top of the YAML, but leaving it for now...

PR #274, opened 2025-04-22
HENS Mini Example

NickGeneva:
# Earth2Studio Pull Request

## Description

- Improves the HENs perturbation to support a single batch.
- Adds a HENs mini example for sphinx gallery with the goal to just show how to load the checkpoints and use the perturbation method. Not be 100% consistent with the paper.

## Checklist

### DIFF
--- a/earth2studio/perturbation/hcbv.py
+++ b/earth2studio/perturbation/hcbv.py
@@ -75,7 +75,7 @@ class HemisphericCentredBredVector:
             if isinstance(noise_amplitude, torch.Tensor)
             else torch.Tensor(
                 [noise_amplitude] * len(self.model.input_coords()["variable"])
-            )
+            )[:,None,None]
         )
         self.integration_steps = integration_steps
         self.seeding_perturbation_method = seeding_perturbation_method
--- a/earth2studio/perturbation/hcbv.py
+++ b/earth2studio/perturbation/hcbv.py
@@ -81,6 +81,7 @@ class HemisphericCentredBredVector:
         self.seeding_perturbation_method = seeding_perturbation_method
         self.set_clip_indices()
         self.generator: Generator | None = None
+        self.device = "cpu"
 
     @torch.inference_mode()
     def create_generator(
@@ -109,8 +110,8 @@ class HemisphericCentredBredVector:
         while True:
             # get unperturbed intital state, assuming tensor always has 6 dims
             xunp = input_data[:1].repeat(batch_size, 1, 1, 1, 1, 1).to(self.device)
-            # Commenting here, that this will not work for any model that needs a
-            # history window greater than one
+            # Commenting here, not tested for when multiple lead times are needed
+            # May work...
             coords["time"] = data_coords["time"][:1]
 
             # generate perturbed initial state
@@ -130,7 +131,7 @@ class HemisphericCentredBredVector:
                     )
 
                 # scale dx
-                dx = self.noise_amplitude[:, None, None] * (dx / hem_norm)
+                dx = self.noise_amplitude * (dx / hem_norm)
 
                 xunp = (
                     input_data[ii + 1]
--- a/earth2studio/perturbation/hcbv.py
+++ b/earth2studio/perturbation/hcbv.py
@@ -81,11 +81,10 @@ class HemisphericCentredBredVector:
         self.seeding_perturbation_method = seeding_perturbation_method
         self.set_clip_indices()
         self.generator: Generator | None = None
-        self.device = "cpu"
 
     @torch.inference_mode()
     def create_generator(
-        self, time: TimeArray, batch_size: int = 1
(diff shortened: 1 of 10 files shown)


NickGeneva:
For the perturbation update here's the script I used to generate two tensors:
```python
from datetime import datetime
import numpy as np
import torch
from earth2studio.data import Random
from earth2studio.models.batch import batch_coords, batch_func
from earth2studio.perturbation import HemisphericCentredBredVector
from earth2studio.perturbation.gaussian import Gaussian
from earth2studio.utils.type import CoordSystem

class FooModel(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.register_buffer("scale", torch.Tensor([0.1]))
        self.index = 0
        self._input_coords = None

    def input_coords(self):
        return self._input_coords

    @batch_coords()
    def output_coords(self, input_coords: CoordSystem):
        output_coords = input_coords.copy()
        output_coords["lead_time"] = np.array([np.timedelta64(1, "s")])
        return output_coords

    @batch_func()
    def forward(self, x, coords):
        self.index += 1
        return self.scale * x, coords

device = "cpu"
time = datetime(2024,1,1)
# Domain coordinates
batch = 2
variable = ['tcwv', 'u10m', 'v10m']
dc = OrderedDict(
    [
        ("lat", [f"{nn}" for nn in range(16)]),
        ("lon", [f"{nn}" for nn in range(16)]),
    ]
)
fc = OrderedDict(
    [
        ("batch", np.empty(0)),
        ("lead_time", np.array([np.timedelta64(0, "h")])),
        ("variable", np.array(variable)),
    ]
)
fc.update(dc)

model = FooModel()
model._input_coords = fc
model = model.to(device)
model.index = 0

torch.manual_seed(0)
np.random.seed(0)

# Initialize Data Source and input tensor
data_source = Random(dc)
x = torch.randn(batch, 1, 1, len(variable), 16, 16).to(device)
coords = OrderedDict(
    [
        ("ensemble", np.arange(batch + 1)),
        ("time", np.array([time])),
        ("lead_time", np.array([np.timedelta64(0, "h")])),
        ("variable", np.array(variable)),
    ]
)
coords.update(dc)


seeding_perturbation_method  = Gaussian()

prtb = HemisphericCentredBredVector(
    model=model,
    data=data_source,
    # time=time,
    seeding_perturbation_method=seeding_perturbation_method,
    noise_amplitude=0.1,
    integration_steps=3,
)
xout, coords = prtb(x, coords)

torch.save(xout, "output_new.pt")
```

Running on both main branch and then this branch, then executing the following:

```python
import torch
old = torch.load("output_old.pt")
new = torch.load("output_new.pt")
assert torch.allclose(old, new), "Failed"
```

Note that for batch greater than two the following is updated (tensor alternates now between positive and negative noise)

```python
import torch
old = torch.load("output_old.pt")
new = torch.load("output_new.pt")
b = old.shape[0]
assert torch.allclose(old[:b//2], new[::2]), "Failed"
assert torch.allclose(old[b//2:], new[1::2]), "Failed"
```

NickGeneva:
Rendered example
_home_ngeneva_Downloads_artifacts__build2_html_examples_11_huge_ensembles html

earth2studio/perturbation/hcbv.py
```diff
@@ -77,40 +75,77 @@ def __init__(
...
-            )
+            )[:, None, None]
         )
         self.integration_steps = integration_steps
         self.seeding_perturbation_method = seeding_perturbation_method
-
-        self.fetch_warmup_ics(time)  # make work with ic_block
         self.set_clip_indices()
+        self.generator: Generator | None = None
+        self.device = "cpu"
```
mariusaurus: do we need to set `self.device` here? Hardcoding it to cpu can be a bit confusing, as it will anyway be set thee first time the perturbation is called. Could alternatively set it to `None`?

mariusaurus (approved):
Really like the generator approach removing the even-numbered batch size requirement and allowing for an interface like the other perturbation methods have.

Running the example took around an hour for the first time as downloads were rather sluggish, also for ICs. But this might have been down to GFS server, so out of our control. Once the workflow is added to the recipe book we should maybe add a reference.

earth2studio/perturbation/hcbv.py
NickGeneva: I initialized it here in case the user invokes the "create_generator" function ahead of the call function for some reason, just so it doesnt fail.

mariusaurus: would it still work though if the user invokes the iterator, initially, but then moves to the GPU? or if the model is already on the GPU? could also set it to model.device in that case?

NickGeneva: In that case it will fail, true. Maybe I should change this into a parameter of the create_generator function instead. Let me see what I can do

NickGeneva: Alright updated this to just remove that prop outright.

PR #279, opened 2025-04-25
ARCO update for async

NickGeneva:
# Earth2Studio Pull Request

## Description

Still using nest_asyncio to help with notebooks, need to understand the fsspec solution more.
https://github.com/fsspec/filesystem_spec/blob/master/fsspec/asyn.py#L108

Zarr 3.0 still disable atm because of the caching issue that exists current, but have tested this.
Seems to be around 10x faster when pulling something like SFNO initial conditions

## Checklist

### DIFF
--- a/earth2studio/data/arco.py
+++ b/earth2studio/data/arco.py
@@ -116,7 +116,7 @@ class ARCO:
             self.zarr_group = zarr.open(fs_map, mode="r")
 
         self.async_timeout = async_timeout
-        self.async_process_limit = 64
+        self.async_process_limit = 128
 
     def __call__(
         self,
--- a/earth2studio/data/arco.py
+++ b/earth2studio/data/arco.py
@@ -116,7 +116,7 @@ class ARCO:
             self.zarr_group = zarr.open(fs_map, mode="r")
 
         self.async_timeout = async_timeout
-        self.async_process_limit = 128
+        self.async_concurrency_limit = 128
 
     def __call__(
         self,
@@ -197,7 +197,8 @@ class ARCO:
             },
         )
 
-        sem = asyncio.Semaphore(self.async_process_limit)
+        # Alternatives: https://death.andgravity.com/limit-concurrency
+        sem = asyncio.Semaphore(self.async_concurrency_limit)
         args = [
             (t, i, v, j) for j, v in enumerate(variable) for i, t in enumerate(time)
         ]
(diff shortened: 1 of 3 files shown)


earth2studio/data/arco.py
```diff
@@ -260,18 +253,42 @@ def fetch_array(self, time: datetime, variable: str) -> np.ndarray:
...
+                data = await zarr_array.getitem(slice(None))
+                output = modifier(data)
+            # Surface variable
+            elif len(shape) == 3:
+                data = await zarr_array.getitem(time_index)
+                output = modifier(data)
+            # Atmospheric variable
+            else:
+                # Load levels coordinate system from Zarr store and check
+                level_coords = await (await self.zarr_group.get("level")).getitem(
```
akshaysubr: Can you not cache this in the constructor so each call doesn't have to do this? Also, each variable will do this so caching will also eliminate redundancy here?

earth2studio/data/arco.py
```diff
@@ -207,30 +197,36 @@ async def create_data_array(
...
 
-    def fetch_array(self, time: datetime, variable: str) -> np.ndarray:
+    async def fetch_wrapper(
+        self,
+        e: tuple[datetime, int, str, int],
+        xr_array: xr.DataArray,
+        sem: asyncio.Semaphore,
+    ) -> None:
+        """Small wrapper that is awaitable for async generator"""
+        async with sem:
```
akshaysubr: I'm not quire following why this is needed. Zarr v3 internally has concurrency limits too right? Also, you're not spawning any threads here, only using async right? So what issue is this semaphore here solving?

akshaysubr (approved):
Overall, looks like this is properly leveraging async zarr and only synchronous calls are all the "compute" elements that need CPU cycles. Great that it yields a 10x speedup. Only had a suggestion to remove some redundant I/O requests and a question about the semaphore pattern.

earth2studio/data/arco.py
NickGeneva: Good point, move it up out of this function

earth2studio/data/arco.py
NickGeneva: This is fair, was trying to provide the user some control built in to rate limit how many requests are sent to a given api at once... but that can be done via zarr if needed. Removed.

PR #280, opened 2025-04-25
Add DLESyM Model

pzharrington:
# Earth2Studio Pull Request

Description from the newly added model:

> DLESyM-V1-ERA5 prognostic model. This is an ensemble forecast model for
    global earth system modeling. This model includes an atmosphere and ocean
    component, using atmospheric variables as well as the sea-surface temperature
    on a HEALPix nside=64 (approximately 1 degree) resolution grid. The model
    architecture is a U-Net with padding operations modified to support using
    the HEALPix grid. Because the atmosphere and ocean models are predicted at
    different times, not all entries in the output tensor are valid. As a result,
    we provide convenience methods for retrieving the valid atmospheric and
    oceanic outputs.

Example:

        pkg = DLESyM.load_default_package()
        model = DLESyM.load_model(pkg)

        # Create iterator
        iterator = model.create_iterator(x, coords)

        for step, (x, coords) in enumerate(iterator):
            # Valid atmos and ocean predictions with their respective coordinates extracted below
            atmos_outputs, atmos_coords = model.retrieve_valid_atmos_outputs(x, coords)
            ocean_outputs, ocean_coords = model.retrieve_valid_ocean_outputs(x, coords)
            ...
 

## Checklist


earth2studio/models/px/dlesym.py
```diff
@@ -0,0 +1,1077 @@
...
+from earth2studio.models.auto import AutoModelMixin, Package
+from earth2studio.models.batch import batch_coords, batch_func
+from earth2studio.models.px.base import PrognosticModel
+from earth2studio.models.px.utils import PrognosticMixin
+from earth2studio.utils import check_extra_imports
+from earth2studio.utils.type import CoordSystem
+
+_ATMOS_VARIABLES = [
+    "z500",
+    "tau300-700",
```
NickGeneva, line 42: whats this? The difference between the two levels?

earth2studio/models/px/dlesym.py
```diff
@@ -0,0 +1,1077 @@
...
+                f"Atmos input times must be equal to or longer than ocean input times, got {len(atmos_input_times)} and {len(ocean_input_times)}"
+            )
+        if len(atmos_output_times) < len(ocean_output_times):
+            raise ValueError(
+                f"Atmos output times must be equal to or longer than ocean output times, got {len(atmos_output_times)} and {len(ocean_output_times)}"
+            )
+
+        self.atmos_input_times = atmos_input_times
+        self.ocean_input_times = ocean_input_times
+        self.full_input_times = np.arange(
```
NickGeneva, line 218: I'm having trouble following what the full_input_times is, its a index list?
I suppose I was kind of expecting a np.concat(atmost_times, ocean_times)

earth2studio/models/px/dlesym.py
```diff
@@ -0,0 +1,1077 @@
...
+        -------
+        CoordSystem
+            Coordinate system dictionary
+        """
+
+        output_coords = OrderedDict(
+            {
+                "batch": np.empty(0),
+                "time": np.empty(0),
+                "lead_time": self.atmos_output_times,
```
NickGeneva: just the atmos varaibles?

earth2studio/models/px/dlesym.py
```diff
@@ -0,0 +1,1077 @@
...
+            cache=False,
+        )
+
+    @classmethod
+    @check_extra_imports("dlesym", [Module, OmegaConf])
+    def load_model(
+        cls,
+        package: Package,
+        atmos_model_idx: int = 0,
+        ocean_model_idx: int = 0,
```
NickGeneva, line 359: These are fine since theres default, but add params section to the doc string to explain what these mean

I should adjust the definition in AutoModelMixin to allow kwargs, this comes up too many times.

earth2studio/models/px/dlesym.py
```diff
@@ -0,0 +1,1077 @@
...
+            for t in self.atmos_input_times + coords["lead_time"][-1]
+        ]
+        ocean_lead_time_idx = [
+            list(coords["lead_time"]).index(t)
+            for t in self.ocean_input_times + coords["lead_time"][-1]
+        ]
+
+        # Atmos inputs: state, insolation, constants, coupled inputs
+        atmos_state = x[:, atmos_lead_time_idx][..., atmos_var_idx, :, :, :].permute(
+            0, 3, 1, 2, 4, 5
```
NickGeneva: if I understand this correctly the healpix face id is the second dim of the models input?

earth2studio/models/px/dlesym.py
```diff
@@ -0,0 +1,1077 @@
...
+            ),
+            atmos_variables=cfg.io.atmos_variables,
+            ocean_variables=cfg.io.ocean_variables,
+            atmos_coupling_variables=cfg.io.atmos_coupling_variables,
+            ocean_coupling_variables=cfg.io.ocean_coupling_variables,
+        )
+
+    def prepare_input_data(
+        self, x: torch.Tensor, coords: CoordSystem
+    ) -> tuple[torch.Tensor, torch.Tensor]:
```
NickGeneva: doesnt this return a tuple[list[Tensor], list[Tensor]] or did I miss the concat?

earth2studio/models/px/dlesym.py
```diff
@@ -0,0 +1,1077 @@
...
+        coords : CoordSystem
+            Input coordinates
+
+        Returns
+        -------
+        output_data : torch.Tensor
+            Output data
+        """
+
+        output_data = torch.zeros(
```
NickGeneva: Would torch empty work here? Should be just a little faster

earth2studio/models/px/dlesym.py
```diff
@@ -0,0 +1,1077 @@
...
+                :, :, :, src_idx, :, :, :
+            ]
+
+        return output_data
+
+    @check_extra_imports("dlesym", [insolation])
+    def _make_insolation_tensor(
+        self, anchor_times: np.ndarray, timedeltas: np.ndarray
+    ) -> torch.Tensor:
+        """Make insolation tensor from anchor times and timedeltas
```
NickGeneva, line 588: needs space between this line and parameters

take advantage of the autodocstring extension in vscode incase youre not using it:
https://nvidia.github.io/earth2studio/userguide/developer/documentation.html#api-documentation

earth2studio/models/px/dlesym.py
```diff
@@ -0,0 +1,1077 @@
...
+        ]
+        ocean_coupling = x[:, :, :, ocean_coupling_channel_idx, :, :]
+
+        # Slice along the lead_time dim, average, and stack along the variable dim
+        # (different time-averaged quantities are defined as different variables)
+        slices = ocean_coupling.chunk(len(self.ocean_input_times), dim=2)
+        ocean_coupling = torch.concat(
+            [s.mean(dim=2, keepdim=True) for s in slices], dim=3
+        )
+        ocean_coupling = ocean_coupling.permute(2, 0, 3, 1, 4, 5)
```
NickGeneva, line 672: why is this different than the _make_atmos_coupling, do the models have different input dimensions?

earth2studio/models/px/dlesym.py
```diff
@@ -0,0 +1,1077 @@
...
+        return (x - self.center) / self.scale
+
+    def _denormalize_output(self, x: torch.Tensor) -> torch.Tensor:
+        """Denormalize output data"""
+
+        return x * self.scale + self.center
+
+    def retrieve_valid_ocean_outputs(
+        self, x: torch.Tensor, coords: CoordSystem
+    ) -> torch.Tensor:
```
NickGeneva: tuple[tensor, coords]

earth2studio/models/px/dlesym.py
```diff
@@ -0,0 +1,1077 @@
...
+        ----------
+        input_coords : CoordSystem
+            Input coordinate system to transform into output_coords
+
+        Returns
+        -------
+        CoordSystem
+            Coordinate system dictionary
+        """
+
```
NickGeneva, line 313: output coords should also be used to validate the input_coords to make sure things look appropriate

e.g. https://github.com/NVIDIA/earth2studio/blob/main/earth2studio/models/px/fcn.py#L151

earth2studio/models/px/dlesym.py
```diff
@@ -0,0 +1,1077 @@
...
+            list(coords["variable"]).index(var) for var in self.ocean_variables
+        ]
+        ocean_outputs = ocean_outputs.index_select(
+            dim=3, index=torch.tensor(ocean_var_idx, device=x.device)
+        )
+        return ocean_outputs, out_coords
+
+    def retrieve_valid_atmos_outputs(
+        self, x: torch.Tensor, coords: CoordSystem
+    ) -> torch.Tensor:
```
NickGeneva: tuple[tensor, coords]

earth2studio/models/px/dlesym.py
```diff
@@ -0,0 +1,1077 @@
...
+        torch.Tensor
+            Atmos outputs
+        CoordSystem
+            Output coordinates
+        """
+
+        out_coords = coords.copy()
+        out_coords["variable"] = np.array(self.atmos_variables)
+
+        atmos_var_idx = [
```
NickGeneva: I see this index search done a lot. Is this needed, or can these be hard coded once? Maybe theres a slice operation that can be used instead? Curious of how important it is to have this flexible and calced every forward pass?

earth2studio/models/px/dlesym.py
```diff
@@ -0,0 +1,1077 @@
...
+        ----------
+        input_coords : CoordSystem
+            Input coordinate system
+
+        Returns
+        -------
+        CoordSystem
+            Output coordinate system
+        """
+        coords = super().output_coords(input_coords)
```
NickGeneva, line 991: same, comment to add some input coord validation, also input coords will be lat/lon. Will this work calling super output corrds?

earth2studio/models/px/dlesym.py
```diff
@@ -0,0 +1,1077 @@
...
+        coords = self.coords_to_ll(coords)
+        return coords
+
+    def to(self, device: Any) -> PrognosticModel:
```
NickGeneva: Are these not torch modules? curious why this is needed vs torch knowing to move `regrid_to_hpx` and `regrid_to_ll` over automatically. I'm not that familiar with e2grid

earth2studio/models/px/dlesym.py
```diff
@@ -0,0 +1,1077 @@
...
+            input_coords["lead_time"][-1] + output_coords["lead_time"]
+        )
+
+        return output_coords
+
+    @classmethod
+    def load_default_package(cls) -> Package:
+        """Default DLESyM model package on NGC"""
+        # TODO use NGC package when ready
+        return Package(
```
NickGeneva: Just adding comment to resolve before merge

earth2studio/models/px/dlesym.py
pzharrington: Correct, `tau300-700 := z300 - z700`. This model will need a slightly tweaked `fetch_data` to plug into existing data sources, shouldn't be too hard in a one-off example but maybe not the easiest to directly drop into the existing `run` workflows.

earth2studio/models/px/dlesym.py
pzharrington: It's more of a merge of the atmos_times and ocean_times rather than a concat -- we need data as far back as `-48h` for the ocean model, whereas we need data on `6h` time resolution for the atmos model; ultimately this all has to go into one `(Tensor, CoordSystem)` pair as input to the model. An easy way to satisfy this is to get the full 48h of data at 6h resolution and then pull out what we need. Now that I think of it though, a simple concat could work too, as long as the `prepare_input_data` method correctly pulls out the desired ocean and atmos timesteps. Might be more efficient, but a bit less intuitive for preparing input data tensors

earth2studio/models/px/dlesym.py
pzharrington: Atmos has the finer time resolution. Again this relates to having to force the atmos and ocean input/output data into one `(Tensor, CoordSystem)` pair. The lead time dimension of the output in theory is from 0 hours to `max(atmos_output_times, ocean_output_times)` with a resolution of `min(atmos_output_time_resolution, ocean_output_time_resolution)`; since the ocean is slower moving than the atmosphere, the atmos time resolution will always be finer than the ocean. Thus the list of atmos output times includes all the actual ocean output times (`+48h`, `+96h`), and the output tensor will just have zeros in the SST for the non-valid output ocean times.

pzharrington: Since that is a departure from other e2studio models and a potential gotcha, I wanna make sure it's documented clearly enough. Let me know if you have suggestions for the `DLESyM` docstring to clarify

earth2studio/models/px/dlesym.py
pzharrington: Correct. Within a face it's simple 2D convs with `healpix_pad` handling the correct padding behavior, so the face dim effectively becomes another batch dim. Internally in the U-Net model forward, the face dim gets folded into the batch dim and then unfolded for conv-like ops

earth2studio/models/px/dlesym.py
pzharrington: Ah you're right, at some point previously I had issues with these but now they're getting moved properly. Will remove

earth2studio/models/px/dlesym.py
pzharrington: You're right, it's a tuple of lists. Will fix

earth2studio/models/px/dlesym.py
pzharrington: The ocean coupling inputs are the outputs of the atmosphere model (for lead times +6h, ..., +96h) with time-averaging applied to them. There is no time averaging for the atmos coupling inputs.

earth2studio/models/px/dlesym.py
pzharrington: The variable indices can definitely be hard-coded in `__init__` and re-used. I think it should also be doable for the lead time indices, though it will add some clutter to `__init__` as the indexing needs to be done for the input states, coupling inputs, and output states for each of the atmos/ocean models. Probably worth it though for speed (and decluttering the other routines)

earth2studio/models/px/dlesym.py
pzharrington: It does work :) we use the super `output_coords` to get the lead times etc correct, but within the `DLESyM.output_coords` code all references to `self` resolve to the `DLESyMLatLon` when called from within `DLESyMLatLon` this way, and so lat and lon are also correctly used during validation. It's a bit of inheritance magic but it seems to work

earth2studio/models/px/dlesym.py
NickGeneva: Ah I see, okay I'm following.

earth2studio/models/px/dlesym.py
NickGeneva: I'm thinking maybe just a inline comment here and then with the comment above with the input variables would be good for developers to remind ourselves why we did this since its a bit of a work around.

earth2studio/models/px/dlesym.py
NickGeneva: What ever is easier to implement / clearer to read. Up to you!

earth2studio/models/px/dlesym.py
NickGeneva: Update to throw a meaningful NotImplemented error maybe saying the package will be getting posted soon.

NickGeneva (approved):

Issue #227, opened 2025-03-27
🐛[BUG]: Stormcast coords should be improved


NickGeneva:
### Version

0.5.0

### Describe the issue

Presently stormcast coords are set to lat lon, when they should be using hrrr_y hrrr_x. This will require some work to figure out what impact this will have on the workflows.

Presently there is a hard coded patch in the utils.py for prep_data to make sure things stay working... may consider moving interpolation methods inside the model as well.

PR #307, opened 2025-05-14
Correcting Stormcast coordinates

NickGeneva:
# Earth2Studio Pull Request

## Description

Closes: https://github.com/NVIDIA/earth2studio/issues/227

What this does, changes the coordinates of the StormCast model to the HRRR index values.

This allows HRRR input to just be used directly.
Lat/Lon values are still provided in the output, but are there as a temporary stopgap. I pre-prended it with an underscore to denote its not connected to a dimension for the time being. We need to delete a lot of this code and then have a more formal solution.

This process was very revealing to the challenges seen when dealing with coords. This also:

- Delete the curvilinear support in the Random data source that was only added for storm cast
- Improvements / fixes for async sources when using cached data
- Improvements to tqdm placements when using built in workflows
- Adds conditioning data source into the load_package now that kwargs with defaults are allowed

## Checklist


earth2studio/models/px/stormcast.py
```diff
@@ -374,8 +410,9 @@ def __call__(
         conditioning_coords.move_to_end("batch", last=False)
 
         # Handshake conditioning coords
-        handshake_coords(conditioning_coords, coords, "lon")
-        handshake_coords(conditioning_coords, coords, "lat")
+        # ugh the interp... have to deal with this for now, no solution
```
pzharrington: Mark with a TODO?

earth2studio/utils/coords.py
```diff
@@ -255,7 +256,8 @@ def map_coords(
             indx = np.where(inc == outc[0])[0][0]
             inc_slice = inc[indx : indx + outc.shape[0]]
             if inc_slice.shape[0] == outc.shape[0] and np.all(inc_slice == outc):
-                x_slice = [slice(None)] * len(input_coords)
+                # Min here, to deal when coords have extra meta-data
+                x_slice = [slice(None)] * min([len(input_coords), x.ndim])
```
pzharrington, line 261: Maybe also mark as TODO since eventually the `_lat`, `_lon` kludge will be removed

earth2studio/models/px/stormcast.py
```diff
@@ -247,8 +247,25 @@ def load_default_package(cls) -> Package:
     @check_extra_imports(
         "stormcast", [OmegaConf, PhysicsNemoModule, deterministic_sampler]
     )
-    def load_model(cls, package: Package) -> DiagnosticModel:
-        """Load StormCast model."""
+    def load_model(
+        cls,
+        package: Package,
+        conditioning_data_source: DataSource | ForecastSource = GFS_FX(cache=False),
```
pzharrington: Why `cache=False` as default here?

pzharrington:
Just a thought. As far as I can tell, the only reason we need `_lat` and `_lon` in the output coords here is so that we can use them in `interp_to` when the model is called and the conditioning data needs to be regridded. However, `_lat` and `_lon` are hard-coded to ` self.lat` and `self.lon` which we can basically take as inherent attributes of the model itself as they come direct from the package. They are also accessible externally already, e.g. when needed for plotting (use `model.lat`). That suggests from a code standpoint we might want to just remove them entirely, since it would clean up a fair chunk of hacks needed elsewhere.

In that scenario the StormCast model would operate like `DLESyM` -- the input/output coords are defined on the same (non-latlon) spatial dimensions, and are consistent with the shape of the returned tensor. There would be model-internal methods that would, in the `__call__`, produce `interp_coords` by merging the times/lead ties/etc in the input `coords` and `self.lat`, `self.lon`, and then use them in the data fetch via `interp_to`.

pzharrington (approved):

PR #325, opened 2025-05-20
Distributed inference

jleinonen:
# Earth2Studio Pull Request

## Checklist


swbg:
I am wondering whether it would make sense to support the distributed recipe on a single GPU as well. So users can test the code with a single GPU and then move to more GPUs later.

dallasfoster:
Will review design of distributed utility closely and likely suggest modifications.

jleinonen:
NOTE: It was reported to me that this had problems with PyTorch 2.7 so we should test with that and fix the problem if necessary before merging.

jleinonen:
NOTE: It was reported to me that this had problems with PyTorch 2.7 so we should test with that and fix the problem if necessary before merging.

Issue #353, opened 2025-06-06
🚀[FEA]: Option to fix the solar forcing for SFNO


gertln:
### Is this a new feature, an improvement, or a change to existing functionality?

New Feature

### How would you describe the priority of this feature request

Medium

### Please provide a clear description of problem you would like to solve.

SFNO forecasts automatically include the seasonal cycle because the solar zenith angle changes over time. While this is appropriate for most use cases, some modeling experiments may need the seasonal cycle disabled so that only the model’s internal variability is examined.
A solution would be to fix the solar forcing at its initialization-day value and hold it constant for the entire run.

PR #354, opened 2025-06-06
feat(sfno): Option to switch seasonal cycle due to solar forcing off

gertln:
# Earth2Studio Pull Request

## Description
Add a `sza_fixed` configuration parameter to the SFNO model to enable rollouts without seasonal cycle effects.

## Checklist

## Dependencies

Closes https://github.com/NVIDIA/earth2studio/issues/353

### DIFF
--- a/earth2studio/models/px/sfno.py
+++ b/earth2studio/models/px/sfno.py
@@ -138,6 +138,8 @@ class SFNO(torch.nn.Module, AutoModelMixin, PrognosticMixin):
         Model scale normalization tensor
     variables : np.array, optional
         Variables associated with model, by default 73 variable model.
+    sza_fixed : bool, optional
+            If True, the SZA is fixed to the first four timesteps to enable rollouts without seasonal cycle, by default False
     """
 
     def __init__(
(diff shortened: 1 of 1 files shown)


earth2studio/models/px/sfno.py
```diff
@@ -281,9 +293,19 @@ def _forward(
             for i, t in enumerate(coords["time"]):
                 # https://github.com/NVIDIA/modulus-makani/blob/933b17d5a1ebfdb0e16e2ebbd7ee78cfccfda9e1/makani/third_party/climt/zenith_angle.py#L197
                 # Requires time zone data
+                if self.sza_fixed:
+                    # convert lead timedelta to float in order to calculate modulo with 24 hours
+                    # this is needed because the lead time is a timedelta64[ns] and the modulo operator is not defined for this type
+                    # the first four sza fields are reused for the following timesteps
+                    lt = [
```
NickGeneva, line 300: My suggestion would be to not have a boolean toggle and rather something more universally extendable.

For example you could have a property that is a lambda callable that could be settable, of have a specific function (compute `sza_timestamp()`) that does this process that can be overridden as needed by a child class.

The way I see this at the moment, where the user really needs to know what they are doing. If this needs to be shown to a user, showing it via example rather than in the load_model or constuctor API is fine I would say, this is directly modifying SFNO's default behavior which spooks me.

NickGeneva: Are you planning to use this feature in an example / recipe? I'm curious. Thanks @gertln !

NickGeneva:
Closing this, would like to keep the wrappers as closely aligned with how the publication intended to run. Introducing edge case toggles isn't too ideal.

Need to see more thorough results / studies on this modification to consider adding it for now.
Can re-open if needed.

Thanks for the suggestion!

PR #364, opened 2025-06-10
Allow HCBV perturbation to handle static inputs

rodrigoalmeida94:
# Earth2Studio Pull Request

## Description
When using HCBV to perturb the inputs for GraphCast, I received the error "zero element in hemispheric norm, maybe noise amplification too small?". 

After digging into this it became clear that the zero elements causing this were the result of using a model that outputs constant inputs (like land sea mask, or geopotential at surface). Because in HCBV we create a `dx` as the difference between a perturbed prediction and unperturbed prediction - for the case where the outputs are constant, those channels will be zero, raising the error above.

In the code there is a comment that mentions: 
```
# if zero elements are requierd, replace NaNs in scaled dx with 0
```
Which then allows the method to be used in such cases. I tested this method with multiple models and the results all look reasonable after this change.

## Checklist

### DIFF
--- a/earth2studio/perturbation/hcbv.py
+++ b/earth2studio/perturbation/hcbv.py
@@ -146,6 +146,7 @@ class HemisphericCentredBredVector:
 
             # scale dx
             dx = self.noise_amplitude * (dx / hem_norm)
+            # set dx back to zero in constant inputs
             dx[..., torch.isnan(dx).all(dim=(0, 1, 2, 4, 5)), :, :] = 0
 
             xunp = (
--- a/earth2studio/perturbation/hcbv.py
+++ b/earth2studio/perturbation/hcbv.py
@@ -25,6 +25,7 @@ from earth2studio.data import DataSource, fetch_data
 from earth2studio.models.px import PrognosticModel
 from earth2studio.perturbation.base import Perturbation
 from earth2studio.utils import handshake_dim, handshake_size
+from earth2studio.utils.coords import map_coords
 from earth2studio.utils.time import to_time_array
 from earth2studio.utils.type import CoordSystem, TimeArray
 
@@ -105,6 +106,9 @@ class HemisphericCentredBredVector:
             lead_time=input_coords["lead_time"],
             device="cpu",
         )
+        input_coords["time"] = warmup_times
+        input_data, data_coords = map_coords(input_data, data_coords, input_coords)
+
         coords = OrderedDict(
             [("batch", np.arange(batch_size))] + list(data_coords.items())
         )
@@ -128,6 +132,10 @@ class HemisphericCentredBredVector:
             xper, _ = self.model(xper, coords)
 
             dx = xper - xunp
+
+            # remove zero elements, when inputs are constant
+            dx[..., dx.sum(dim=(0, 1, 2, 4, 5)) == 0, :, :] = torch.nan
+
             hem_norm = self.hemispheric_norm(dx, device)
 
             # if zero elements are requierd, replace NaNs in scaled dx with 0
@@ -138,6 +146,7 @@ class HemisphericCentredBredVector:
 
             # scale dx
             dx = self.noise_amplitude * (dx / hem_norm)
+            dx[..., torch.isnan(dx).all(dim=(0, 1, 2, 4, 5)), :, :] = 0
 
             xunp = (
                 input_data[ii + 1]


NickGeneva:
Thanks for this @rodrigoalmeida94 , we are in the middle of shipping a version so a little slow with reviews. Will be taking a look soon

earth2studio/perturbation/hcbv.py
```diff
@@ -138,6 +146,8 @@ def create_generator(
 
             # scale dx
             dx = self.noise_amplitude * (dx / hem_norm)
```
NickGeneva, line 141: Would adjusting the `noise_amplitude` tensor so its 0 for the static fields work as well?

So then the user controls what fields are deemed static / should be ignored rather than this condition? My worry is that this type of hidden behavior for some specific case could cause confusion down the line.

earth2studio/perturbation/hcbv.py
```diff
@@ -105,6 +106,9 @@ def create_generator(
             lead_time=input_coords["lead_time"],
             device="cpu",
         )
+        input_coords["time"] = warmup_times
+        input_data, data_coords = map_coords(input_data, data_coords, input_coords)
```
NickGeneva, line 111: Was graphcast a model where this would be needed? I'm trying to understand why the fetch data would not be sufficient. Thanks!

earth2studio/perturbation/hcbv.py
rodrigoalmeida94: Unfortunately not as it is currently implemented - the noise_amplitude is only applied after the subtraction `dx = xper - xunp` which results in the zeros for the static fields - and the check for the 0s happens after all that and after the hemispheric norm is computed.  

We could move the check for 0s for after the noise_amplitude is applied and only check the variables where noise_amplitude is non-zero?

earth2studio/perturbation/hcbv.py
rodrigoalmeida94: Ah yes, this was the case for GraphCast specifically for the 1 degree version that requires downscaling from the data source - I guess we could just add the `interp_to` key in there to get the correct resolution? Don't remember exactly why I did it like this... Will try it out!

earth2studio/perturbation/hcbv.py
```diff
@@ -128,6 +132,10 @@ def create_generator(
             xper, _ = self.model(xper, coords)
 
             dx = xper - xunp
+
+            # remove zero elements, when inputs are constant
+            dx[..., dx.sum(dim=(0, 1, 2, 4, 5)) == 0, :, :] = torch.nan
```
dallasfoster: I don't know if I love having fixed integers here for the sum. Is there a way to use the coordinates to determine which dimensions to sum over?

earth2studio/perturbation/hcbv.py
rodrigoalmeida94: I tried this out: indeed without using the `map_coords` bit this fails with graphcast with:
```
RuntimeError: The size of tensor a (1440) must match the size of tensor b (360) at non-singleton dimension 5
```

earth2studio/perturbation/hcbv.py
rodrigoalmeida94: Ah great point, changed it 👍

earth2studio/perturbation/hcbv.py
NickGeneva: Hi @rodrigoalmeida94 

The map coords makes sense here

earth2studio/perturbation/hcbv.py
NickGeneva: Super sorry for the delay on this @rodrigoalmeida94 !

I wanted to understand this myself and I'm curious if the following can be used to get the same effect:

```python
            xunp, _coords = self.model(xunp, coords)
            xper, _ = self.model(xper, coords)

            dx = xper - xunp

            # Replace zero elements with 1 in hemispheric norm, we do this because when
            # hem norm is zero we know dx is 0, so this just makes the next line computable
            hem_norm[hem_norm == 0] = 1
            # scale dx
            dx = self.noise_amplitude * (dx / hem_norm)

            xunp = (
                input_data[ii + 1]
                .unsqueeze(dim=0)
                .repeat(batch_size, 1, 1, 1, 1, 1)
                .to(device)
            )
 ```

So you just set the hem_norm to 1 for any zero value, since you know if its zero then dx is zero. I ran a small test and seemed to produce the same result as whats currently there. Thoughts?

**Comment by NickGeneva** [comment]
Snippet I was testing with:

```python
import numpy as np
import torch
from datetime import datetime
from typing import Generator, Iterator
from earth2studio.perturbation import HemisphericCentredBredVector, Gaussian
from earth2studio.models.batch import batch_func
from earth2studio.utils.type import CoordSystem


class TestPrognostic:

    def input_coords(self):
        return OrderedDict({
            "batch": np.empty(0),
            "lead_time": np.array([np.timedelta64(0, "h")]),
            "variable": np.array(["t2m", "tcwv"]),
            "lat": np.linspace(-90, 90, 31),
            "lon": np.linspace(0, 360, 60, endpoint=False)
        })

    def output_coords(self, input_coords):
        return OrderedDict({
            "batch": input_coords['batch'],
            "lead_time": input_coords["lead_time"] + np.array([np.timedelta64(6, "h")]),
            "variable": np.array(["t2m", "tcwv"]),
            "lat": np.linspace(-90, 90, 31),
            "lon": np.linspace(0, 360, 60, endpoint=False)
        })

    def _forward(self, x, coords):
        coords = self.output_coords(coords)
        return x, coords

    @batch_func()
    def __call__(
        self,
        x: torch.Tensor,
        coords: CoordSystem,
    ) -> tuple[torch.Tensor, CoordSystem]:
        return self._forward(x, coords)
    
    @batch_func()
    def _default_generator(
        self, x: torch.Tensor, coords: CoordSystem
    ) -> Generator[tuple[torch.Tensor, CoordSystem], None, None]:
        coords = coords.copy()

        self.output_coords(coords)
        yield x, coords

        while True:
            # Front hook
            x, coords = self.front_hook(x, coords)

            # Forward is identity operator
            x, coords = self._forward(x, coords)

            # Rear hook
            x, coords = self.rear_hook(x, coords)
            yield x, coords

    def create_iterator(
        self, x: torch.Tensor, coords: CoordSystem
    ) -> Iterator[tuple[torch.Tensor, CoordSystem]]:
        yield from self._default_generator(x, coords)

torch.manual_seed(42)

pt = HemisphericCentredBredVector(
    model=TestPrognostic(),
    data=GFS(),
    noise_amplitude=0.1,
    integration_steps=2,  # use cfg.breeding_steps
    seeding_perturbation_method=Gaussian(noise_amplitude=torch.Tensor([0,1]).unsqueeze(-1).unsqueeze(-1)),
)

coords = OrderedDict({
    "time": np.asarray([datetime(2024,1,1)]),
    "lead_time": np.array([np.timedelta64(0, "h")]),
    "variable": np.array(["t2m", "tcwv"]),
    "lat": np.linspace(-90, 90, 31),
    "lon": np.linspace(0, 360, 60, endpoint=False)
})
x = torch.randn(1,1,1,31,60)

out, coord = pt(x, coords)
```

earth2studio/perturbation/hcbv.py
NickGeneva: (if the answer is no, then just let me know and we can merge. whats there looks good, just trying to simplify is possible)

rodrigoalmeida94: Ah yes that is much cleaner than handling messy nan values. I like this better too!

NickGeneva (approved):

NickGeneva:
Thanks again for the PR @rodrigoalmeida94 !

Issue #283, opened 2025-04-30
🚀[FEA]: Async Zarr IO


NickGeneva:
### Is this a new feature, an improvement, or a change to existing functionality?

New Feature

### How would you describe the priority of this feature request

Medium

### Please provide a clear description of problem you would like to solve.

Right now the Zarr io object is pretty in efficient which can be an issue for inference jobs. With Zarr 3.0 we should migrate to async IO objects and check performance more rigorously.

Related convo: https://github.com/NVIDIA/earth2studio/pull/271

NickGeneva:
https://github.com/NickGeneva/earth2studio/tree/ngeneva/io_rework

PR #377, opened 2025-06-12
Non-Blocking Async Zarr IO

NickGeneva:
# Earth2Studio Pull Request

## Description

Fixes random bug in DLWP with lead time coord being a scalar array.

Adds a new async zarr data source that has non-blocking write call capabilities.
It functions quite different than existing ones both to ensure thread safety but also help make initialization easier.  Have a look at the unit tests to see how things work. Some of the inputs arent super trivial, but simple enough I think.

Tested for in-memory, file store and s3. But should (likely) work with other fsspec file systems.

Likely needs an example or something... not sure. Open to suggestions.

Closes: https://github.com/NVIDIA/earth2studio/issues/283

TODO: Read controls? Not trivial due to async usage. Might be best to tell people to just open the zarr store seperately

Heres the readme example

```python
import numpy as np

from earth2studio.models.px import DLWP
from earth2studio.io import AsyncZarrBackend
from earth2studio.run import deterministic as run

model = DLWP.load_model(DLWP.load_default_package())
ds = GFS()

nsteps = 10
times = [datetime(2024,1,1)]
index_coords = {
    "time": np.array(times),
    "lead_time": np.array([timedelta(hours=6*i) for i in range(nsteps)])
}
io = AsyncZarrBackend("output.zarr", index_coords)

run(times, nsteps, model, ds, io)
```

Heres a little sample script that just builds a little HRRR dataset

```python
import functools
import os
import warnings
import numpy as np

import s3fs
from earth2studio.data import ForecastSource
from earth2studio.data.hrrr import HRRR_FX
from earth2studio.io import IOBackend, AsyncZarrBackend

from earth2studio.data import prep_data_array
from earth2studio.utils.coords import split_coords

warnings.simplefilter(action='ignore', category=FutureWarning)

def fetch(source: ForecastSource, iob: IOBackend, 
    times: list[datetime], 
    lead_times: list[timedelta],
    variables: list[str]
) -> None:
    for date0 in times:
        data = source([date0], lead_times, variables)
        x, coords = prep_data_array(data)
        x, coords, array_names = split_coords(x, coords)
        iob.write(x, coords, array_names.tolist())

    iob.close()
         
if __name__ == "__main__":

    fs_factory = functools.partial(
        s3fs.S3FileSystem,
        key=os.environ["S3FS_CI_KEY"],
        secret=os.environ["S3FS_CI_SECRET"],
        client_kwargs={"endpoint_url": os.environ.get("S3FS_CI_ENDPOINT", None)},
        asynchronous=True,
    )
    times = [(datetime(2021, 1, 1) + timedelta(hours=3*i)) for i in range(60)]
    lead_time = [timedelta(hours=1*i) for i in range(8)]
    index_coords = {
        "time": np.array(times, dtype=np.datetime64),
        "lead_time": np.array(lead_time, dtype=np.timedelta64)
    }
    iob = AsyncZarrBackend(
        "earth2studio/data_hrrr.zarr", index_coords=index_coords, fs_factory=fs_factory, blocking=False
    )
    source = HRRR_FX(cache=False)

    fetch(source, iob, times[:4], lead_time, ["t2m", "tcwv"])
```

## Checklist

### DIFF
--- a/test/io/test_async_zarr.py
+++ b/test/io/test_async_zarr.py
@@ -354,7 +354,7 @@ async def test_async_zarr_split_variables(
 
 
 @pytest.mark.slow
-@pytest.mark.timeout(60)
+@pytest.mark.timeout(120)
 @pytest.mark.parametrize("blocking", [True, False])
 @pytest.mark.parametrize("device", ["cpu", "cuda:0"])
 @pytest.mark.skipif(
@@ -394,8 +394,8 @@ def test_async_zarr_remote(
         {
             "time": np.asarray(times),
             "variable": variable,
-            "lat": np.linspace(-90, 90, 16),
-            "lon": np.linspace(0, 360, 64, endpoint=False),
+            "lat": np.linspace(-90, 90, 8),
+            "lon": np.linspace(0, 360, 16, endpoint=False),
         }
     )
     shape = [v.shape[0] for v in total_coords.values()]


earth2studio/io/base.py
```diff
@@ -25,11 +25,6 @@
 class IOBackend(Protocol):
     """Interface for a generic IO backend."""
 
-    def __init__(
```
NickGeneva, line 28: Removed this because I think unused / not needed for protocol. Linter was complaining

earth2studio/models/px/dlwp.py
```diff
@@ -403,15 +403,15 @@ def _default_generator(
 
             # Rear hook for first predicted step
             coords_out = coords.copy()
-            coords_out["lead_time"] = coords["lead_time"][0]
+            coords_out["lead_time"] = coords["lead_time"][0:1]
```
NickGeneva, line 406: Bug fix

earth2studio/io/async_zarr.py
```diff
@@ -0,0 +1,639 @@
...
+        Index coordinates. These are coordinates that will be used to define chunks in
+        each zarr array. Each element in these coordinates will be written in parallel.
+        Typically index coordinates reflect the dimensions that are being iteratively
+        generated during inference, such as time or lead_time. The remaining coordinates
+        of a given array will be populated upon the first write to the array, by default {}
+    fs_factory : Callable[..., fsspec.spec.AbstractFileSystem], optional
+        FSSpec file system factory method. This is a callable object that should return
+        an instance of the desired filesystem to use, by default LocalFileSystem
+    blocking : bool, optional
+        Blocking write calls in the synchronous API. When set to fall, the IO backend
```
dallasfoster: "When set to fall" -> "When set to false"

earth2studio/io/async_zarr.py
```diff
@@ -0,0 +1,639 @@
...
+    Parameters
+    ----------
+    file_name : str
+        Path location to place zarr store
+    index_coords : CoordSystem, optional
+        Index coordinates. These are coordinates that will be used to define chunks in
+        each zarr array. Each element in these coordinates will be written in parallel.
+        Typically index coordinates reflect the dimensions that are being iteratively
+        generated during inference, such as time or lead_time. The remaining coordinates
+        of a given array will be populated upon the first write to the array, by default {}
```
dallasfoster: Just the doc string is confusing. "These are coordinates that will be used to define chunks". What does that mean? What are the chunk sizes? Shouldn't there be a separate `chunks` argument?

earth2studio/io/async_zarr.py
```diff
@@ -0,0 +1,639 @@
...
+        Blocking write calls in the synchronous API. When set to fall, the IO backend
+        will execute write calls in seperate threads. Users should call the `close()`
+        API to ensure all threads have finished / cleaned up, by default True
+    pool_size : int, optional
+        The thread / async loop pool used with the synchronous write API in non-blocking
+        mode, by default 4
+    async_timeout : int, optional
+        Async operation timeout for a given write operation, by default 600
+    zarr_kwargs : dict[str, Any], optional
+        Additional key work arguments to provide to the ` zarr.api.asynchronous.open`
```
dallasfoster: "key work" -> "keyword"

earth2studio/io/async_zarr.py
```diff
@@ -0,0 +1,639 @@
...
+    torch.int64: np.int64,
+    torch.float16: np.float16,
+    torch.float32: np.float32,
+    torch.float64: np.float64,
+    torch.complex64: np.complex64,
+    torch.complex128: np.complex128,
+}
+
+
+class AsyncZarrBackend:
```
dallasfoster, line 54: What happens if we have Zarr v2?

dallasfoster: I see the exception below, maybe add that to the documentation.

```
Raises
-------
ImportError ...
```

test/io/test_async_zarr.py
```diff
@@ -0,0 +1,590 @@
...
+    assert np.allclose(data, x.to("cpu").numpy())
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "fs_factory",
+    [MemoryFileSystem, LocalFileSystem],
+)
+@pytest.mark.parametrize("device", ["cpu", "cuda:0"])
+async def test_async_zarr_async_write(
```
pzharrington, line 129: Seems mostly duplicated from the prev test, maybe could just parameterize whether or not to make the `write` async?

earth2studio/io/async_zarr.py
```diff
@@ -0,0 +1,639 @@
...
+        self.blocking = blocking
+        if blocking:
+            pool_size = 1
+        self.async_timeout = async_timeout
+        self.io_futures: list[concurrent.futures._base.Future] = []
+        self.pool_index = 0
+        self.loop_pool = self._initialize_loop_pool(pool_size)
+        self.fs_pool = []
+        self.zarr_pool = []
+        logger.warning(
```
pzharrington: nit: does this need to be a warning or could it just be debug? Would add to chatter in multi-process runs

earth2studio/io/async_zarr.py
```diff
@@ -0,0 +1,639 @@
...
+        Raises
+        ------
+        ValueError
+            If some coords are index coords and container new values no in self.index_coords
+        """
+        # DOESNT WORK????
+        # current_arrays = [key async for key in self.root.array_keys()]
+        # ======
+        # Coordinate arrays
+        # ======
```
pzharrington, line 283: What do the above comments mean?

pzharrington:
I think a more complete example than the demos you pasted above would be helpful. Something that shows the different speeds in blocking vs. non-blocking mode, and the different ways `write` could be used (splitting by variable, etc) along with how to define and use `index_coords`.

earth2studio/io/async_zarr.py
pzharrington: As is, looks like the chunk size is 1 along each `index_coord` (which I assume is a threadsafety requirement) and full-size everywhere else. Might be nice to support controlling the chunk size of other non-index dims. IMO the name "index_coord" is a bit confusing but can't really think of a better one

NickGeneva: Right, I couldn't think of a better name. I'll think about it a little more or ask a LLM lol.
Let me see about potentially supporting custom chunking in other dims... I don't have anything at the moment because I wanted to make sure the chunks were safe for the threading / writing.

I may put in some secret property to enable the control of the other dimensions, just not expose it immediately until someone complains.

earth2studio/io/async_zarr.py
NickGeneva: Good call, fixed!

earth2studio/io/async_zarr.py
NickGeneva: Agreed, added!

earth2studio/io/async_zarr.py
NickGeneva: Good catch, fixed. Thanks!

earth2studio/io/async_zarr.py
NickGeneva: I renamed to `parallel_coords` per the advise of an llm... and redid the doc string.

There is also a internal property now `chunked_coords` that someone can set post initialization to better control how the arrays get chunked. I dont expose it in the constructor though since it people need to be careful with it.

earth2studio/io/async_zarr.py
NickGeneva: Nice catch, fixed. Thanks!

earth2studio/io/async_zarr.py
NickGeneva: Left over from debugging removed, thanks!

test/io/test_async_zarr.py
NickGeneva: I've updated the first one to not be `@pytest.mark.asyncio` so its a sync test with no loop initialized. Then this test is a async test that already has a loop initialized.

PR #421, opened 2025-07-16
Add Downscaling feature to HENS recipe

albertocarpentieri:
- Add example script to run corrdiff on HENS forecasts
- Add corrdiff as optional diagnostic model to the HENS recipe:
     - multiple corrdiff models can be added (dict structure as diagnostic models)
     - each corrdiff model output is saved on independent netcdf files
     - support to run multiple corrdiff ensemble members per HENS ensemble member

### DIFF
--- a/examples/18_corrdiff_hens.py
+++ b/examples/18_corrdiff_hens.py
@@ -15,6 +15,29 @@
 # limitations under the License.
 
 # %%
+"""
+CorrDiff on HENS Ensemble
+=========================
+
+This example shows how to run CorrDiff on a HENS ensemble.
+
+In this example you will learn:
+
+- How to run CorrDiff on a HENS ensemble
+- How to save the output to a Zarr backend
+- How to post process the output
+
+"""
+# /// script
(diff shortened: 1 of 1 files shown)


albertocarpentieri:
@NickGeneva @dallasfoster

recipes/hens/src/hens_ensemble.py
```diff
@@ -138,9 +158,6 @@ def __init__(
 
     def move_models_to_device(
```
dallasfoster, line 167: This change breaks device movement so please try to revert some it back. In particular

```python
if self.device is not None:
    self.device = (...)
``` 
is not right, it should read 
```python
if self.device is None:
    self.device = (...)
```

Can you also add in the ability to set the device from the distributed manager? Something like this:
```python
if self.device is None:
            self.device = (
                torch.device(
                    DistributedManager().device if torch.cuda.is_available() else "cpu"
                    )
            )
```

dallasfoster (approved):

PR #424, opened 2025-07-18
Use ocean output times for ocean coupling

pzharrington:
# Earth2Studio Pull Request

## Description
Uses the length of ocean output times to chunk ocean coupled inputs, rather than input times. Does not change behavior of current default package since the output times are same length as the input times, but this should be more flexible/correct for custom or future DLESyM variants.

## Checklist

### DIFF
--- a/earth2studio/models/px/dlesym.py
+++ b/earth2studio/models/px/dlesym.py
@@ -679,7 +679,7 @@ class DLESyM(torch.nn.Module, AutoModelMixin, PrognosticMixin):
 
         # Slice along the lead_time dim, average, and stack along the variable dim
         # (different time-averaged quantities are defined as different variables)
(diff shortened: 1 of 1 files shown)


NickGeneva (approved):
Approving, but would like to know how this managed to not be caught in the unit tests.

pzharrington:
Well either way is correct for the current DLESyM version (using `len(ocean_input_times)` or `len(ocean_output_times)` in that line), so it had no effect on current behavior really. There are no tests that try instantiating a DLESyM wrapper with non-standard config, the unit tests for models typically test the APIs for call/iter/exception/package load only for the standard/default args. To be honest I'm not even sure it makes sense to run dlesym inference with `len(ocean_output_times) != len(ocean_input_times)` since ocean is the slowest component

NickGeneva:
Ah if this is a non-standard case, then not a big deal. All good. Thanks for clarification

PR #455, opened 2025-08-15
Updates to CorrDiffTaiwan model wrapper

CharlelieLrt:
# Earth2Studio Pull Request

## Description
This PR updates the CorrDiffTaiwan model wrapper with latest APIs and performance optimizations. It should bring up to 10x inference speed (_at best_) thanks to Apex group norm and compilation.

## Checklist

### DIFF
--- a/earth2studio/models/dx/corrdiff.py
+++ b/earth2studio/models/dx/corrdiff.py
@@ -880,7 +880,7 @@ class CorrDiffTaiwan(torch.nn.Module, AutoModelMixin):
 
     @classmethod
     @check_optional_dependencies()
-    def load_model(cls, package: Package) -> DiagnosticModel:
+    def load_model(cls, package: Package, device: str | None = None) -> DiagnosticModel:
         """Load diagnostic from package"""
 
         if StackedRandomGenerator is None or deterministic_sampler is None:
@@ -897,15 +897,33 @@ class CorrDiffTaiwan(torch.nn.Module, AutoModelMixin):
             str(
                 checkpoint_zip.parent
                 / Path("corrdiff_inference_package/checkpoints/diffusion.mdlus")
-            )
-        ).eval()
+            ),
+            override_args={"use_apex_gn": True}
+        )
+        residual.use_fp16, residual.profile_mode = True, False
+        residual = residual.eval()
+        if device is not None:
+            residual = residual.to(device)
+        residual = residual.to(memory_format=torch.channels_last)
 
         regression = PhysicsNemoModule.from_checkpoint(
             str(
                 checkpoint_zip.parent
                 / Path("corrdiff_inference_package/checkpoints/regression.mdlus")
-            )
-        ).eval()
+            ),
+            override_args={"use_apex_gn": True}
+        )
+        regression.use_fp16, regression.profile_mode = True, False
+        regression = regression.eval()
+        if device is not None:
+            regression = regression.to(device)
+        regression = regression.to(memory_format=torch.channels_last)
+
+        # Compile models
+        torch._dynamo.config.cache_size_limit = 264
+        torch._dynamo.reset()
+        residual = torch.compile(residual)
+        regression = torch.compile(regression)
 
         # Get dataset for lat/lon grid info and centers/stds'
         try:
@@ -1011,10 +1029,11 @@ class CorrDiffTaiwan(torch.nn.Module, AutoModelMixin):
         x = self._interpolate(x)
 
         # Add sample dimension
-        x = x.unsqueeze(0)
+        x = x.unsqueeze(0).to(memory_format=torch.channels_last)
         x = (x - self.in_center) / self.in_scale
 
         # Create grid channels
+        # TODO: why do we need this grid concatenated to the input?
         x1 = np.sin(np.linspace(0, 2 * np.pi, 448))
         x2 = np.cos(np.linspace(0, 2 * np.pi, 448))
         y1 = np.sin(np.linspace(0, 2 * np.pi, 448))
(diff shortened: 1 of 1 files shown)


CharlelieLrt:
@NickGeneva a few questions on the CorrDiffTaiwan wrapper:
- I do not understand why we need to compute a grid and concatenate it to the input before the forward pass. It is not done in PhysicsNeMo.
- Currently physicsNeMo uses the stochatsic_sampler by default, but the wrapper uses the deterministic sampler. I don't think that's a problem, but the stochatsic_sampler should be provided as an option at least.
LMK what you think.

earth2studio/models/dx/corrdiff.py
```diff
@@ -880,7 +880,7 @@ def load_default_package(cls) -> Package:
 
     @classmethod
     @check_optional_dependencies()
-    def load_model(cls, package: Package) -> DiagnosticModel:
+    def load_model(cls, package: Package, device: str | None = None) -> DiagnosticModel:
```
NickGeneva, line 891: The model wrappers are modules themselves, so users are expected to use wrapper.to(device).

So load_model (unless its gpu only model) just places things on the cpu and then users move it to the device. Can this be done with the optimization settings here?

earth2studio/models/dx/corrdiff.py
```diff
@@ -1011,10 +1029,11 @@ def _forward(self, x: torch.Tensor) -> torch.Tensor:
         x = self._interpolate(x)
 
         # Add sample dimension
-        x = x.unsqueeze(0)
+        x = x.unsqueeze(0).to(memory_format=torch.channels_last)
         x = (x - self.in_center) / self.in_scale
 
         # Create grid channels
+        # TODO: why do we need this grid concatenated to the input?
```
NickGeneva: I think because the original corrdiff had a spatial encoding which was this grid

earth2studio/models/dx/corrdiff.py
```diff
@@ -1032,43 +1051,44 @@ def _forward(self, x: torch.Tensor) -> torch.Tensor:
...
         x = torch.cat((x, grid), dim=1)
 
-        # Repeat for sample size
-        sample_seeds = torch.arange(self.number_of_samples)
-        x = x.repeat(self.number_of_samples, 1, 1, 1)
+        # Create seeds for each sample
+        seed = self.seed if self.seed is not None else np.random.randint(2**32)
+        gen = torch.Generator(device=x.device)
+        gen.manual_seed(seed)
+        sample_seeds = torch.randint(0, 2**32, (self.number_of_samples,), device=x.device, generator=gen)
```
NickGeneva: instead of doing `np.random.randint(2**32)` for if seed is None, just passing in None as the generator should be fine if theres no seed set

NickGeneva:
Can you also add a line in the changelog about these updates? Thanks!

earth2studio/models/dx/corrdiff.py
CharlelieLrt: Yes, I could do that and move the `.to(device)` in the Triton server instead. There's however one caveat: the `.to(memory_format=channels_last)` has to be done _after_ the `.to(device)`. So the user code (and here the Triton server) would have the respnsability of doing both the `.to(device)` and the `.to(memory_formet=channels_last)`. Would that be okay?

dallasfoster: I think this is one of the few instances where it might be beneficial to allow the user to specify the device here.

CharlelieLrt: I'll leave the `device` argument here for now

earth2studio/models/dx/corrdiff.py
CharlelieLrt: Ok. I'm not sure with which codebase was the "original corrdiff" trained but physicsnemo training does not do that. So, I think this does not work with newer checkpoints trained with physicsnemo. I'll leave it as it is for now, as I'm not sure how to support both.

earth2studio/models/dx/corrdiff.py
CharlelieLrt: Changed to `gen=None` when `seed` is `None`

earth2studio/models/dx/corrdiff.py
```diff
@@ -1018,43 +1037,44 @@ def _forward(self, x: torch.Tensor) -> torch.Tensor:
         # Concat Grids
         x = torch.cat((x, grid), dim=1)
 
-        # Repeat for sample size
-        sample_seeds = torch.arange(self.number_of_samples)
-        x = x.repeat(self.number_of_samples, 1, 1, 1)
+        # Create seeds for each sample
+        seed = self.seed if self.seed is not None else np.random.randint(2**32)
```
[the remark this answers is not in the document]
CharlelieLrt: Added the seed attribute in the `__init__`

earth2studio/models/dx/corrdiff.py
```diff
@@ -1018,43 +1037,44 @@ def _forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        W_hr = coord["lon"].shape[1]
 
-        mean = self.unet_regression(
+        mean_hr = self.unet_regression(
             self.regression_model,
-            torch.zeros_like(latents),
-            x,
-            num_steps=self.number_of_steps,
+            img_lr=x,
+            output_channels=len(self.output_variables),
```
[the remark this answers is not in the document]
CharlelieLrt: Added the attribute `self.output_variables = OUT_VARIABLES`

earth2studio/models/dx/corrdiff.py
```diff
@@ -1018,43 +1037,44 @@ def _forward(self, x: torch.Tensor) -> torch.Tensor:
...
-            x,
-            randn_like=rnd.randn_like,
-            num_steps=self.number_of_steps,
-            solver=self.solver,
+
+        res_hr = diffusion_step(
+            net=self.residual_model,
+            sampler_fn=sampler_fn,
+            img_shape=(H_hr, W_hr),
+            img_out_channels=len(self.output_variables),
```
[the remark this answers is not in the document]
CharlelieLrt: Added the attribute `self.output_variables = OUT_VARIABLES`

earth2studio/models/dx/corrdiff.py
```diff
@@ -1018,43 +1037,44 @@ def _forward(self, x: torch.Tensor) -> torch.Tensor:
...
-        mean = self.unet_regression(
+        mean_hr = self.unet_regression(
             self.regression_model,
-            torch.zeros_like(latents),
-            x,
-            num_steps=self.number_of_steps,
+            img_lr=x,
+            output_channels=len(self.output_variables),
+            number_of_samples=self.number_of_samples,
         )
```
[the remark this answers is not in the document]
CharlelieLrt: Removed `lead_time_label` as it is not used for Taiwan CorrDiff

earth2studio/models/dx/corrdiff.py
```diff
@@ -1018,43 +1037,44 @@ def _forward(self, x: torch.Tensor) -> torch.Tensor:
...
-            randn_like=rnd.randn_like,
-            num_steps=self.number_of_steps,
-            solver=self.solver,
+
+        res_hr = diffusion_step(
+            net=self.residual_model,
+            sampler_fn=sampler_fn,
+            img_shape=(H_hr, W_hr),
+            img_out_channels=len(self.output_variables),
+            rank_batches=sample_seeds,
```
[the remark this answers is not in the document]
CharlelieLrt: Changed to `rank_batches=[sample_seeds]`, which is suitable for single rank

Issue #459, opened 2025-08-20
🚀[FEA]: FCN3 surface pressure


jleinonen:
### Is this a new feature, an improvement, or a change to existing functionality?

Improvement

### How would you describe the priority of this feature request

Medium

### Please provide a clear description of problem you would like to solve.

FourCastNet3 (FCN3) doesn't provide the surface pressure (`'sp'`) variable due to it causing training instability. However, this variable is important in weather analysis and needed by many downstream applications. Thus a method to interpolate/extrapolate the surface pressure from the pressure levels would be useful.

I'm planning to submit a PR to add this - creating an issue to notify developers ahead of time.

PR #470, opened 2025-08-27
Surface pressure interpolation + Diagnostic model wrapper

jleinonen:
# Earth2Studio Pull Request

## Description

Adds a derived diagnostic model `DerivedSurfacePressure` to interpolate surface pressure from the pressure levels, as outlined in https://github.com/NVIDIA/earth2studio/issues/459.

Additionally, adds a generic wrapper to bundle variables from diagnostic models into the output of a prognostic model. The amended prognostic model follows the `PrognosticModel` interface and can therefore be used in any workflow that expects a prognostic model.

Below is an example of using the two new components to add surface pressure and precipitation from `PrecipitationAFNOv2` (which requires surface pressure) to the output of FCN3:

```python

import numpy as np
import torch
import xarray as xr

from earth2studio.data import fetch_data, GFS
from earth2studio.models.px import DiagnosticWrapper, FCN3
from earth2studio.models.dx import DerivedSurfacePressure, PrecipitationAFNOv2
from earth2studio.utils.coords import map_coords

data = GFS()

fcn3_package = FCN3.load_default_package()
px_model = FCN3.load_model(fcn3_package)

orography_fn = fcn3_package.resolve("orography.nc")
with xr.open_dataset(orography_fn) as ds:
    z_surface = torch.as_tensor(ds["Z"][0].values)
z_surf_coords = OrderedDict({
    d: px_model.input_coords()[d] for d in ["lat", "lon"]
})
sp_model = DerivedSurfacePressure(
    p_levels=[50, 100, 150, 200, 250, 300, 400, 500, 600, 700, 850, 925, 1000],
    surface_geopotential=z_surface,
    surface_geopotential_coords=z_surf_coords
)

precip_model = PrecipitationAFNOv2.load_model(PrecipitationAFNOv2.load_default_package())

wrapped_model = DiagnosticWrapper(px_model=px_model, dx_models=[sp_model, precip_model])
wrapped_model.to(device='cuda')

(x, coords) = fetch_data(
    data,
    [np.datetime64("2025-08-21T00:00:00")],
    variable=wrapped_model.input_coords()["variable"],
    device=device
)
(x, coords) = map_coords(x, coords, wrapped_model.input_coords())

(x, coords) =  wrapped_model(x, coords)
```

## Checklist


earth2studio/models/dx/derived.py
```diff
@@ -393,3 +393,224 @@ def __call__(
...
+        `corr_adjustment=(0.0, 1.0)`.
+    """
+
+    Rs = 287.053  # gas constant for dry air (J/kg/K)
+    g = 9.8067  # Earth's gravitational constant (m/s**2)
+    L = -6.5e-3  # average temperature lapse rate (K/m)
+
+    def __init__(
+        self,
+        p_levels: list[int],
```
dallasfoster: Is it possible to do either `torch.Tensor` or `list[int] | np.array | torch.Tensor`?

earth2studio/models/dx/derived.py
```diff
@@ -393,3 +393,224 @@ def __call__(
...
+    ) -> None:
+        super().__init__()
+        self.temperature_correction = temperature_correction
+        self.corr_adjustment = corr_adjustment
+
+        p_levels = sorted(p_levels, reverse=True)
+        self.in_variables = [f"z{level}" for level in p_levels]
+        if self.temperature_correction:
+            self.in_variables.extend(f"t{level}" for level in p_levels)
+        self.out_variables = ["sp"]
```
dallasfoster: Can we just save these as np arrays to avoid the casting below?

earth2studio/models/dx/derived.py
```diff
@@ -393,3 +393,224 @@ def __call__(
...
+    @torch.inference_mode()
+    @batch_func()
+    def __call__(
+        self,
+        x: torch.Tensor,
+        coords: CoordSystem,
+    ) -> tuple[torch.Tensor, CoordSystem]:
+        """Forward pass of diagnostic"""
+        num_levels = len(self.log_p_levels)
+
```
dallasfoster, line 601: Normally we would want a coordinate check at the beginning.

earth2studio/models/dx/derived.py
```diff
@@ -393,3 +393,224 @@ def __call__(
...
+        # unpack inputs, swap variable dimension to dim 0, flatten batch dims to dim 1
+        variable_dim = list(coords).index("variable")
+        z_levels = x[:, :num_levels].transpose(0, variable_dim)
```
dallasfoster, line 604: Without a coordinate check at the beginning we can't be sure that `x[:, :num_levels]` is doing the right thing.

earth2studio/models/px/dxwrapper.py
```diff
@@ -0,0 +1,268 @@
...
+      must be mappable to the next model using `earth2studio.utils.coords.map_coords`.
+    - If interpolate_coords == True, the coordinates of each model in the chain must
+      be possible to interpolate to the next model using
+      `earth2studio.utils.interp.LatLonInterpolation`.
+
+    Parameters
+    ----------
+    px_model : PrognosticModel
+        The prognostic model to use as the base model.
+    dx_models : List[DiagnosticModel]
```
dallasfoster: This doc doesn't mass the model API below.

earth2studio/models/px/dxwrapper.py
```diff
@@ -0,0 +1,268 @@
...
+        # check variable compatibility
+        px_output_vars = self.px_model.output_coords(self.px_model.input_coords())[
+            "variable"
+        ]
+        dx_input_vars = self.dx_model.input_coords()["variable"]
+        missing_vars = list(set(dx_input_vars) - set(px_output_vars))
+        if missing_vars:
+            raise ValueError(
+                f"The following inputs of the diagnostic model are not outputs of the prognostic model: {missing_vars}"
+            )
```
dallasfoster: Why is this needed over something like `handshake_dim` or `handshake_coord`?

earth2studio/models/px/dxwrapper.py
```diff
@@ -0,0 +1,268 @@
...
+            in_coords_dx["lon"] = out_coords_px["lon"]
+        out_coords_dx = self.dx_model.output_coords(in_coords_dx)
+
+        if self.keep_px_output:
+            variables = np.concatenate(
+                [out_coords_px["variable"], out_coords_dx["variable"]]
+            )
+            out_coords_dx["variable"] = variables
+
+        return out_coords_dx
```
dallasfoster: I am a little concerned about this coordinate handling here. First, I don't see any real coordinate checking occurring here and I am thinking of the following concrete scenarios:

1. What if the prognostic model was StormCast and the diagnostic model was precipitation afno? Do I get an error because of the domain mismatch?
2. What if the prognostic model was FCN and the diagnostic model was a tropical cyclone tracker? What does the output coordinates look like? Do I get an error? Should I?
3. What if the prognostic model was SFNO and the diagnostic model was CorrDiff? What happens with the sample coordinate dimension? How are the px outputs handled? Is this desirable?

test/models/px/test_dxwrapper.py
```diff
@@ -0,0 +1,159 @@
...
+    wrapped_model = DiagnosticWrapper(
+        px_model=px_model, dx_models=[sp_model, precip_model]
+    ).to(device=device)
+
+    dc = {k: wrapped_model.input_coords()[k] for k in ["lat", "lon"]}
+    data = Random(dc)
+
+    (x, coords) = fetch_data(
+        data,
+        [np.datetime64("2025-08-21T00:00:00")],
```
dallasfoster: Can we parametrize the input time to also include a test with more than one timestamp?

test/models/px/test_dxwrapper.py
```diff
@@ -0,0 +1,159 @@
...
+        variable=px_model.input_coords()["variable"],
+        device=device,
+    )
+    (x, coords) = map_coords(x, coords, wrapped_model.input_coords())
+
+    (x, coords) = wrapped_model(x, coords)
+
+    expected_shape = tuple(len(v) for v in coords.values())
+    assert x.shape == expected_shape == (1, 1, 1, 4, 448, 448)
```
dallasfoster: It would be nice to see a case where num_samples > 1.

earth2studio/models/dx/derived.py
jleinonen: Changed to `list[int] | np.array | torch.Tensor` and added handling in https://github.com/NVIDIA/earth2studio/pull/470/commits/1022cc29bffdfd4743ef5851303b5fe864fd5264.

earth2studio/models/dx/derived.py
jleinonen: Moved casting here in https://github.com/NVIDIA/earth2studio/pull/470/commits/1022cc29bffdfd4743ef5851303b5fe864fd5264.

earth2studio/models/dx/derived.py
jleinonen: Moved calling `self.output_coords(coords)` from the end of the function to the beginning in https://github.com/NVIDIA/earth2studio/pull/470/commits/1022cc29bffdfd4743ef5851303b5fe864fd5264. I think this takes care of it?

earth2studio/models/dx/derived.py
jleinonen: I think moving `self.output_coords(coords)` to the beginning fixes this too?

earth2studio/models/px/dxwrapper.py
jleinonen: Fixed the docstring in https://github.com/NVIDIA/earth2studio/pull/470/commits/1022cc29bffdfd4743ef5851303b5fe864fd5264.

earth2studio/models/px/dxwrapper.py
jleinonen: I think this can be removed. I thought it was good to have as a more explicit error message but I guess it can also add errors in cases like where the inputs and outputs of a model change after creation. Removed in https://github.com/NVIDIA/earth2studio/pull/470/commits/a58e04e0816cd58cc9d3e39aeb08d45231be48da.

test/models/px/test_dxwrapper.py
jleinonen: Added in https://github.com/NVIDIA/earth2studio/pull/470/commits/4a8e0c5ed44735eec6eed8940357745b11270700.

test/models/px/test_dxwrapper.py
jleinonen: Added in https://github.com/NVIDIA/earth2studio/pull/470/commits/4a8e0c5ed44735eec6eed8940357745b11270700.

earth2studio/models/px/dxwrapper.py
jleinonen: Case 3 is addressed equivalently by the `test_fcn3_corrdiff` test, right? So SFNO+CorrDiff works but you have to set `keep_px_output=False`. In order to use `keep_px_output=True`, the input of the diagnostic model must be possible to concatenate to the output, which is not the case for CorrDiff.

In my view, case 1 should be an error (since there is no way to regrid StormCast output to Precipitation AFNO input grid). Case 2 should work if `keep_px_output=False`. Let me investigate these cases.

jleinonen: https://github.com/NVIDIA/earth2studio/pull/470/commits/d33286822d8c342103bc00a63354a385572c1309 adds a test for case 2 and ensures that case 3 fails with an understandable error message if `keep_px_outputs==True`.

Case 1 is tricky because interpolation may produce an output if you try to interpolate from the StormCast grid to the diagnostic model grid. Therefore Precipitation AFNO may be able to run but the outputs probably will not make sense, at least not outside the StormCast domain. Any suggestions how to handle this situation?

earth2studio/models/dx/derived.py
```diff
@@ -393,3 +393,227 @@ def __call__(
...
+
+class DerivedSurfacePressure(torch.nn.Module):
+    """Interpolates the surface pressure in hPa from pressure level geopotential,
+    surface geopotential and optionally temperature. The calculation is based on
+    linear interpolation of the logarithm of pressure, as well as an optional
+    second-order correction based on temperature and an empirical adjustment.
+
+    Note
+    ----
+    TODO: add link here for description of the method
```
NickGeneva: Missing link?

jleinonen: Added in https://github.com/NVIDIA/earth2studio/pull/470/commits/82c5774f950b85b96734b0df5a9d11644af22ccf.

earth2studio/models/px/dxwrapper.py
dallasfoster: Case 1 should definitely fail with a domain mismatch.

dallasfoster (approved):

PR #473, opened 2025-08-28
Inference output dataset source

NickGeneva:
# Earth2Studio Pull Request

## Checklist

### DIFF
--- a/earth2studio/data/xr.py
+++ b/earth2studio/data/xr.py
@@ -93,33 +93,32 @@ class InferenceOuputSource:
     filter_dict : dict, optional
         Dictionary of selections applied before transformation (e.g.
         ``{"ensemble": 0}``). Coordinates not in the required
-        dimensions are dropped after selection, by default None
+        dimensions are dropped after selection, by default {}
     **xr_args : Any
         Additional keyword arguments forwarded to ``xarray.open_dataset``.
     """
 
-    def __init__(self, file_path: str, filter_dict: dict = None, **xr_args: Any):
+    def __init__(self, file_path: str, filter_dict: dict = {}, **xr_args: Any):
         self.file_path = file_path
         self.da = xr.open_dataset(self.file_path, **xr_args)
         self.da = self.da.to_array("variable")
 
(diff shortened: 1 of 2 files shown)


earth2studio/data/xr.py
```diff
@@ -62,8 +62,117 @@ def __call__(
...
+                ):
+                    filter_dict[k] = [filter_dict[k]]
+
+            self.da = self.da.sel(filter_dict)
+
+        # Validate remaining dimensions
+        if not required_dims.issubset(set(self.da.dims)):
+            raise ValueError(
+                f"Missing required dims. Data array loaded has dims {self.da.dims} but "
+                + "needs {required_dims}. Use filter_dict to select a subset of the data."
```
loliverhennigh: This string also needs to be an f-string

earth2studio/data/xr.py
```diff
@@ -62,8 +62,117 @@ def __call__(
...
+    def __init__(self, file_path: str, filter_dict: dict = None, **xr_args: Any):
+        self.file_path = file_path
+        self.da = xr.open_dataset(self.file_path, **xr_args)
+        self.da = self.da.to_array("variable")
+
+        # The following dimensions and their order is required for the data to be used as a datasource
+        required_dims = {"time", "lead_time", "variable"}
+        if filter_dict:
+            # Need to keep these dims, so make then a list if scalar value
+            for k in ("time", "lead_time"):
```
loliverhennigh: Don't you want to check for variable filters here too maybe? Like ```for k in ("time", "lead_time", "variable"):```.
Also seems like the if statement bellow could be a bit less open ended but its fine.

loliverhennigh (approved):
LGTM

earth2studio/data/xr.py
NickGeneva: Great catch, fixed!

earth2studio/data/xr.py
NickGeneva: Nah, its just the time ones we need to make a list form to the dims dont get removed, probably a slicker way to do this but oh well.

Did clean up one if statement though!

PR #482, opened 2025-09-08
Adding tp06, vertical velocity, and two invariant fields to CDS downloader

### LINKED ISSUE #456: 🐛[BUG]: Cannot run GraphCast with CDS datasource
### Version

0.8.1

uv

### Describe the issue

In this code:
```
from earth2studio.models.px import GraphCastOperational
from earth2studio.io import NetCDF4Backend
from earth2studio.data import CDS
import earth2studio.run as run
pmodel = GraphCastOperational.load_model(GraphCastOperational.load_default_package())
ds = CDS()
io = NetCDF4Backend("graphcast-output.nc")
run.deterministic(["INPUT-DATE"], 120, pmodel, ds, io)
```

I got the following error message:
`2025-08-18 17:19:38.395 | ERROR    | earth2studio.data.cds:_build_requests:254 - variable id tp06 not found in CDS lexicon`

It looks like the tp06 is accumulated total precipitation in 6 hours, while the ERA5 provides hourly value, so one has  to add 6 hours up to get a 6-hours total. Since CDS did not provide this directly, is it possible to add an extra procedure in CDS datasource code to get tp06?

Mike

Joshua-Elms:
# Earth2Studio Pull Request

## Description

closes #456

Three major additions to CDS API: 
1. 6-hourly total precipitation (tp06): calculated by downloading previous 6 hours of 1-hourly total precipitation and summing, then caches the resulting tp06 dataset.
2. Vertical velocity (w) fields added to CDS lexicon on standard thirteen levels, allowing for their download. 
3. Two invariant fields, geopotential_at_surface (z) and land_sea_mask (lsm), which are required to run GraphCastOperational. Because these fields do not exist in the CDS ERA5 0.25 degree pressure levels or single levels datasets, and the ERA5-Land dataset is 0.1 degree, these two fields are marked in the lexicon as belonging to the "google_cloud_dataset", which sets them to be downloaded via gcsfs from the example GraphCastOperational dataset provided by Google. An alternative would be interpolating them from the ERA5-Land fields, but this is the simplest way to ensure that these fields are similar to what GraphCastOperational was trained on. 

In addition, the CDS test cases have been updated to include tests of all three contributions and pass as before. 

## Checklist

- [X] An [issue](https://github.com/NVIDIA/earth2studio/issues) is linked to this pull request.

### DIFF
--- a/earth2studio/data/cds.py
+++ b/earth2studio/data/cds.py
@@ -24,6 +24,7 @@ from datetime import datetime, timedelta
 from time import sleep
 from typing import Any
 
+import gcsfs
 import numpy as np
 import xarray as xr
 from cfgrib.xarray_to_grib import to_grib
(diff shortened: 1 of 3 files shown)


NickGeneva:
Thanks for opening this PR @Joshua-Elms !

Greatly appreciate it.

The tp06 addition seems reasonable to me since its just a summation, there's no real nuance with its calculation. I hopefully test this out within the next week. Thank you for adding this!

The addition of the invariant, although adds some convenience for the GC use case, has the major issue of intermixing data sets between data sources. We do not want to have this type of functionality in the CDS data source because usage of the CDS data source implies only accessing CDS data (not Googles). This would be misleading for users.

Maybe a better approach would be to actually just cook the invariants right into the model. Namely make them part of the models package. So ideally the model always knows to go get the Google invariant fields and use them since they are unchanged, and the user would never need to be concerned. Would that work?

This will be a little more involved and for a separate PR.

Joshua-Elms:
That makes sense @NickGeneva. Mixing data sources certainly seems like a bad idea, but I hadn't thought of retrieving fields from within the model package. Since GraphCast is the only model I've seen that uses those two fields, it seems like it would be convenient for now. Could you lay out a rough model for how that would work? 

I'll remove the invariant commit from this PR, and while GraphCastOperational still won't work with the CDS, the tp06 fix should at least get FuXi working for you to test. 

Thanks for the discussion.

Joshua-Elms:
After much git difficulty I removed the invariants commit and associated test, so this should be ready for you test now @NickGeneva.

NickGeneva:
For docs, here is the manual sanity check used to verify tp06:

```python
import numpy as np
from earth2studio.data import CDS

time = datetime(2025,1,1)
ds = CDS(cache=False)
data_tp06 = ds(time, "tp06")

# Get tp at 6 hourly times (T-5h, T-4h, T-3h, T-2h, T-1h, T)
times_tp = [time - timedelta(hours=i) for i in [5, 4, 3, 2, 1, 0]]

# Fetch each hourly tp value
tp_sum = None
for t in times_tp:
    data_tp = ds(t, "tp")
    if tp_sum is None:
        tp_sum = data_tp.values.copy()
    else:
        tp_sum += data_tp.values

# Compare tp06 with sum of individual tp values
# Use allclose to account for floating point precision
np.testing.assert_allclose(
    data_tp06.values, 
    tp_sum, 
    rtol=1e-5, 
    atol=1e-8,
    err_msg="tp06 should equal the sum of 6 hourly tp values"
)
```

NickGeneva (approved):

NickGeneva:
xref: https://github.com/ecmwf/cfgrib/issues/433

Pointed to cfgrib fork with patch

Thanks again @Joshua-Elms apologies this took a while to get merged!

PR #483, opened 2025-09-10
Add AIFS diagnostic variables

### LINKED ISSUE #481: 🚀[FEA]: Diagnostic variables for AIFS

New Feature

Hello folks!

I was looking into AIFS, and noticed that currently we only have access to the prognostic variables produced by the model. 

If I understood the code correctly, from the output of `model.predict_step(x)` (which includes the diagnostic variables) we extract only the prognostic variables, and join them with the forcings from the input, and that is the output we make available (plus the recomputed variable forcings).

I would like to use the `tp06` diagnostic variables. Do you see a way to make this variable available? 
Would a separate diagnostic model that extracts the variables be a possible way? 
Or adding NaN data to the input tensor to pad the missing diagnostic variables, and removing that before the inference?

I am guessing the difficulty here is that this variable should be an output - but not an input to the model, which makes the autoregressive call difficult (since input variables != output variables).

Any pointers would be great! I would be happy to make a PR for this.

rodrigoalmeida94:
# Earth2Studio Pull Request

## Description
Adds support for diagnostic variables in AIFS. Closes #481 

## Checklist

### DIFF
--- a/earth2studio/statistics/rank.py
+++ b/earth2studio/statistics/rank.py
@@ -46,20 +46,27 @@ class rank_histogram:
     Parameters
     ----------
     ensemble_dimension: str
-        A name corresponding to a dimension to perform the
-        ranking over. Example: 'ensemble'
+        A name corresponding to a dimension to perform the ranking over.
+        Example: 'ensemble'
     reduction_dimensions: list[str]
-        A list of dimensions over which to bin the ranks.
-    number_of_bins: int
-        The number of bins to discretize the unit interval over.
-        by default, 10
+        A list of dimensions over which to bin the ranks
+    number_of_bins: int | None, optional
+        The number of bins to discretize the unit interval over. Best set to
+        ensemble_size + 1, which is set automatically when None, by default None
+    randomize_ties: bool, optional
+        When True randomize the rank in cases where multiple ensemble
+        members are exactly equal to the observation. This produces an unbiased
+        distribution of ranks in cases where ties occur frequently. If False, the rank
+        will be computed as if the observation were larger than the tied ensemble
+        members, by default True
     """
 
     def __init__(
         self,
         ensemble_dimension: str,
         reduction_dimensions: list[str],
-        number_of_bins: int = 10,
+        number_of_bins: int | None = None,
+        randomize_ties: bool = True,
     ):
         if not isinstance(ensemble_dimension, str):
             raise ValueError(
@@ -69,6 +76,7 @@ class rank_histogram:
         self.ensemble_dimension = ensemble_dimension
         self._reduction_dimensions = reduction_dimensions
         self.number_of_bins = number_of_bins
+        self.randomize_ties = randomize_ties
 
     def __str__(self) -> str:
         return "rank_histogram"
@@ -76,6 +84,12 @@ class rank_histogram:
     @property
     def reduction_dimensions(self) -> list[str]:
         return [self.ensemble_dimension] + self._reduction_dimensions
+
+    def _get_number_of_bins(self, input_coords: CoordSystem) -> int:
+        if self.number_of_bins is None:
+            return len(input_coords[self.ensemble_dimension]) + 1
+        else:
+            return self.number_of_bins
 
     def output_coords(self, input_coords: CoordSystem) -> CoordSystem:
         """Output coordinate system of the computed statistic, corresponding to the given input coordinates
@@ -91,6 +105,7 @@ class rank_histogram:
             Coordinate system dictionary
         """
         output_coords = input_coords.copy()
+        number_of_bins = self._get_number_of_bins(input_coords)
         for dimension in self.reduction_dimensions:
             handshake_dim(input_coords, dimension)
             output_coords.pop(dimension)
@@ -99,7 +114,7 @@ class rank_histogram:
             OrderedDict(
                 {
                     "histogram_data": np.array(["bin_centers", "bin_counts"]),
-                    "bin": np.arange(self.number_of_bins),
+                    "bin": np.arange(number_of_bins),
                 }
             )
             | output_coords
@@ -161,6 +176,8 @@ class rank_histogram:
                 handshake_coords(y_coords, x_coords, c)
                 coord_count += 1
 
+        number_of_bins = self._get_number_of_bins(x_coords)
+
         # Get the dimension index of the ensemble dimension
         dim = list(x_coords).index(self.ensemble_dimension)
 
@@ -168,7 +185,16 @@ class rank_histogram:
         x = torch.movedim(x, dim, 0)
 
         # Compute the ranks over the ensemble dimension
-        _ranks = torch.mean(1.0 * torch.ge(y, x), dim=0)
+        _ranks = torch.sum(torch.ge(y, x).to(dtype=torch.int32), dim=0)
+
+        if self.randomize_ties:
+            _ranks_gt = torch.sum(torch.gt(y, x).to(dtype=torch.int32), dim=0)
+            rank_diff = _ranks - _ranks_gt
+            ties = rank_diff > 0
+            _ranks[ties] -= (
+                torch.rand_like(_ranks[ties], dtype=torch.float32)
+                * (rank_diff[ties] + 1)
+            ).to(dtype=torch.int32)
 
         # Reshape ranks using reduction dimensions
         dims = [list(y_coords).index(rd) for rd in self._reduction_dimensions]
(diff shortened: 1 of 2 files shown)


test/models/px/test_aifs.py
```diff
@@ -23,6 +23,7 @@
 
 try:
     import anemoi  # noqa: F401
+    from dotdict import DotDict
```
NickGeneva: Is this dependency only needed for testing? What does it do here?

Ideally would like to avoid adding test case deps unless they have use across multiple isntances.

Also this seems to be a [GPL](https://github.com/nicholasmireles/DotDict?tab=GPL-3.0-1-ov-file#readme) license which unfortunately means we will need to avoid.

Is there a way to not use this?

test/models/px/test_aifs.py
```diff
@@ -45,8 +46,52 @@ def make_two_nnz_per_first_row_csr(n_rows, n_cols, device):
 
 
 class PhooAIFSModel(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+        # Mock data indices, annoying
```
NickGeneva, line 68: @loliverhennigh to help with the testing as needed.

earth2studio/models/px/aifs.py
```diff
@@ -176,207 +192,63 @@ def __init__(
...
-            101,
-            102,
-        ]
+        self.input_variables = self.get_input_variables()
+        self.output_variables = self.get_output_variables()
 
     def __str__(self) -> str:
         return "aifs-single-1.0"
 
+    def get_input_variables(self) -> list[str]:
```
NickGeneva: Could just unify this and the property with:

```
@property
def input_variables(self) -> list[str]:
```

Then can eliminate the props above, is there a reason to keep these as a function that I'm missing?

NickGeneva: Same deal with `get_output_variables`

earth2studio/models/px/aifs.py
```diff
@@ -716,17 +586,15 @@ def _prepare_input(
...
         coords: CoordSystem,
     ) -> torch.Tensor:
-        """Copy output tensor to input tensor following the AIFS model."""
+        """Update time based inputs."""
 
-        # Copy prognostic field output to the input tensor
-        x = x.roll(-1, dims=1)
-        x[:, -1, :, self.prognostic_input_mask] = y[..., self.prognostic_output_mask]
+        # Select only inputs
+        x = x[..., self.model.data_indices.data.input.full]
```
NickGeneva, line 598: Is this something that has an API doc for or did you have to dig around anemoi  to find it?
(just wanted to see if theres a link that could be added as a comment to serve as reference.

earth2studio/models/px/aifs.py
```diff
@@ -133,11 +132,32 @@
...
+    "sin_latitude",
+    "sin_longitude",
+    "cos_julian_day",
+    "cos_local_time",
+    "sin_julian_day",
+    "sin_local_time",
+    "insolation",
+    "u100m",
+    "v100m",
+    "hcc",
```
NickGeneva, line 148: Some of these new fields will need to get logged in https://github.com/NVIDIA/earth2studio/blob/main/earth2studio/lexicon/base.py

So we know what they mean and the units

earth2studio/models/px/aifs.py
```diff
@@ -41,8 +41,7 @@
 except ImportError:
     OptionalDependencyFailure("aifs")
 
-
-VARIABLES = [
+ALL_VARIABLES = [
```
NickGeneva: I would keep this as `VARIABLES` just because these variable lists are commonly called that across models, so its just a little more consistent. No problem with having it be the output variables.

NickGeneva:
Thank you for this PR @rodrigoalmeida94 and taking a dive into anemoi to figure out how to get these added.

Added a few comments from an initial pass. Please ping @loliverhennigh or myself for reviews (we will be off for the rest of this week).

Greatly appreciate this!

NickGeneva:
@loliverhennigh please actively help with this PR as needed since you have the expertise from the original integration :+1:

test/models/px/test_aifs.py
rodrigoalmeida94: Uhr yeah, I think it should be included in the anemoi but it is not. I mocked it for the test.

test/models/px/test_aifs.py
rodrigoalmeida94: Getting failures on these tests, not really sure why:
```
FAILED test/models/px/test_aifs.py::test_aifs_exceptions[cpu-dc1] - RuntimeError: mat1 and mat2 shapes cannot be multiplied (542080x1038240 and 1036800x188)
FAILED test/models/px/test_aifs.py::test_aifs_exceptions[cpu-dc2] - RuntimeError: mat1 and mat2 shapes cannot be multiplied (542080x1038240 and 720x188)
FAILED test/models/px/test_aifs.py::test_aifs_exceptions[cuda:0-dc1] - RuntimeError: mat1 and mat2 shapes cannot be multiplied (542080x1038240 and 1036800x188)
FAILED test/models/px/test_aifs.py::test_aifs_exceptions[cuda:0-dc2] - RuntimeError: mat1 and mat2 shapes cannot be multiplied (542080x1038240 and 720x188)
```
Maybe the shapes need to be changed?

earth2studio/models/px/aifs.py
rodrigoalmeida94: Yes absolutely, changed

earth2studio/models/px/aifs.py
rodrigoalmeida94: Yup its in the docs in [AnemoiModelInterface ](https://anemoi.readthedocs.io/projects/models/en/latest/modules/data_indices.html#usage-information)

Added comment

earth2studio/models/px/aifs.py
rodrigoalmeida94: Ah so VARIABLES is input + output + forcings, so it includes all variables that are used by the model - would that be okay?

earth2studio/models/px/aifs.py
rodrigoalmeida94: I adde the missing variables to the base lexicon - I left out the forcings i.e. cos_latitude, since I don't think it makes sense to put them there, what do you think?

rodrigoalmeida94:
I just added a commit that in a very hacky way fixes two problems I found:
1. Error when requesting to write out specific variables that are only diagnostic:
For instance when doing:
```
out = ensemble([t], 3, 2, aifs, data, io, Zero(), batch_size=1, output_coords={
                        "variable": np.array(["t2m", "tp06"]),
                        "lat": np.linspace(-90, 90, 181, endpoint=True),
                        "lon": np.linspace(0, 360, 360, endpoint=False),
                    })
```
I would get an error because `tp06` is not the in the input (i.e. tilmestep 0). My solution has been to pad the input with 0s for these variables and add them in the coordinates. Not ideal, but I feel it makes sense?

2. Error when using the ensemble method
Not really sure why but it seems like the `lead_time` variable gets carried between ensemble members (maybe some setting is happening in place?) which causes this exception:
```
ValueError: Attempting to set a selection of 1 elements with an array of 65160 elements.
```
I just copied the "lead_time" update that was in FuXi and that works. 

Hope this makes sense! Let me know what you think.

I'm working in another [branch](https://github.com/rodrigoalmeida94/earth2studio/tree/aifs-crps) on implementing AIFS ENS but it's been a bit of a pain to get the dependencies to work - since it requires different versions of anemoi than AIFS Single.

earth2studio/models/px/aifs.py
```diff
@@ -133,11 +132,32 @@
...
+    "sf",
+    "ssrd06",
     "stl1",
     "stl2",
+    "strd06",
     "swvl1",
     "swvl2",
-]
+    "tcc",
+]  # from config.json >> dataset.variables
```
rodrigoalmeida94, line 160: Note that these are different between models (AIFS Single v. ENS) and between checkpoints (I checked v1.1 and the order of the variables changes)

loliverhennigh: I am looking at this now. This is a bit rough. We can probably make a separate PR for 1.1 that allows for the variable changes. For ENS it looks like we would just need a new model file for it.

test/models/px/test_aifs.py
loliverhennigh: Just found the issue and pushed a fix

loliverhennigh:
The input padding solution looks good to me. I can't really think of another way around it. I fixed the mat test issue and will try to get this through CI now. If I can I think we are good to merge unless you want to update anything else./

Amazing work!

PR #485, opened 2025-09-15
TempestExtremes tracking

mariusaurus:
# Earth2Studio Pull Request

## Description
This PR adds a connector to TempestExtremes in the form of a diagnostic model.
Since TE natively only supports command-line interface, the model is implemented by spawning a sub-process call. The command for that call has to be provided by the user, providing flexibility in usage and installation of TE.

Due to the nature of track data (multiple orders of magnitude smaller than field data and unstructured), the diagnostic here does not return the tracks as tensor, but finishes with TE writing its file to disk. That leaves the question if this and potentially other trackers fit in this category. Maybe this PR can be basis for discussion.

## Checklist

### DIFF
--- a/earth2studio/models/dx/tempest_extremes.py
+++ b/earth2studio/models/dx/tempest_extremes.py
@@ -18,95 +18,13 @@ from rich.table import Table
 from rich.traceback import Traceback
 
 from earth2studio.io import KVBackend
-from earth2studio.utils.coords import CoordSystem, map_coords, split_coords
-
-
-def tile_xx_to_yy(
-    xx: torch.Tensor, xx_coords: CoordSystem, yy: torch.Tensor, yy_coords: CoordSystem
-) -> tuple[torch.Tensor, CoordSystem]:
-    """Tile tensor xx to match the leading dimensions of tensor yy.
-
-    Parameters
-    ----------
-    xx : torch.Tensor
-        Source tensor to be tiled
-    xx_coords : CoordSystem
-        Coordinate system for xx tensor
-    yy : torch.Tensor
-        Target tensor whose shape determines tiling
-    yy_coords : CoordSystem
-        Coordinate system for yy tensor
-
-    Returns
-    -------
-    Tuple[torch.Tensor, CoordSystem]
-        Tuple containing the tiled tensor and updated coordinate system
-
-    Raises
-    ------
-    ValueError
-        If xx has more dimensions than yy
-    """
-    n_lead = len(yy.shape) - len(xx.shape)
-
-    if n_lead < 0:
-        raise ValueError("xx must have fewer dimensions than yy.")
-
-    out_shape = yy.shape[:n_lead] + tuple([-1 for _ in range(len(xx.shape))])
-    out_coords = copy.deepcopy(yy_coords)
-    for key, val in xx_coords.items():
-        out_coords[key] = val
-
-    return xx.expand(out_shape), out_coords
-
-
-def cat_coords(
-    xx: torch.Tensor,
-    cox: CoordSystem,
-    yy: torch.Tensor,
-    coy: CoordSystem,
-    dim: str = "variable",
-) -> tuple[torch.Tensor, CoordSystem]:
-    """
-    concatenate data along coordinate dimension.
-
-    Parameters
-    ----------
-    xx : torch.Tensor
-        First input tensor which to concatenate
-    cox : CoordSystem
-        Ordered dict representing coordinate system that describes xx
-    yy : torch.Tensor
-        Second input tensor which to concatenate
-    coy : CoordSystem
-        Ordered dict representing coordinate system that describes yy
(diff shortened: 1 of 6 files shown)


mariusaurus:
hi @NickGeneva @dallasfoster ,
from my side the PR is ready to review.
testing and ensemble size>1 is now supported as well.

earth2studio/models/dx/tempest_extremes.py
```diff
@@ -0,0 +1,1221 @@
...
+    _cox["lat"] = coy["lat"]
+    xx, cox = map_coords(xx, cox, _cox)
+
+    coords = cox.copy()
+    dim_index = list(coords).index(dim)
+
+    zz = torch.cat((xx, yy), dim=dim_index)
+    coords[dim] = np.append(cox[dim], coy[dim])
+
+    return zz, coords
```
dallasfoster: can we move this to a utils.coords file and add unit tests for it?

earth2studio/models/dx/tempest_extremes.py
```diff
@@ -0,0 +1,1221 @@
...
+
+from earth2studio.io import KVBackend
+from earth2studio.utils.coords import CoordSystem, map_coords, split_coords
+from earth2studio.utils.imports import (
+    OptionalDependencyFailure,
+    check_optional_dependencies,
+)
+
+
+def tile_xx_to_yy(
```
dallasfoster: Can this be added to a utils file and have unit tests for it.

earth2studio/models/dx/tempest_extremes.py
```diff
@@ -0,0 +1,1221 @@
...
+                command=self.stitch_cmd + ["--in", node_file, "--out", track_file],
+                print_output=self.print_te_output,
+            )
+
+        # remove helper files
+        self.tidy_up(insies, outsies)
+
+        print(f"took {(time.time() - then):.1f}s to track cyclones")
+
+        return
```
dallasfoster, line 580: Until we make a decision on !494, this PR will require returning an nd array before merging.

dallasfoster:
Are there any instructions for installation or checks if the software is not available?

mariusaurus:
> Are there any instructions for installation or checks if the software is not available?

yes, just added install instructions to the docs and checks if software is available is done at class initialisation. Since TE is not available as python package, users have to install it (locally) on thier system. AT class ini, they have the option to provide the path to the executable and until that path is known it is not possible to test install.

The default dependency failure does not work here for the same reason, and also cannot be hacked to write a message which is reasonable for the present case. So I provided a custom one. Hope that's fine

earth2studio/models/dx/tempest_extremes.py
mariusaurus: moved function and associated unit test to utils

earth2studio/models/dx/tempest_extremes.py
mariusaurus: moved function and associated unit test to utils

mariusaurus:
would a small example of using TE help? if so, where should I put it? could do at the end of the tempest_extremes.py file, just not sure if it justifies a full example.

mariusaurus:
Hi!
have you found a solution where to best put this? thanks!

earth2studio/utils/coords.py
```diff
@@ -429,3 +430,93 @@ def convert_multidim_to_singledim(
...
+    """
+    n_lead = len(yy.shape) - len(xx.shape)
+
+    if n_lead < 0:
+        raise ValueError("xx must have fewer dimensions than yy.")
+
+    out_shape = yy.shape[:n_lead] + tuple([-1 for _ in range(len(xx.shape))])
+    out_coords = deepcopy(yy_coords)
+    for key, val in xx_coords.items():
+        out_coords[key] = val
```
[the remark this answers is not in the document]
mariusaurus: added check that trailing dims of yy have same name as all dims of xx. Also added a test verifying that otherwise the method fails

test/utils/test_coords.py
```diff
@@ -420,3 +425,67 @@ def test_convert_multidim_to_singledim_additional():
...
+            "lon": np.linspace(0, 360, 1440),
+        }
+    )
+
+    result, result_coords = tile_xx_to_yy(xx, xx_coords, yy, yy_coords)
+
+    # Result should have yy's leading dims + all of xx's dims
+    # yy.shape = (3, 4, 5, 721, 1440), xx.shape = (2, 721, 1440)
+    # n_lead = 5 - 3 = 2, so we prepend yy's first 2 dims to xx
+    # Result shape should be (3, 4, 2, 721, 1440)
```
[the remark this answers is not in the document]
mariusaurus: no, should be five minus three as xx has three dimensions and yy has five

mariusaurus:
...will be added through tc_tracking recipe in new PRs.

PR #486, opened 2025-09-17
Balance ties in rank histogram

### LINKED ISSUE #452: 🐛[BUG]: Biased rank histogram with tied values
### Version

Current main

source

### Describe the issue

In the rank histogram code, ranks are computed as
https://github.com/NVIDIA/earth2studio/blob/7b6510a74434295e338c3b795900cb9deaec349e/earth2studio/statistics/rank.py#L171

This will return biased ranks if ensemble members are exactly equal to the real observations. This can happen in models that often return a particular exact value. For instance a precipitation model might often give an output of 0 (no precipitation). In that case, if there are 32 ensemble members that are all 0 and the observed precipitation is also 0, the code above returns a normalized rank of 1, which makes it incorrectly seem that the prediction is highly biased.

The normal procedure to handle such cases is to randomize ties. The easiest way to achieve this is to add a small amount of random noise to both the observation and the predictions before the rank calculation.

jleinonen:
# Earth2Studio Pull Request

## Description

This fixes the imbalance issue in rank histogram in case of ties (reported in https://github.com/NVIDIA/earth2studio/issues/452) and adds a check for the appropriate number of bins. Main changes:

1. Add option `randomize_ties` with default of `True`. If enabled, it will detect ties and randomize the rank between the tied ensemble members. For example, if the observation is `0` and the ensemble of predictions is `[0, 0, 0, 1, 2]`, the rank can be any value among `(0, 1, 2, 3)`, with equal probability. This comes with a small performance penalty for detecting ties; if ties are not an issue and/or exactly deterministic results are needed, one can set `randomize_ties=False`.
2. Set `number_of_bins=None` by default, in which case the number of bins is set to `ensemble_size + 1`. There are `ensemble_size + 1` possible discrete values of the rank and therefore the rank histogram usually only really makes sense if the number of bins is equal to this number. The user can still set it manually to an int if they really want a different number of bins.
3. Update tests to test for tie randomization and enforcing the appropriate number of bins.
4. Add docstrings for the new options.

## Checklist


jleinonen:
~Hold on this a bit - I think we can just default `number_of_bins=None` and set it based on the ensemble size in `output_coords`. I'll make a fix tomorrow.~

Done - the code is now refactored to use `number_of_bins=None` by default.

dallasfoster (approved):
Looks good, thank you for the submission.

PR #492, opened 2025-09-19
Fix FCN3 Random Noise Handling

dallasfoster:
# Earth2Studio Pull Request

## Description
Updates a few aspects of the FCN3 handling of random noise.

First, adds an ability to set the internal random seed of the FCN3 noise process.
Second, adds handling of the internal state. The noise process in FCN3 is stateful, which means we must track the noise for each ensemble member and each time as the model rolls out. 

## Checklist


earth2studio/models/px/fcn3.py
```diff
@@ -278,6 +296,54 @@ def load_model(
...
+        """Get the internal RNG state for the given ensemble and time index
+
+        Parameters
+        ----------
+        ensemble_index : int
+            Ensemble index
+        time_index : int
+            Time index
+        """
+        return self._internal_noise_states[ensemble_index][time_index]
```
NickGeneva, line 309: Does `self._internal_noise_states` need to be initialized in the constructor?

Like `self._internal_noise_states = []`

Its created when `_reset_internal_state` is called, is there a chance for this function to try to access this without it being initialized?

Maybe it starts as none and this can throw an initialization error that _reset_internal_state has not been called?

earth2studio/models/px/fcn3.py
```diff
@@ -155,16 +157,32 @@ def __init__(
         self,
         core_model: torch.nn.Module,
         variables: np.array = np.array(VARIABLES),
+        seed: int = 333,
     ):
         super().__init__()
         self.model = core_model
         self.variables = variables
+        self.seed = seed
```
NickGeneva: Maybe this should get moved down into `set_rng` so if a user updates `set_rng` the property is updated too.

NickGeneva (approved):
LGTM, only two minor questions but logic makes sense

Approving to unblock :+1:

earth2studio/models/px/fcn3.py
dallasfoster: I started with this approach but was not satisfied because of the mypy errors. I could through some errors but users should not be using this hidden functions and they are being properly initialized now.

earth2studio/models/px/fcn3.py
dallasfoster: I can do that

PR #500, opened 2025-09-25
Add AIFS ENS

rodrigoalmeida94:
# Earth2Studio Pull Request

## Description
Adds the AIFS ENS model (as a separate module because it needs different versions of anemoi to run).

## Checklist

### DIFF
--- a/earth2studio/models/px/aifsens.py
+++ b/earth2studio/models/px/aifsens.py
@@ -717,6 +717,7 @@ class AIFSENS(torch.nn.Module, AutoModelMixin, PrognosticMixin):
         tuple[torch.Tensor, CoordSystem]
             Output tensor and coordinate system 6 hours in the future
         """
+        _ = self.output_coords(coords)  # Ensure coordinate validation like AIFS
         x = self._prepare_input(x, coords)
         x, coords = self._forward(x, coords)
         x = self._prepare_output(x, coords)


loliverhennigh:
Ill be taking a look at this @rodrigoalmeida94 when its ready. Thanks for the PR! Really excited to see this getting added.

rodrigoalmeida94:
@loliverhennigh getting back on this - and unfortunately I seem so have found a bug in this implementation (cc @mariusaurus ). 

It seems like t2m is giving out really strange values after step 1, these plots are the t2m bias of the forecast to ERA5:
image
image
image
(All other variables I tested - u and v, t850, z500, tp06 - actually look great which is even weirder)

It's really strange because there seems to be some sort of hemispheric mismatch (east and west) that causes this very high bias in the forecasts. I was thinking maybe that this would have something to do with the forcings, and the way I am overwriting them, but so far haven't been able to track down the root cause.

Have you encountered this in AIFS as well? Would you have any pointers? Any help would be amazing!

hansbrenna:
Sorry to butt in to this thread, but I'm pretty interested in this addition to Earth2Studio. 

Just by looking at the plots you share my initial hypothesis is that it's related to the sine/cosine of longitude and/or the sine/cosine of local time. Those are the input parameters that could lead to this east-west bias pattern. If you're willing to share your recipe for running the model in Earth2Studio, I would be happy to help investigate.

rodrigoalmeida94:
@hansbrenna that would be awesome! I just pushed a notebook I have been using to play around. I think your hypothesis is supported by what I have found so far - somehow the order of the forcings seems to be incorrect. https://github.com/NVIDIA/earth2studio/pull/500/commits/eedd6daf8e219c7d7e6e8e2fcbac9e6f9069cb96

You should be able to run this if you replace the custom dataset I use in there with the regular ARCO dataset.

hansbrenna:
Thank you for sharing :). I will try to set it up to run on our infrastructure when I have time, probably tomorrow. 

Just looking at your plots of the forcings in the notebook, there seems to be an unexplained change in the sine of longitude between lead 0 and 1. This should be a constant forcing like you show in the final cell. 
image

I'm also unsure if the labels correspond correctly to the plotted fields. If that is what the model gets it could explain the errors:

- cos(longitude) should be an even function
- sin(longitude) should be an odd function

rodrigoalmeida94:
Okay I think I got it: https://github.com/NVIDIA/earth2studio/pull/500/commits/cb2af8aca46322513c21f8bca038630948a63db8

Apparently I wasn't updating the forcings with the correct date, reason why we got this hemispheric pattern. Now the forecasts at least appear not to have any systematic bias:
image (3)
image (4)
image (5)

I'm computing some large scale metrics to double check, and will report back!

hansbrenna:
That's great! I was able to set up AIFSENS on our cluster using the recipe you provided, but when I try to run it using the ARCO data source it's not able to find the "sdor" variable even though I can find it in the ARCO bucket. Is this something you had to work around?

By the way I would prefer to run the model with input data from ecmwf-opendata of with data from CDS. Any input would be appreciated.

mariusaurus:
hi @hansbrenna,

not the best level of support, I know, but as a quick workaround have a look at the lexicon at this branch https://github.com/mariusaurus/earth2studio/tree/mkoch/aifs-crps/earth2studio/lexicon and copy the base, ARCO and CDS files to yours. That should find those variables.
It's a hacky solution I did for a small test, that's why I haven't upstreamed it to @rodrigoalmeida94's branch, but can probably do so if I find some time tomorrow. iirc some of the variables were invariants, so there is additional potential for streamlining the whole thing.

loliverhennigh:
This looks really good @rodrigoalmeida94. Is this about ready for review? I can help tomorrow get a few more things added to it, tests, adding to the docs, etc... We are getting asked by others for this model in E2S so going to make this high priority for me.

rodrigoalmeida94:
@loliverhennigh yes for sure! 
I don't think I will have time to make additional improvements in the next weeks so feel free to take it over - in terms of the metrics: I computed errors for a long time span and the issue we had before is definitely gone, so we are good to go!

One aspect that I would highlight in the docs: This model is intended to run with IFS perturbed initial conditions, so the performance will be ' worse' when using a single unperturbed ERA5 IC, but I have found the performance to be quite good (in line with ENS) even with ERA5 ICs.

loliverhennigh:
@rodrigoalmeida94, Fantastic! Thanks for the awesome work. Ill take over and try to get this in ASAP.

loliverhennigh:
This is turning into a CI nightmare for me. We should be good to merge when I manage to get it through. Its kinda complicated but the jupyter broke the ability to run CI (even though I removed it). Then I monkeyed about and broke some stuff. Will try to get this finally in by EOD.

NickGeneva:
Thanks for the PR again @rodrigoalmeida94 we got it merged :pray:

rodrigoalmeida94:
> Thanks for the PR again @rodrigoalmeida94 we got it merged 🙏

Wonderful! Thanks for pushing this through

PR #501, opened 2025-09-26
Allow skipping the DiagnosticWrapper bounds check

swbg:
# Earth2Studio Pull Request

## Description
Users may want to skip the latitude/longitude coordinate bounds check. For example, when running `SolarRadiationAFNO1h` on top of `InterpModAFNO`. The change will allow to pass a flag to skip the bounds check.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - InferenceOutputSource can now accept an Xarray Dataset directly as an argument
 - InferenceOutputSource returns data consistently in `("time", "lead_time", "variable")`
   order
+- Added support for ERA5 model levels and additional variables in ARCO data source
+- Changed the HRRR_X and HRRR_Y coordinates of the HRRR data source to match the native
+  LCC coordinates
 
 ### Deprecated
 
--- a/docs/userguide/advanced/lexicon.md
+++ b/docs/userguide/advanced/lexicon.md
@@ -5,7 +5,7 @@
 As discussed in detail in the {ref}`data_userguide` section, Earth2Studio tracks the
 geo-physical representation of tensor data inside workflows.
 This includes the name of the variable / parameter / property the data represents, which
-is tacked explicitly via Earth2Studios lexicon.
+is tracked explicitly via Earth2Studios lexicon.
 Similar to ECMWF's [parameter database](https://codes.ecmwf.int/grib/param-db/),
 Earth2Studio's lexicon aims to provide an opinioned and explicit list of short variables
 names that is used across the package found in {py:obj}`earth2studio.lexicon.base.E2STUDIO_VOCAB`.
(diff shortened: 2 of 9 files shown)


NickGeneva:
Why are the bounds of SolarRadiationAFNO1h and InterpModAFNO different? Because of the latitude is not pole including?

Maybe the check should just log a warning, that theres not a full overlap and values will be extrapolated rather than adding a new parameter.

swbg:
I am wondering whether we should skip the check altogether. If we want to warn about extrapolation in general, the warning should happen in map_coords, not specifically here. Furthermore, then why not also warn about interpolation? @jleinonen any thoughts?

jleinonen:
> I am wondering whether we should skip the check altogether. If we want to warn about extrapolation in general, the warning should happen in map_coords, not specifically here. Furthermore, then why not also warn about interpolation? @jleinonen any thoughts?

We decided to add this check with @dallasfoster because in general it's a bad idea to extrapolate (imagine if we tried to run e.g. a global precip diagnostic from StormCast outputs). It's really only a good idea in this edge case where you just need to extrapolate one row pixels on the edge to match the dimensions. That's why I thought this should preferably be an option so it can be used as an "I know what I'm doing" switch.

jleinonen:
> Why are the bounds of SolarRadiationAFNO1h and InterpModAFNO different? Because of the latitude is not pole including?
> 
> Maybe the check should just log a warning, that theres not a full overlap and values will be extrapolated rather than adding a new parameter.

Yes exactly, `InterpModAFNO` uses the AFNO encoder/decoder where the inputs have to be a multiple of the patch size, so the input excludes the south pole to crop to 720 rows. While the solar radiation AFNOs have a redesigned encoder that can handle the full 721 rows.

I'm also happy if we change it to a warning, though.

dallasfoster:
I think this is a bad idea and it should be solved in some other way: let the user extrapolate their data in their preferred way.

swbg:
@dallasfoster what would you suggest to do? This PR is only about whether we want to raise an error when extrapolating specifically with the DiagnosticWrapper. The extrapolation usually happens automatically in map_coords. I would suggest to remove the checks from DiagnosticWrapper and, optionally, add a warning to map_coords.

jleinonen:
I tend to agree with @swbg: this is just allowing `DiagnosticWrapper` to do what `map_coords` will happily do already. And here it's really only about enabling the very safe edge case of extrapolating a single row to fill in a (721, 1440) input from a (720, 1440) output. I think it would be a pity if we exclude these use cases from `DiagnosticWrapper` because writing the workflows that handle this manually will need considerable extra code.

NickGeneva:
Hmmm, fair point about the `map_coords` behavior... but I've always viewed map_coords as a necessary evil rather than a model for other parts of the package.

The main reason I ask about not having the boolean and then automatic, but then have it print a warning to console is that with so many models having the 720 lat issue right now its just broken out of the box for a lot which really limits the user experience as @jleinonen mentioned.  This class is more about being a convenience item than anything... adding a boolean toggle doesn't do a whole lot for most users that wont know about this option and clutters API.

I personally feel that there should at least be a warning to show that a user is potentially doing something wrong.  Granted I could see people complaining that its going to spam the console log, but thats better than nothing imo. The can adjust the loguru level in their script to error.

I do have the same concerns as Dallas with just magic stuff happening under the hood without the user knowing, but you can make the argument about interpolation I reckon.

swbg:
I have changed it to what I think would make sense because I am not sure whether we are all talking about the same thing. Please let me know what you think.

NickGeneva:
closed with: https://github.com/NVIDIA/earth2studio/pull/532

PR #514, opened 2025-10-07
distilled model super resolution

loliverhennigh:
# Earth2Studio Pull Request

## Description
super resolution distilled model

### DIFF
--- a/earth2studio/models/dx/cbottle_sr.py
+++ b/earth2studio/models/dx/cbottle_sr.py
@@ -43,6 +43,7 @@ try:
         StackedRandomGenerator,
         edm_sampler,
         edm_sampler_from_sigma,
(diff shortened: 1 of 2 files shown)


earth2studio/models/dx/cbottle_sr.py
```diff
@@ -402,6 +411,9 @@ def load_model(
             Number of diffusion steps, by default 18
         sigma_max : float, optional
             Noise amplitude used to generate latent variables, by default 800
+        distilled_model : bool, optional
```
NickGeneva, line 414: Lets not expose this in the API right now since its not released

I was hoping zip names can be the same package location should change.
and the distilled package should be able to switch the sampler mode just based on its contents.

PR #516, opened 2025-10-10
Changes to enable FCN3+StormCast workflows

### LINKED ISSUE #512: 🐛[BUG]: Lead times in NetCDF4Backend outputs not decoded correctly
### Version

Current main

source

### Describe the issue

I noticed that lead times in files produced by `NetCDF4Backend` were not correctly decoded to `timedelta64` in `xarray.open_dataset`. It seems the reason is that here https://github.com/NVIDIA/earth2studio/blob/3885668d192f2aac64d7646fd8e19e1675765a62/earth2studio/io/netcdf4.py#L30 we have
```python
units_map = {
    "h": "hour",
    "D": "day",
    "s": "second",
    "m": "minute",
    "Y": "year",
}
```
but in the [Xarray docs for `open_dataset`](https://docs.xarray.dev/en/stable/generated/xarray.open_dataset.html) it says

> decode variables and coordinates with time units in {“days”, “hours”, “minutes”, “seconds”, “milliseconds”, “microseconds”} into timedelta objects

It seems that the units being singular (e.g. `"hour"`) rather than plural (`"hours"`) is preventing them from being decoded properly.

I changed the units to plural in a private copy and now the decoding in Xarray works. I can PR a fix but I wanted to ask first if you see any unintended consequences from this. That is, is there something that depends on the units being singular?

### LINKED ISSUE #513: 🐛[BUG]: StormCast ignores order of conditioning dimensions
### Version

Current main

source

### Describe the issue

I was trying to use `earth2studio.data.xr.InferenceOuputSource` as the `conditioning_data_source` of the `StormCast` model (to use the results of a global forecast model as the conditioning). While the inference ran without errors, surprisingly all the results were garbage.

The root cause seems to be that while most data sources return the data with dimensions `('time', 'lead_time', 'variable', 'lat', 'lon')`, `InferenceOuputSource` returns `('variable', 'time', 'lead_time', 'lat', 'lon')`. There doesn't seem to be anything wrong about the latter order of dimensions but the `StormCast` wrapper implicitly assumes the former order, as `fetch_data` returns the data in the order found in the source, and there is no check of the ordering of the dimensions in the wrapper:
https://github.com/NVIDIA/earth2studio/blob/3885668d192f2aac64d7646fd8e19e1675765a62/earth2studio/models/px/stormcast.py#L402-L431

I was able to get this to work by transposing the order of dimensions in `InferenceOuputSource`, but I think it would be more robust to check and reorder the dimensions in the wrapper as needed. I'm not sure if other wrappers such as CorrDiff are also affected by this. Perhaps the input coordinate checking combined with `map_coords` will catch/correct such issues more robustly in those models?

One suggestion I can offer to make enforcing the order of dimensions easier is to add an option to `fetch_data` to return the dimensions in a specified order.

Tagging @pzharrington for visibility.

jleinonen:
# Earth2Studio Pull Request

## Description

Various minor changes/fixes to unblock a workflow using StormCast conditioned by FCN3:

 1. Change to `StormCast` wrapper: Use `map_coords` to ensure variables pulled from data source have dimensions in the expected order (resolves https://github.com/NVIDIA/earth2studio/issues/513)
 1. Change to `NetCDF4Backend`: Lead time timedelta units are set to plural spelling to enable correct decoding in Xarray (resolves https://github.com/NVIDIA/earth2studio/issues/512)
 1. Changes to `InferenceOutputSource`
    1. Typo fix in name: `InferenceOuputSource` renamed to `InferenceOutputSource`
    2. `InferenceOutputSource` can now also accept an Xarray Dataset directly as an argument, in addition to a path to a file containing one.
    3. Following the suggestion of @gertln, also ensure that `InferenceOutputSource` handles its dimensions in the expected `("time", "lead_time", "variable")` order rather than `("variable", "time", "lead_time")`. This change is not strictly necessary for the current PR because of the change to `StormCast` but may avoid similar problems arising in the future.
 4. Add tests for the above changes/fixes.

For reference, below is the workflow that this PR unblocks. With these changes, it works using both `fcn_result_storage="memory"` and `fcn_result_storage="file"`:

```python
from datetime import datetime
from typing import Literal

import torch
import xarray as xr

from earth2studio.data import GFS, HRRR, InferenceOutputSource
from earth2studio.io import NetCDF4Backend, XarrayBackend
from earth2studio.models.px import DiagnosticWrapper, FCN3, InterpModAFNO, StormCast
from earth2studio.models.dx import DerivedSurfacePressure
from earth2studio import run


def stormcast_fcn3_workflow(
    start_time: datetime,
    num_hours: int,
    output_file: str,
    fcn3_result_storage: Literal["memory", "file"] = "memory",
    device: str | torch.device = 'cuda'
):
    # models
    ## FourCastNet 3
    fcn3_package = FCN3.load_default_package()
    fcn3 = FCN3.load_model(fcn3_package)
    
    ## Surface pressure interpolation (needed by downstream models)
    orography_fn = fcn3_package.resolve("orography.nc")
    with xr.open_dataset(orography_fn) as ds:
        z_surface = torch.as_tensor(ds["Z"][0].values)
    z_surf_coords = OrderedDict({
        d: fcn3.input_coords()[d] for d in ["lat", "lon"]
    })
    sp_model = DerivedSurfacePressure(
        p_levels=[50, 100, 150, 200, 250, 300, 400, 500, 600, 700, 850, 925, 1000],
        surface_geopotential=z_surface,
        surface_geopotential_coords=z_surf_coords
    )

    ## Bundle surface pressure to FCN3
    fcn3_sp = DiagnosticWrapper(px_model=fcn3, dx_models=sp_model)
    
    ## Temporal interpolation to 1h resolution
    fcn3_interp = InterpModAFNO.from_pretrained()
    fcn3_interp.px_model = fcn3_sp
    fcn3_interp.to(device=device)

    ## Pretrained StormCast model
    stormcast = StormCast.from_pretrained()

    # initial conditions
    gfs_ic = GFS()
    hrrr_ic = HRRR()

    # output backends
    if fcn3_result_storage == "memory":
        fcn3_results = XarrayBackend()
    else:
        fcn3_results = NetCDF4Backend("/tmp/fcn3_output.nc") # when using temporary file
    stormcast_results = NetCDF4Backend(output_file)
    
    # run interpolated FCN3
    run.deterministic([start_time], num_hours, fcn3_interp, gfs_ic, fcn3_results, device=device)
    
    # set StormCast to use FCN3 output
    if fcn3_result_storage == "memory":
        source = InferenceOutputSource(fcn3_results.root)
    else:
        source = InferenceOutputSource("/tmp/fcn3_output.nc") # when using temporary file
    stormcast.conditioning_data_source = source

    # run StormCast
    run.deterministic([start_time], num_hours, stormcast, hrrr_ic, stormcast_results, device=device)

```

## Checklist


pzharrington (approved):
StormCast changes look good to me

earth2studio/data/xr.py
```diff
@@ -127,6 +136,15 @@ def __init__(self, file_path: str, filter_dict: dict = {}, **xr_args: Any):
                 + "Use filter_dict to select a subset of the data."
             )
 
+        # ensure ["time", "lead_time", "variable"] ordering of required_dims (other dims unaffected)
+        dims = list(self.da.dims)
```
NickGeneva, line 140: Kinda confused by this block does the following logic work?

```python
required_dims = ["time", "lead_time", "variable"]
dims = [dim for dim in self.da.dims if dim not in required_dims]

time_index = list(self.da.dims).index("time")
dims = dims[:time_index] + required_dims + dims[time_index:]
self.da = self.da.transpose(*dims)
 ```

NickGeneva: So something like [dim_a, time, dim_b, lead_time, variable, dim_c]

Would become [dim_a, time, lead_time, variable, dim_b, dim_c]

NickGeneva (approved):
Looks good, one comment / clarification needed. But approving to unblock

earth2studio/data/xr.py
jleinonen: The intention was to permute `required_dims` to the specified order while leaving the position of other dims unchanged. So `["dim_a", "variable", "dim_b", "lead_time", "time", "dim_c"]` would become `["dim_a", "time", "dim_b", "lead_time", "variable", "dim_c"]`. But I'm not sure what is the most logical thing to in cases where there are other dimensions between the required dimensions (which is probably usually not the case). Let me know if you have an opinion on that.

NickGeneva: Ah I see now, I dont have any particular opinion on this. I'm not sure there will be many cases where this would be of concern. So all good.

Thanks for clarification

PR #524, opened 2025-10-15
Fea distilled cbottle with window function

loliverhennigh:
# Earth2Studio Pull Request

### DIFF
--- a/earth2studio/models/px/cbottle_video.py
+++ b/earth2studio/models/px/cbottle_video.py
@@ -267,7 +267,7 @@ class CBottleVideo(torch.nn.Module, AutoModelMixin, PrognosticMixin):
         out, _ = self.core_model.sample(input_batch, seed=self.seed)
         # Regrid if needed
         if self.lat_lon:
-            out = self.output_regridder(out.double())
+            out = self.output_regridder(out.contiguous().double())
         # [time, vars, lead, ...] -> [time, lead, vars, ...]
         out = out.transpose(1, 2)
 


earth2studio/models/dx/cbottle_sr.py
```diff
@@ -403,126 +364,135 @@ def load_model(
...
+            sr_model,
             lat_lon=lat_lon,
             output_resolution=output_resolution,
             super_resolution_window=super_resolution_window,
-            sampler_steps=sampler_steps,
-            sigma_max=sigma_max,
             seed=seed,
         )
 
+    @staticmethod
```
NickGeneva: AI function

earth2studio/models/dx/cbottle_sr.py
```diff
@@ -390,10 +347,14 @@ def load_model(
...
         lat_lon: bool = True,
         output_resolution: tuple[int, int] = (2161, 4320),
-        super_resolution_window: None | tuple[int, int, int, int] = None,
+        super_resolution_window: tuple[int, int, int, int] | None = None,
         sampler_steps: int = 18,
         sigma_max: int = 800,
+        distilled_model: bool = False,
+        window_function: str | None = None,
+        window_alpha: float | None = None,
         seed: int | None = None,
```
NickGeneva, line 389: move seed up before distilled_model so no api breaks

earth2studio/models/dx/cbottle_sr.py
```diff
@@ -403,126 +384,149 @@ def load_model(
...
             Random generator seed for latent variables, by default None
+        device : str | torch.device, optional
+            Device on which to instantiate the underlying cBottle model, by default CUDA when available
 
         Returns
         -------
         DiagnosticModel
             Diagnostic model
         """
+        if device is None:
```
NickGeneva: Nit: Just set the device as type string or torch.device

Then do device = torch.device(device) and set default to cpu, no need fr the static method

So consistent with SFNO

earth2studio/models/dx/cbottle_sr.py
```diff
@@ -403,126 +384,149 @@ def load_model(
...
+
+        if distilled_model:
+            if window_function is None:
+                window_function = "uniform"
+            if window_alpha is None:
+                window_alpha = 1.0
+            load_kwargs["window_function"] = window_function
+            load_kwargs["window_alpha"] = window_alpha
+        else:
+            if window_function is not None or window_alpha is not None:
```
NickGeneva: Why would you want to run without the window function if distilled? Should these just get turned on if people are using the distilled model?

NickGeneva: If this is how the distilled model should be ran, then I'm fine with just fixing it here and removing the options from the load_package... no need to expose those controls if they arent needed

earth2studio/models/dx/cbottle_sr.py
```diff
@@ -101,130 +110,100 @@ class CBottleSR(torch.nn.Module, AutoModelMixin):
...
+        and will only apply super-resolution to the specified window. For lat/lon output, the result will
+        just be returned for the specified window with the specified output resolution.
     seed : int, optional
         Random generator seed for latent variables, by default None
     """
 
     def __init__(
         self,
-        core_model: torch.nn.Module,
+        sr_model: SuperResolutionModel,
```
NickGeneva: I would leave this torch module... will cause lint / tpye problems since this is a CBottle type thats an optional import with mypy

earth2studio/models/dx/cbottle_sr.py
```diff
@@ -101,130 +110,100 @@ class CBottleSR(torch.nn.Module, AutoModelMixin):
...
     output_resolution : Tuple[int, int], optional
         High-resolution output dimensions for lat/lon output. Only used when
         lat_lon=True, by default (2161, 4320)
-    super_resolution_window : Union[None, Tuple[int, int, int, int]], optional
-        Super-resolution window. If None, super-resolution is done
-        on the entire global grid. If provided, the super-resolution window is a tuple
-        of (lat_south, lon_west, lat_north, lon_east) and will only apply super-resolution
-        to the specified window. For lat/lon output, the result will just be returned for
-        the specified window with the specified output resolution.
-    sampler_steps : int, optional
```
NickGeneva, line 119: Sampler steps should stay and remain controllable on the fly after the model is loaded... every forward pass you should set the num steps of the superres model to this

earth2studio/models/dx/cbottle_sr.py
```diff
@@ -101,130 +110,100 @@ class CBottleSR(torch.nn.Module, AutoModelMixin):
...
-        Number of diffusion steps, by default 18
-    sigma_max : int, optional
```
NickGeneva, line 121: Same deal as sampler_steps

E.g. https://github.com/NVIDIA/earth2studio/blob/main/earth2studio/models/dx/cbottle_tc.py#L435

NickGeneva:
Just some nits, approving to unblock

NickGeneva (approved):

earth2studio/models/dx/cbottle_sr.py
NickGeneva: This is important for debugging / speed control and CI testing

NickGeneva:
The pipeline `36704839` cleared, something is broken with the watch dog... so merging

PR #531, opened 2025-10-28
CMIP6 Multi-Realm Support with Automatic Regridding

gertln:
# Earth2Studio Pull Request

## Description
This PR extends CMIP6 data source capabilities by introducing a new CMIP6_multi_realm class that enables access to variables across multiple CMIP6 realms (atmosphere, ocean, sea ice) in a single interface. Previously, users could only access variables from one realm at a time; now they can request combinations of atmospheric, oceanic, and sea ice variables together.

Multi-Realm Data Access
- New CMIP6_multi_realm class allows combining multiple CMIP6 data sources from different realms
- Variables can now be requested from atmosphere (Amon), ocean (Omon), and sea ice (SImon) realms simultaneously
- Automatic variable routing - each variable is fetched from the first available source in the provided list

Automatic Regridding
- Automatic interpolation from curvilinear grids to common equirectangular grid
- Configurable interpolation method

Other Improvements
- Fixed lexicon units
- CMIP6 variables are returned in their native unit and can be also requested by using native names

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Moved StormCast and DLESyM checkpoints to Huggingface
+
 ### Deprecated
 
 ### Removed
--- a/earth2studio/models/px/dlesym.py
+++ b/earth2studio/models/px/dlesym.py
@@ -105,6 +105,7 @@ class DLESyM(torch.nn.Module, AutoModelMixin, PrognosticMixin):
 
     - https://arxiv.org/abs/2409.16247
     - https://arxiv.org/abs/2311.06253v2
+    - https://huggingface.co/nvidia/dlesym-v1-era5
 
     For more information about the HEALPix grid see:
 
@@ -356,7 +357,7 @@ class DLESyM(torch.nn.Module, AutoModelMixin, PrognosticMixin):
     def load_default_package(cls) -> Package:
         """Default DLESyM model package on NGC"""
         package = Package(
-            "ngc://models/nvidia/earth-2/dlesym-v1-era5@1.0.1",
+            "hf://nvidia/dlesym-v1-era5",
             cache_options={
                 "cache_storage": Package.default_cache("dlesym"),
                 "same_names": True,
--- a/earth2studio/models/px/stormcast.py
+++ b/earth2studio/models/px/stormcast.py
@@ -97,6 +97,7 @@ class StormCast(torch.nn.Module, AutoModelMixin, PrognosticMixin):
     For more information see the following references:
 
     - https://arxiv.org/abs/2408.10958
+    - https://huggingface.co/nvidia/stormcast-v1-era5-hrrr
 
     Parameters
     ----------
@@ -243,7 +244,7 @@ class StormCast(torch.nn.Module, AutoModelMixin, PrognosticMixin):
     def load_default_package(cls) -> Package:
         """Load prognostic package"""
         package = Package(
-            "ngc://models/nvidia/earth-2/stormcast-v1-era5-hrrr@1.0.2",
+            "hf://nvidia/stormcast-v1-era5-hrrr",
             cache_options={
                 "cache_storage": Package.default_cache("stormcast"),
                 "same_names": True,
--- a/test/models/px/test_dlesym.py
+++ b/test/models/px/test_dlesym.py
@@ -300,17 +300,11 @@ def test_dlesym_iterator(device, grid_type, batch_size):
         )
 
 
-@pytest.fixture(scope="function")
-def model() -> DLESyM:
-    package = DLESyM.load_default_package()
-    p = DLESyM.load_model(package)
-    return p
-
-
 @pytest.mark.package
 @pytest.mark.parametrize("device", ["cuda:0"])
-def test_dlesym_package(device, model):
+def test_dlesym_package(device):
     torch.cuda.empty_cache()
+    model = DLESyM.load_model(DLESyM.load_default_package())
     model = model.to(device)
 
     nside = 64


earth2studio/data/cmip6.py
```diff
@@ -465,3 +532,205 @@ def _convert_times_to_datetime(
                 )
 
         return converted
+
+
+@check_optional_dependencies()
+class CMIP6_multi_realm:
```
CharlelieLrt: Just a detail: prefer `PascalCase` for classes

earth2studio/data/cmip6.py
```diff
@@ -465,3 +532,205 @@ def _convert_times_to_datetime(
...
+
+    Note
+    ----
+    When multiple sources have different grids, curvilinear grids (e.g., from ocean
+    or sea ice models) will be interpolated to the first regular lat/lon grid found
+    using nearest-neighbor interpolation.
+    """
+
+    def __init__(self, cmip6_source_list: list[CMIP6]):
+        self.cmip6_source_list = cmip6_source_list
```
CharlelieLrt, line 650: To be safe, would be better with some argument validation to make sure that each item of the list is indeed an instance of `CMIP6`

earth2studio/data/cmip6.py
```diff
@@ -465,3 +532,205 @@ def _convert_times_to_datetime(
...
+        combines all variables into a single DataArray.
+
+        Parameters
+        ----------
+        time : datetime | list[datetime] | TimeArray
+            Timestamps to return data for.
+        variable : str | list[str] | VariableArray
+            Variable(s) to retrieve. Each variable will be fetched from the first
+            source in the list that has it available.
+        interp_method : str, optional
```
CharlelieLrt: When you have a finite number of choice a literal type int such as `'nearest' | 'linear' | 'cubic'` would be more informative

earth2studio/data/cmip6.py
CharlelieLrt: Just a general suggestion about the `CMIP6_multi_realm` class: its public API slightly differs from that of the single-source `CMIP6`, because the latter has a public method `CMIP6.available()` to check if a requested timestamp is available in the ESGF archive.

Nothing very severe, but IMO a user would expect this API to be also available in the `CMIP6_multi_realm` data source.

This comment also applies to some public attributes like `available_variables`, that exists in `CMIP6` but not in `CMIP6_multi_realm`. Having slightly more uniform APIs between the two would greatly help.

earth2studio/data/cmip6.py
gertln: Fixed in eb2ac03f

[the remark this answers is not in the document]
gertln: Fixed in eb2ac03f

earth2studio/data/cmip6.py
```diff
@@ -465,3 +532,205 @@ def _convert_times_to_datetime(
...
+                continue
+
+            da_list.append(cmip6_source(time, var_todo))
+            var_done.extend(var_todo)
+
+        # Regrid all data arrays to a common grid if needed
+        da_list = self._regrid_to_common_grid(da_list, interp_method)
+
+        # Combine all data arrays along variable dimension
+        return xr.concat(da_list, dim="variable")
```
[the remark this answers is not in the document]
gertln: Fixed in eb2ac03f

earth2studio/data/cmip6.py
```diff
@@ -465,3 +532,205 @@ def _convert_times_to_datetime(
...
+
+        Returns
+        -------
+        xr.DataArray
+            Regridded data array on regular grid
+        """
+        # Get source curvilinear coordinates
+        source_lons = da_curvilinear["_lon"].values.flatten()
+        source_lats = da_curvilinear["_lat"].values.flatten()
+        points = np.column_stack((source_lons, source_lats))
```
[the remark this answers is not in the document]
gertln: Fixed in eb2ac03f

earth2studio/data/cmip6.py
```diff
@@ -465,3 +532,205 @@ def _convert_times_to_datetime(
...
+        for i_time in range(n_time):
+            for i_var in range(n_var):
+                slice_values = values[i_time, i_var, :, :].flatten()
+
+                regridded_data[i_time, i_var, :, :] = griddata(
+                    points,
+                    slice_values,
+                    (target_lon_grid, target_lat_grid),
+                    method=interp_method,
+                )
```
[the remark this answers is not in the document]
gertln: removed support for linear and cubic because it is currently not needed and to keep things simple

earth2studio/data/cmip6.py
gertln: Fixed in eb2ac03f

earth2studio/data/cmip6.py
gertln: Good point. Thanks.

earth2studio/data/cmip6.py
gertln: Added public method `available()` and public attribute `available_variables` in eb2ac03f

earth2studio/data/cmip6.py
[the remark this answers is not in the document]
CharlelieLrt: Maybe I am missing something in the logic in the `__call__`, but let's assume that one of the requested variables in `variable` is not found in _any_ of the CMIP6 sources, then this variable is silently ignored, correct?
It would be better to raise an error or at least a warning to be more explicit.

earth2studio/data/cmip6.py
CharlelieLrt: @gertln in the same line, I just realize that there is also the `cache` property that is exposed in other data sources, but not this one. It would be good to add it. The only difference is there it would return multiple cache locations...

Actually I just realized, if the multiple CMIP6 sources passed to the constructor don't have an explicitly set cache location, then they would all default to the same location, correct? Isn't it a potential source of conflicts?

earth2studio/data/cmip6.py
```diff
@@ -465,3 +569,308 @@ def _convert_times_to_datetime(
...
+        # Find the target regular grid (first one with 'lat'/'lon' coords)
+        target_idx = None
+        for idx, da in enumerate(da_list):
+            if "lat" in da.coords and "lon" in da.coords:
+                target_idx = idx
+                break
+
+        if target_idx is None:
+            # All grids are curvilinear - use first one as-is
+            return da_list
```
CharlelieLrt: This behavior (when no regular grid is found) should be explicitly detailed in the doctsring of the class of of the `__call__` method, otherwise it might lead to silent unexpected behaviors for the user.

earth2studio/data/cmip6.py
```diff
@@ -465,3 +569,308 @@ def _convert_times_to_datetime(
...
+        list[xr.DataArray]
+            List of regridded data arrays on common grid
+        """
+        if len(da_list) == 1:
+            return da_list
+
+        # Find the target regular grid (first one with 'lat'/'lon' coords)
+        target_idx = None
+        for idx, da in enumerate(da_list):
+            if "lat" in da.coords and "lon" in da.coords:
```
CharlelieLrt, line 849: Genuine question: in the `CMIP6` datasource, I see that the case where the coordinates are called `latitude` and `longitude` is also considered. Should it be done here as well?

earth2studio/lexicon/cmip6.py
```diff
@@ -38,48 +38,70 @@ def build_vocab() -> dict[str, tuple[str, int]]:
...
-            "v10m": ("vas", -1),  # northward 10-m wind
-            "t2m": ("tas", -1),  # 2-m air temperature
-            "q2m": ("huss", -1),  # surface specific humidity
-            "msl": ("psl", -1),  # mean sea-level pressure
+            "u10m": ("uas", -1),  # eastward 10-m wind (m s-1)
+            "v10m": ("vas", -1),  # northward 10-m wind (m s-1)
+            "uas": ("uas", -1),  # eastward near-surface wind (m s-1)
+            "vas": ("vas", -1),  # northward near-surface wind (m s-1)
+            "t2m": ("tas", -1),  # near-surface air temperature (K)
+            "tas": ("tas", -1),  # near-surface air temperature (K)
```
CharlelieLrt, line 46: I'm a little confused, it's not a 1:1 mapping? I see both `"t2m"` and `"tas"` mapping to `"tas"`. Could you confirm it's allowed and indeed the intended behavior?

gertln: We mapped ERA5 variables to CMIP6 and introduced aliases for the CMIP6 variables. I find the process can be confusing and occasionally misleading because there is not always a clear mapping from CMIP6 to ERA5. Therefore, I believe users should have the option to work directly with the original CMIP6 variable names. I kept ERA5 aliases for convenience and backwards compatibility.

earth2studio/lexicon/cmip6.py
```diff
@@ -130,7 +152,16 @@ def mod(x: np.ndarray) -> np.ndarray:  # noqa: D401
                     return x + 273.15
                 return x
 
-        elif cmip6_key[0] in ["zg", "z", "orog"]:
+        elif cmip6_key[0] in ["siconc", "clt", "r2m", "hurs", "hursmax", "hursmin"]:
```
CharlelieLrt: Why removing `"zg", "z", "orogo"` from the variables that need this modifier function? It seems like this is a backward-compatibility breaking change

earth2studio/lexicon/cmip6.py
CharlelieLrt: Ok, I see. @dallasfoster could you chime in and confirm it's okay to do that? I lack experience with the lexicon to make this call

earth2studio/data/cmip6.py
gertln: I added cache() to the new dataset and it returns a list.

Multiple CMIP6 instances can safely share the same base cache directory because the intake-esgf package automatically creates subdirectories for each unique dataset, organized by project, institution, model, experiment, variant, table, variable, grid, and version. This should prevent cache conflicts between different CMIP6 sources.

For example, files are cached in paths like:
        ``<cache>/CMIP6/ScenarioMIP/CCCma/CanESM5/ssp585/r1i1p2f1/day/tas/gn/v20190429/clt_day_CanESM5_ssp585_r1i1p2f1_gn_20150101-21001231.nc``

CharlelieLrt: Ok good to know, thanks for the explanations!

earth2studio/data/cmip6.py
gertln: The variable name normalization happens in CMIP6. The latitude/longitude names should never make it to the CMIP6MultiRealm layer. So I think lat, lon is enough here.

earth2studio/lexicon/cmip6.py
gertln: `z` and `orog` still have a modifier function but I removed the modifier function for `zg `which is a CMIP6 native name and should also keep the CMIP6 native unit which is m.

earth2studio/data/cmip6.py
gertln: This is an edge case which is currently not required. I raise a NotImplementedError in this case now 
https://github.com/NVIDIA/earth2studio/pull/531/commits/fd6e13271ea24054e79a60e5a9a1e493df6c8d8f

earth2studio/lexicon/cmip6.py
CharlelieLrt: @dallasfoster confirmed it's fine.

earth2studio/data/cmip6.py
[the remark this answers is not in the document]
gertln: added check on which variables are missing

earth2studio/lexicon/cmip6.py
```diff
@@ -38,48 +38,70 @@ def build_vocab() -> dict[str, tuple[str, int]]:
...
-            "t": "ta",  # air temperature
-            "z": "zg",  # geopotential height
+            "u": "ua",  # eastward wind (m s-1)
+            "ua": "ua",  # eastward wind (m s-1)
+            "v": "va",  # northward wind (m s-1)
+            "va": "va",  # northward wind (m s-1)
+            "t": "ta",  # air temperature (K)
+            "ta": "ta",  # air temperature (K)
+            "z": "zg",  # geopotential height (m2 s-2)
+            "zg": "zg",  # geopotential height (m)
```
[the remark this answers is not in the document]
gertln: z and zg use different conversion functions

earth2studio/lexicon/cmip6.py
```diff
@@ -130,7 +152,16 @@ def mod(x: np.ndarray) -> np.ndarray:  # noqa: D401
                     return x + 273.15
                 return x
 
-        elif cmip6_key[0] in ["zg", "z", "orog"]:
+        elif cmip6_key[0] in ["siconc", "clt", "r2m", "hurs", "hursmax", "hursmin"]:
+
+            def mod(x: np.ndarray) -> np.ndarray:  # noqa: D401
+                """concentration from fraction to percentage if it is less than 1."""
```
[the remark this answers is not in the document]
gertln: fixed in https://github.com/NVIDIA/earth2studio/pull/531/commits/bcf0277c8a1e9dfdf7747e96a1d789b1d87733f3

earth2studio/lexicon/cmip6.py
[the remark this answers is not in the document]
gertln: fixed in https://github.com/NVIDIA/earth2studio/pull/531/commits/bcf0277c8a1e9dfdf7747e96a1d789b1d87733f3

earth2studio/data/cmip6.py
```diff
@@ -465,3 +589,346 @@ def _convert_times_to_datetime(
...
+                target_idx = idx
+                break
+
+        if target_idx is None:
+            raise NotImplementedError(
+                "Regridding between multiple curvilinear grids is not yet supported. "
+                "All CMIP6 sources use curvilinear grids. "
+                "At least one source with a regular lat/lon grid"
+                "is required to serve as the target grid for interpolation. "
+            )
```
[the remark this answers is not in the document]
gertln: fixed in https://github.com/NVIDIA/earth2studio/pull/531/commits/bcf0277c8a1e9dfdf7747e96a1d789b1d87733f3

test/data/test_cmip6.py
```diff
@@ -178,3 +179,392 @@ def test_cmip6_available(time, source_id, expected):
...
+        table_id="Amon",
+        variant_label="r1i1p2f1",
+        exact_time_match=True,
+    )
+
+    # Request a time that's unlikely to match exactly (mid-month, specific time)
+    time = datetime(2015, 1, 15, 12, 30, 45)
+
+    with pytest.raises(ValueError, match="Exact time match required"):
+        _ = ds(time, ["u10m"])
```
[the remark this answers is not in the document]
gertln: using the year 1800 now

test/data/test_cmip6.py
```diff
@@ -178,3 +179,392 @@ def test_cmip6_available(time, source_id, expected):
...
+        source_id="CanESM5",
+        table_id="Omon",
+        variant_label="r1i1p2f1",
+    )
+
+    multi = CMIP6MultiRealm([atmos, ocean])
+
+    time = datetime(2015, 1, 16)
+    # Request variables that don't exist in any source
+    variables = ["u10m", "sst", "nonexistent_var"]
```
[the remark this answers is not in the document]
gertln: The test makes sure that partial success doesn't mask missing variables.

PR #532, opened 2025-10-28
Ngeneva/wrapper refactor

NickGeneva:
# Earth2Studio Pull Request

## Description

This is a refactor of how the diagnostic wrapper prognostic is set up.

The existing solution has a few limitations:

- Does not allow customization / control over some of the mapping parts, leading to updates to the model needed when edge cases occur
- Presently has strong requirements for lat lon grids which are not always present with new HPX models
- Internal factory set up that's a little too hidden from the end user

This PR does the following:

- Cleans up the internals of this wrapper into 4 key steps
- Exposes the intermediate remapping functions to the user, allowing for full customization in theory via dependency injection
- Removed the factory feature in favor of the user creating the hierarchy if needed
- Default methods remain close to same functionality as the original but are a little more picky on how the outputs are handled

## Checklist


earth2studio/models/px/dxwrapper.py
```diff
@@ -187,72 +210,106 @@ def _convert_to_2d(
...
-            )
-            self.interp_lat0 = lat0
-            self.interp_lon0 = lon0
-            self.interp_lat1 = lat1
-            self.interp_lon1 = lon1
+        (lat0, lon0) = _convert_to_2d(px_coords["lat"], px_coords["lon"])
+        (lat1, lon1) = _convert_to_2d(dx_coords["lat"], dx_coords["lon"])
+
+        if not hasattr(self, "interp") or self.interp is None:  # type: ignore
+            self.interp = LatLonInterpolation(lat0, lon0, lat1, lon1).to(x.device)
```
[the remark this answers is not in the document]
NickGeneva: No

earth2studio/models/px/dxwrapper.py
```diff
@@ -187,72 +213,106 @@ def _convert_to_2d(
...
```
swbg: Why do we reuse interp for other dx_coords? Wouldn't we possibly need one interp per dx_model?

jleinonen: By the way, this kind of complications were why the current class design uses recursion rather than explicitly iterating through a list. It leads to simpler code within the class when you can assume there is only one diagnostic model. Of course that also doesn't work if you have different interpolation etc. for the different models.

If you think the recursive method is confusing, we can drop that in favor of supporting only one diagnostic model at a time and requiring the user to use explicit composition.

NickGeneva: I think we are thinking about supporting two different patterns:

Presently the wrapper does (assuming that dx1 produces same domain coords as px)

```
px -> dx1 -> dx2
```

I am trying to just support:

```
px -> dx1
   |-> dx2
```

So the depth is still one. For something like px -> dx1 -> dx2 with will be on the use to build Wapper(Wrapper(px, dx1), dx2) then selves. I prefer to force the end user to instantiate the objects themselves so they are fully aware of what the object stack looks like.


@swbg is correct I need to support the multiple interp functions

NickGeneva:
@jleinonen 

The interaction with the cases you listed should basically be the same with the following differences

1. The difference is that, if the prognostic / diagnostic have the same domain coords, they are concated. If not, just the dx are returned. The toggle is forfeited for full customization.

2. To have a chain of models, its now on the user to do this. E.g.

```python
wrapped_model = DiagnosticWrapper(
        px_model=px_model, dx_model=[sp_model, ws_model]
    ).to(device=device)
wrapped_model = DiagnosticWrapper(px_model=wrapped_model, dx_model=[dx_model]).to(
    device=device
)
```

The tests are largely the same (although I removed / trimmed some down for speed). The key is that now, if you want that behavior to change, they user can now do that entirely with the callables `prepare_dx_input_tensor` and `prepare_output_tensor`. 

So now the call functions is simple, and the user just needs to follow the protocol when something needs to change.

```python
px_x, px_coords = self.px_model(x, coords)
dx_x = []
dx_coords = []
for model, prepare_dx_input in zip(self.dx_model, self.prepare_dx_input_tensor):
    dx_x0, dx_coords0 = prepare_dx_input(
        px_x, px_coords.copy(), model.input_coords()
    )
    dx_x0, dx_coords0 = model(dx_x0, dx_coords0)
    dx_x.append(dx_x0)
    dx_coords.append(dx_coords0)
x, coords = self.prepare_output_tensor(px_x, px_coords, dx_x, dx_coords)

```

Does this help?

earth2studio/models/px/dxwrapper.py
NickGeneva: This is now updated, for these occurrences

jleinonen:
@NickGeneva 

> 1. The difference is that, if the prognostic / diagnostic have the same domain coords, they are concated. If not, just the dx are returned. The toggle is forfeited for full customization.

If it's required that prognostic and diagnostic have the same domain coords, then we can't, for example, concatenate a (721, 1440) SFNO prognostic to a (720, 1440) precipitation diagnostic. The current implementation is subtly different: it can do the concatenation if the _input of the diagnostic_ has the same domain as the _output of the diagnostic_. This way, the logic is:

```
prognostic_output = px(prognostic_input)
interpolated_prognostic_output = interp(prognostic_output)
diagnostic_output = dx(interpolated_prognostic_output)
final_output = concatenate(interpolated_prognostic_output, diagnostic_output)
```

Would it be okay to default to this behavior? I think this makes sense as the default as it allows us to work interchangeably with the 720/721 row models.

earth2studio/models/px/dxwrapper.py
```diff
@@ -274,7 +425,52 @@ def __call__(
...
+        dx_x = []
+        dx_coords = []
+        for model, prepare_dx_input in zip(self.dx_model, self.prepare_dx_input_tensor):
+            dx_x0, dx_coords0 = prepare_dx_input(
+                px_x, px_coords.copy(), model.input_coords()
+            )
+            dx_x0, dx_coords0 = model(dx_x0, dx_coords0)
+            dx_x.append(dx_x0)
+            dx_coords.append(dx_coords0)
+        x, coords = self.prepare_output_tensor(px_x, px_coords, dx_x, dx_coords)
```
jleinonen, line 509: This looks identical to the loop in `_default_generator` so it could be factored into its own function.

earth2studio/models/px/dxwrapper.py
```diff
@@ -14,89 +14,362 @@
...
+            else:
+                return (lat, lon)
+
+        (lat0, lon0) = _convert_to_2d(px_coords["lat"], px_coords["lon"])
+        (lat1, lon1) = _convert_to_2d(dx_coords["lat"], dx_coords["lon"])
+
+        if self.interp is None:
+            self.interp = LatLonInterpolation(lat0, lon0, lat1, lon1).to(x.device)
+
+        x = self.interp(x)
```
jleinonen, line 158: Are we always using `self.interp` (a `LatLonInterpolation` object) here? In the current version we have https://github.com/NVIDIA/earth2studio/blob/37e74e54244f5d1f213beb16791d4e9369fc3ca9/earth2studio/models/px/dxwrapper.py#L233-L234. Interpolation will probably work even if we're mapping between identical/similar grids (that should be tested though) but it'll be considerably slower than `map_coords` in cases where `map_coords` would suffice.

NickGeneva: Added a full module skip if it sees that lat lons are the same

NickGeneva:
@jleinonen 

(ugh greptile not using the json and spamming the PR...)

Okay so I added three strategies for the output:
1. Attempt to concat px outputs and all dx outputs
2. Attempt to concat lat/lon sub-region (contiguous) px outputs and all dx outputs (this is your case with the 720)
3. Concat just dx outputs

I added (with cursors help) quite a few tests targeting this module since its a little more complicated to meet these needs.

NickGeneva:
Sample script:

```python
# %%
import os
from datetime import datetime

import numpy as np
import torch

from earth2studio.data import ARCO
from earth2studio.io import ZarrBackend
from earth2studio.models.dx import PrecipitationAFNOv2
from earth2studio.models.px import DiagnosticWrapper, SFNO
from earth2studio.run import deterministic

sfno_package = SFNO.load_default_package()
sfno_model = SFNO.load_model(sfno_package)

precip_package = PrecipitationAFNOv2.load_default_package()
precip_model = PrecipitationAFNOv2.load_model(precip_package)

wrapped_model = DiagnosticWrapper(
    px_model=sfno_model,
    dx_model=precip_model,
)

data_source = ARCO()

# Define time range for forecast
time = datetime(2023, 1, 15, 0)
nsteps = 3  # Number of forecast steps

io = ZarrBackend(file_name="test.zarr")
deterministic(
    time=[time],
    nsteps=nsteps,
    prognostic=wrapped_model,
    data=data_source,
    io=io,
    device=torch.device("cuda" if torch.cuda.is_available() else "cpu"),
)

import xarray as xr
ds = xr.open_zarr("test.zarr")
print(ds)

import cartopy.crs as ccrs

# Extract precipitation data
precip_data = ds["tp06"].isel(time=0)  # First time step
lat = ds["lat"].values
lon = ds["lon"].values

# Create plot
fig, axes = plt.subplots(
    2, 2, figsize=(15, 10), subplot_kw={"projection": ccrs.PlateCarree()}
)
axes = axes.flatten()

# Plot 4 different forecast steps

# Plot 4 different forecast steps
import xarray as xr
ds = xr.open_zarr("test.zarr")
print(ds)

import cartopy.crs as ccrs

# Extract precipitation data
precip_data = ds["tp06"].isel(time=0)  # First time step
lat = ds["lat"].values
lon = ds["lon"].values

# Create plot
fig, axes = plt.subplots(
    2, 2, figsize=(15, 10), subplot_kw={"projection": ccrs.PlateCarree()}
)
axes = axes.flatten()

for var, cmap, vmax in [("tp06", "YlGnBu", 0.005), ("t2m", "turbo", 320)]:
    # Plot 4 different forecast steps
    for idx, step in enumerate([0, 1, 2, 3]):
        ax = axes[idx]
        precip = ds[var].isel(time=0, lead_time=step)

        # Plot precipitation
        im = ax.pcolormesh(
            lon,
            lat,
            precip.values,
            transform=ccrs.PlateCarree(),
            cmap=cmap,
            vmax=vmax,
        )

        # Add features
        ax.coastlines(linewidth=0.5)
        ax.add_feature(cfeature.BORDERS, linewidth=0.5)
        ax.add_feature(cfeature.STATES, linewidth=0.3, edgecolor="gray")

    plt.tight_layout()
    plt.savefig(f"test{var}.png", dpi=150, bbox_inches="tight")
```

```text
Dimensions:    (time: 1, lead_time: 4, lat: 720, lon: 1440)
Coordinates:
  * time       (time) datetime64[ns] 8B 2023-01-15
  * lead_time  (lead_time) timedelta64[h] 32B 00:00:00 06:00:00 ... 18:00:00
  * lat        (lat) float64 6kB 90.0 89.75 89.5 89.25 ... -89.25 -89.5 -89.75
  * lon        (lon) float64 12kB 0.0 0.25 0.5 0.75 ... 359.0 359.2 359.5 359.8
Data variables: (12/74)
    z400       (time, lead_time, lat, lon) float32 17MB dask.array<chunksize=(1, 4, 720, 1440), meta=np.ndarray>
    z600       (time, lead_time, lat, lon) float32 17MB dask.array<chunksize=(1, 4, 720, 1440), meta=np.ndarray>
    v700       (time, lead_time, lat, lon) float32 17MB dask.array<chunksize=(1, 4, 720, 1440), meta=np.ndarray>
    u600       (time, lead_time, lat, lon) float32 17MB dask.array<chunksize=(1, 4, 720, 1440), meta=np.ndarray>
    t300       (time, lead_time, lat, lon) float32 17MB dask.array<chunksize=(1, 4, 720, 1440), meta=np.ndarray>
    u10m       (time, lead_time, lat, lon) float32 17MB dask.array<chunksize=(1, 4, 720, 1440), meta=np.ndarray>
    ...         ...
    t200       (time, lead_time, lat, lon) float32 17MB dask.array<chunksize=(1, 4, 720, 1440), meta=np.ndarray>
    z250       (time, lead_time, lat, lon) float32 17MB dask.array<chunksize=(1, 4, 720, 1440), meta=np.ndarray>
    v300       (time, lead_time, lat, lon) float32 17MB dask.array<chunksize=(1, 4, 720, 1440), meta=np.ndarray>
    q200       (time, lead_time, lat, lon) float32 17MB dask.array<chunksize=(1, 4, 720, 1440), meta=np.ndarray>
    t150       (time, lead_time, lat, lon) float32 17MB dask.array<chunksize=(1, 4, 720, 1440), meta=np.ndarray>
    z300       (time, lead_time, lat, lon) float32 17MB dask.array<chunksize=(1, 4, 720, 1440), meta=np.ndarray>
```

image

image

swbg:
In my opinion, the following two things should work out of the box but I think they currently don't:

Chaining FCN - PrecipitationAFNOv2 results in:

```
Traceback (most recent call last):
  File "/workspace/repos/earth2studio/earth2studio/utils/coords.py", line 63, in handshake_dim
    input_dims[required_index]
    ~~~~~~~~~~^^^^^^^^^^^^^^^^
IndexError: list index out of range

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/workspace/repos/earth2studio-data/scripts/test_dxwrapper.py", line 35, in <module>
    main()
  File "/workspace/repos/earth2studio-data/scripts/test_dxwrapper.py", line 25, in main
    run.deterministic(
  File "/workspace/repos/earth2studio/earth2studio/run.py", line 110, in deterministic
    total_coords = prognostic.output_coords(prognostic.input_coords()).copy()
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/repos/earth2studio/earth2studio/models/batch.py", line 336, in _wrapper
    out_coords = func(model, flatten_coords)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/repos/earth2studio/earth2studio/models/px/dxwrapper.py", line 459, in output_coords
    dx_coords.append(model.output_coords(coords))
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/repos/earth2studio/earth2studio/models/batch.py", line 336, in _wrapper
    out_coords = func(model, flatten_coords)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/repos/earth2studio/earth2studio/models/dx/precipitation_afno_v2.py", line 150, in output_coords
    handshake_dim(input_coords, "lon", 5)
  File "/workspace/repos/earth2studio/earth2studio/utils/coords.py", line 65, in handshake_dim
    raise ValueError(
ValueError: Required index 5 outside dimensionality of input coordinate system of 5
```

I believe the issue here may however be FCN (which does not have time in the input/output coords) rather than the diagnostic wrapper implementation.

Chaining SFNO - PrecipitationAFNOv2 - SolarRadiationAFNO1H (or the other way round) results in:

```
2025-11-05 13:55:01.851 | ERROR    | earth2studio.models.px.dxwrapper:forward:270 - Failed to concatenate outputs of diagnostic models. The outputs of the models cannot be concatenated.
Running inference:   0%|                                                                                                                                                                                       | 0/4 [00:20<?, ?it/s]
Traceback (most recent call last):                                                                                                                                                                             | 0/4 [00:20<?, ?it/s]
  File "/workspace/repos/earth2studio-data/scripts/test_dxwrapper.py", line 36, in <module>
    main()
  File "/workspace/repos/earth2studio-data/scripts/test_dxwrapper.py", line 26, in main
    run.deterministic(
  File "/workspace/repos/earth2studio/earth2studio/run.py", line 138, in deterministic
    for step, (x, coords) in enumerate(model):
  File "/workspace/repos/earth2studio/earth2studio/models/px/dxwrapper.py", line 549, in create_iterator
    yield from self._default_generator(x, coords)
  File "/workspace/repos/earth2studio/earth2studio/models/px/dxwrapper.py", line 527, in _default_generator
    x, coords = self.prepare_output_tensor(px_x, px_coords, dx_x, dx_coords)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1751, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1762, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/repos/earth2studio/earth2studio/models/px/dxwrapper.py", line 274, in forward
    raise e
  File "/workspace/repos/earth2studio/earth2studio/models/px/dxwrapper.py", line 266, in forward
    x = torch.concat(x, dim=list(dx_target).index("variable"))
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Sizes of tensors must match except in dimension 2. Expected size 721 but got size 720 for tensor number 1 in the list.
```

I think it would be easier if the diagnostic wrapper would only support one diagnostic model. Users could then combine multiple models by stacking diagnostic wrappers, e.g.:

```
px = SFNO.load_model(SFNO.load_default_package())  # 721
dx1 = SolarRadiationAFNO1H.load_model(SolarRadiationAFNO1H.load_default_package())  # 721
dx2 = PrecipitationAFNOv2.load_model(PrecipitationAFNOv2.load_default_package())  # 720
wrapped_model = DiagnosticWrapper(px, dx1)
wrapped_model = DiagnosticWrapper(wrapped_model, dx2)
```

Above code even works with the current implementation (while passing a list of the two dx does not).

NickGeneva:
@swbg Thanks for the checks

The FCN one is indeed a bug, will fix... Actually found a very similar issue with the interpolation model recently.

For the second one, this is by design. Two dx models in the list with different output coordinates will error since I didnt want the model to be deciding which one to map to. Its on the user to manually define the mapping chain like what you did with the working example.

```
px = SFNO.load_model(SFNO.load_default_package())  # 721
dx1 = SolarRadiationAFNO1H.load_model(SolarRadiationAFNO1H.load_default_package())  # 721
dx2 = PrecipitationAFNOv2.load_model(PrecipitationAFNOv2.load_default_package())  # 720
wrapped_model = DiagnosticWrapper(px, dx1)
wrapped_model = DiagnosticWrapper(wrapped_model, dx2)
```

The reason a list is supported is because if you two models that need the same px data. See the following example:

image

So with what your referring, yes it only supports 1 deep dx model. Does not do any chaining.

NickGeneva:
@swbg The issue discussed is fixed. Although in the example you gave, it doesnt work still since FCN does not have specific humid

But it does now error correctly with some information for the user:

```python
fcn_package = FCN.load_default_package()
fcn_model = FCN.load_model(sfno_package)

precip_package = PrecipitationAFNOv2.load_default_package()
precip_model = PrecipitationAFNOv2.load_model(precip_package)

wrapped_model = DiagnosticWrapper(
    px_model=fcn_model,
    dx_model=precip_model,
)
```

````
2025-11-05 23:52:31.158 | SUCCESS  | earth2studio.run:deterministic:106 - Fetched data from GFS
2025-11-05 23:52:31.244 | INFO     | earth2studio.run:deterministic:136 - Inference starting!
2025-11-05 23:52:31.257 | ERROR    | earth2studio.models.px.dxwrapper:__call__:166 - Mapping between prognostic output and diagnostic input failed.
Running inference:   0%|                                                                                                                                                                                                                                                                                       | 0/4 [00:00<?, ?it/s]
Traceback (most recent call last):                                                                                                                                                                                                                                                                             | 0/4 [00:00<?, ?it/s]
  File "/localhome/local-ngeneva/earth2studio/test.py", line 31, in <module>
    deterministic(

    ......

  File "/localhome/local-ngeneva/earth2studio/earth2studio/utils/coords.py", line 282, in map_coords
    raise ValueError(
ValueError: For non-numeric coordinate, variable, all values of output coords must be in the input coordinates. Some elements of ['u10m' 'v10m' 't2m' 'sp' 'msl' 'tcwv' 'u500' 'u850' 'u1000' 'v500' 'v850'
 'v1000' 'z50' 'z500' 'z850' 'z1000' 't500' 't850' 'q500' 'q850'] are not in ['u10m' 'v10m' 't2m' 'sp' 'msl' 't850' 'u1000' 'v1000' 'z1000' 'u850'
 'v850' 'z850' 'u500' 'v500' 'z500' 't500' 'z50' 'r500' 'r850' 'tcwv'
 'u100m' 'v100m' 'u250' 'v250' 'z250' 't250'].
 ```

File: earth2studio/models/px/dxwrapper.py [hunk 5]
```diff
@@ -14,89 +14,423 @@
...
+        coords = px_coords.copy()
+        coords["lat"] = dx_coords["lat"]
+        coords["lon"] = dx_coords["lon"]
+        # Map remaining coords
+        try:
+            x, coords = map_coords(x, coords, dx_coords)
+        except Exception as e:
+            logger.error(
+                "Mapping between prognostic output and diagnostic input failed. Check coordinate systems are compatable."
```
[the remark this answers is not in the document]
NickGeneva: Thanks greptile

swbg:
I had not thought about the requirement for multiple diagnostics at stage 1, then a separate stage 2, etc., but that makes sense. I am happy with the functionality. I don't want to claim that I have a full grasp of all edge cases but the wrapper does what I would it expect to do and the functionality for preparing coords/tensors seems flexible enough for custom requirements.

The wrapper is a powerful tool that demonstrates how the Earth2Studio coord interface makes assembling pipelines really easy. How do we make sure users find and use it? We could:
1. Get rid of the diagnostic workflow and replace the diagnostic example using the wrapper
2. Add a new example demonstrating the functionality with at least two diagnostic models
3. ??

What do you think?

earth2studio/models/px/dxwrapper.py
```diff
@@ -14,89 +14,423 @@
...
+        CoordSystem
+            Expected output coords from model for a given time-step
+        """
+        dx_target = dx_coords[-1]
+        if _can_concat_directly(px_coords, dx_target):
+            variables = [px_coords["variable"]] + [c["variable"] for c in dx_coords]
+        elif _can_concat_with_subregion(px_coords, dx_target):
+            variables = [px_coords["variable"]] + [c["variable"] for c in dx_coords]
+        else:
+            variables = [c["variable"] for c in dx_coords]
```
swbg, line 200: Small comment here, it would be super nice to have the diagnostic wrapper log a short info whether everything will be concatenated, a subregion will be concatenated, or only the diagnostics will be concatenated. I think it is not easy to implement in a nice way (e.g., only fire one log but per diagnostic wrapper but a separate one for each diagnostic wrapper in the pipeline, etc.) but if you have a good idea it would be helpful for anticipating what the wrapper does (the wrapper logic is relatively complicated but it will usually do what the user expects, so the log statement would assure the user of their belief).

swbg (approved):

NickGeneva:
> The wrapper is a powerful tool that demonstrates how the Earth2Studio coord interface makes assembling pipelines really easy. How do we make sure users find and use it?

The idea thus far has been to keep it a bit of a silent release to make sure we get any issues sorted out (this whole 721 vs 720 deal is a good example). I think its important its communicated as a convenience tool with some core assumptions rather than a silver bullet.

The diagnostic workflow is still useful since its there as a reference for people to build from.

An example I think would be worth it, but not this release. Let it get tested a little more. I can open an issue to track that.

earth2studio/models/px/dxwrapper.py
NickGeneva: This is one of the reasons I wanted to make sure the output_coords worked as expected.

With your wrapper model `wrapper_model = DiagnosticWrapper()` just doing `wrapper_model.output_coords(wrapper_model.input_coords())`

Its a bit tricky to have some written description, because it could be doing any number of things. So my idea is users could just find out. via the coords

NickGeneva: Tried add in some way, but ended up being too messy. Similarly it became confusing when the wrappers are stacked. Leaving this for now.

NickGeneva:
Tests here are just issues with cuda and model package cache, merging

Issue #523, opened 2025-10-14
🐛[BUG]: h5py fault


NickGeneva:
### Version

main

### On which installation method(s) does this occur?

source

### Describe the issue

The most recent h5py update (3.15.0) seems to break some of our xarray loads of netcdf files in the CI environment.

Its unknown what the issue is, but for now its recommended to downgrade to 3.14.0 of h5py... 
Switching the xarray engine from `h5netcdf` -> `netcdf4` is a work around for NC files.

I have not locked the h5py range since this seems to only surface in a few tests, and the new uv.lock file should select 3.14.0 h5py. If you encounter a similar issue, downgrade h5py

Test failure looks like:

```
RuntimeError: Unspecified error in H5DSget_num_scales (return value <0)
Task exception was never retrieved
future: <Task finished name='Task-1678' coro=<tqdm_asyncio.gather.<locals>.wrap_awaitable() done, defined at /usr/local/lib/python3.12/dist-packages/tqdm/asyncio.py:75> exception=RuntimeError('Unspecified error in H5DSget_num_scales (return value <0)')>
Traceback (most recent call last):
  File "/usr/lib/python3.12/asyncio/tasks.py", line 316, in __step_run_and_handle_result
    result = coro.throw(exc)
             ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/tqdm/asyncio.py", line 76, in wrap_awaitable
    return i, await f
              ^^^^^^^
  File "/builds/modulus/earth-2/earth2studio/src/earth2studio/data/ncar.py", line 274, in fetch_wrapper
    out = await self.fetch_array(
          ^^^^^^^^^^^^^^^^^^^^^^^
  File "/builds/modulus/earth-2/earth2studio/src/earth2studio/data/ncar.py", line 336, in fetch_array
    ds = await asyncio.to_thread(
         ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/threads.py", line 25, in to_thread
    return await loop.run_in_executor(None, func_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/futures.py", line 287, in __await__
    yield self  # This tells Task to wait for completion.
    ^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/tasks.py", line 385, in __wakeup
    future.result()
[...]

PR #533, opened 2025-10-29
Add ACE2-ERA5 model

pzharrington:
# Earth2Studio Pull Request

## Description
Adds ACE2-ERA5 model: https://arxiv.org/pdf/2411.11268v1

Since the model requires forcing data inputs for each step, this PR also adds data sources that can pull the initial condition and forcing data from the ACE2 huggingface repo. I also added another forcing data source that can pull from ARCO and override the CO2 with a user-supplied callable function that generates CO2 as a function of time. 

## Checklist

### DIFF
--- a/earth2studio/models/px/ace2.py
+++ b/earth2studio/models/px/ace2.py
@@ -88,7 +88,8 @@ def _npdatetime64_to_cftime(dt64_array: np.ndarray) -> np.ndarray:
 
 def _cftime_to_npdatetime64(cftime_array: np.ndarray) -> np.ndarray:
     """Convert cftime.DatetimeProlepticGregorian array to np.datetime64[s] array
-    (vectorized-safe). Only supports up to seconds precision. Out-of-range years become NaT.
+    (vectorized-safe). Only supports up to seconds precision. Out-of-range years become
+    NaT.
     """
 
     def _convert_single(t: cftime.DatetimeProlepticGregorian) -> np.datetime64:
--- a/test/models/px/test_ace2.py
+++ b/test/models/px/test_ace2.py
@@ -20,10 +20,14 @@ import numpy as np
 import pytest
 import torch
 
-# pytest.importorskip("fme")
-from earth2studio.data import Random, fetch_data  # noqa: E402
-from earth2studio.models.px.ace2 import ACE2ERA5  # noqa: E402
-from earth2studio.utils import handshake_dim  # noqa: E402
+from earth2studio.data import Random, fetch_data
+from earth2studio.data.ace2 import ACE_GRID_LAT, ACE_GRID_LON
+from earth2studio.models.px.ace2 import (
+    ACE2ERA5,
+    _cftime_to_npdatetime64,
+    _npdatetime64_to_cftime,
+)
+from earth2studio.utils import handshake_dim
 
 
 class PhooOutput:
@@ -153,7 +157,8 @@ def test_ACE2ERA5_call(device):
     # Use a single timestamp; forcing source will handle needed adjustments
     time = np.array([np.datetime64("2001-01-01T00:00")])
 
-    p = ACE2ERA5(PhooStepper()).to(device)
+    forcing_source = Random({"lat": ACE_GRID_LAT, "lon": ACE_GRID_LON})
+    p = ACE2ERA5(PhooStepper(), forcing_source).to(device)
 
     # Build a Random data source over the model grid
     dc = p.input_coords()
@@ -184,6 +189,8 @@ def test_ACE2ERA5_call(device):
     handshake_dim(out_coords, "variable", 2)
     handshake_dim(out_coords, "lead_time", 1)
     handshake_dim(out_coords, "time", 0)
+    np.testing.assert_array_equal(out_coords["lat"], ACE_GRID_LAT)
+    np.testing.assert_array_equal(out_coords["lon"], ACE_GRID_LON)
 
 
 @pytest.mark.parametrize("batch", [1, 2])
@@ -192,7 +199,8 @@ def test_ACE2ERA5_iter(batch, device):
     torch.cuda.empty_cache()
     time = np.array([np.datetime64("2001-01-01T00:00")])
 
-    p = ACE2ERA5(PhooStepper()).to(device)
+    forcing_source = Random({"lat": ACE_GRID_LAT, "lon": ACE_GRID_LON})
+    p = ACE2ERA5(PhooStepper(), forcing_source).to(device)
 
     dc = p.input_coords()
     del dc["batch"]
@@ -224,6 +232,8 @@ def test_ACE2ERA5_iter(batch, device):
         assert (out_coords["batch"] == np.arange(batch)).all()
         assert (out_coords["time"] == time).all()
         assert out_coords["lead_time"][0] == np.timedelta64(6 * (i + 1), "h")
+        np.testing.assert_array_equal(out_coords["lat"], ACE_GRID_LAT)
+        np.testing.assert_array_equal(out_coords["lon"], ACE_GRID_LON)
         if i > 2:
             break
 
@@ -269,3 +279,41 @@ def test_ace2era5_package(device):
     handshake_dim(out_coords, "variable", 2)
     handshake_dim(out_coords, "lead_time", 1)
     handshake_dim(out_coords, "time", 0)
+
+
+def test_time_conversion_helpers_roundtrip():
+    dt_in = np.array(
+        [
+            np.datetime64("2001-01-02T03:04:05"),
+            np.datetime64("2020-12-31T23:59:59"),
+        ],
+        dtype="datetime64[s]",
+    )
+
+    cftime_arr = _npdatetime64_to_cftime(dt_in)
+    assert cftime_arr.shape == dt_in.shape
+    dt_out = _cftime_to_npdatetime64(cftime_arr)
+    np.testing.assert_array_equal(dt_out, dt_in)
+
+    # Multi-dimensional shape preserved
+    dt2 = dt_in.reshape(1, 2)
+    cf2 = _npdatetime64_to_cftime(dt2)
+    assert cf2.shape == (1, 2)
+    dt2_back = _cftime_to_npdatetime64(cf2)
+    np.testing.assert_array_equal(dt2_back, dt2)
+
+    # Out-of-range years become NaT in reverse conversion
+    try:
+        import cftime  # local import for test
+    except Exception:
+        pytest.skip("cftime not available")
+    cf_bad = np.array(
+        [
+            cftime.DatetimeProlepticGregorian(3000, 1, 1, 0, 0, 0),  # beyond supported
+            cftime.DatetimeProlepticGregorian(2005, 1, 1, 0, 0, 0),  # valid
+        ],
+        dtype=object,
+    )
+    dt_bad = _cftime_to_npdatetime64(cf_bad)
+    assert np.isnat(dt_bad[0])
+    assert dt_bad[1] == np.datetime64("2005-01-01T00:00:00")


pzharrington:
@NickGeneva FYI I also hit [this issue](https://github.com/NVIDIA/earth2studio/issues/523) so restricted the h5py version for `ace2` optional dep. Issue regarding the torch-harmonics dep has been filed with ACE2 here: https://github.com/ai2cm/ace/issues/48

This is working for me if I use the dependency overrides as specified in the updated `pyproject.toml`, and apply this fix to the fme source:
```
diff --git a/fme/sht_fix.py b/fme/sht_fix.py
index ba2484b..cce2615 100644
--- a/fme/sht_fix.py
+++ b/fme/sht_fix.py
@@ -216,5 +216,5 @@ class InverseRealSHT(nn.Module):
 
         return x
 
-torch_harmonics.RealSHT = RealSHT
-torch_harmonics.InverseRealSHT = InverseRealSHT
+#torch_harmonics.RealSHT = RealSHT
+#torch_harmonics.InverseRealSHT = InverseRealSHT
```

earth2studio/data/ace.py
```diff
@@ -0,0 +1,219 @@
...
+
+    _IC_ALLOWED_YEARS = [1940, 1950, 1979, 2001, 2020]
+
+    def __init__(
+        self,
+        base_path: str = "allenai/ACE2-ERA5",
+        mode: Literal["forcing", "initial_conditions"] = "forcing",
+        forcing_subdir: str = "forcing_data",
+        ic_subdir: str = "initial_conditions",
+        local: bool = False,
```
NickGeneva: Whats local doing versus caching the results?

earth2studio/data/ace2.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def __init__(
+        self,
+        base_path: str = "allenai/ACE2-ERA5",
+        mode: Literal["forcing", "initial_conditions"] = "forcing",
+        forcing_subdir: str = "forcing_data",
+        ic_subdir: str = "initial_conditions",
+        local: bool = False,
+        co2_fn: Callable[[Sequence[datetime]], np.ndarray] | None = None,
+    ):
```
NickGeneva, line 88: Can you update this to include the standard constructor args for data sources:

cache: bool = True,
verbose: bool = True

https://github.com/NVIDIA/earth2studio/blob/main/earth2studio/data/arco.py#L78

For setting the cache you can follow the pattern here:
https://github.com/NVIDIA/earth2studio/blob/main/earth2studio/data/arco.py#L308

https://github.com/NVIDIA/earth2studio/blob/main/earth2studio/data/arco.py#L180

earth2studio/data/ace.py
pzharrington: It's not super important...in case someone has it already downloaded, or has some custom local forcing data files they want to use instead of the official ones. They could just move those into the cache location, but I figure this would be easier to use and less prone to mishaps

earth2studio/data/ace.py
```diff
@@ -0,0 +1,219 @@
...
+    References
+    ----------
+    - ACE2-ERA5 paper: https://arxiv.org/html/2411.11268v1
+    """
+
+    _IC_ALLOWED_YEARS = [1940, 1950, 1979, 2001, 2020]
+
+    def __init__(
+        self,
+        base_path: str = "allenai/ACE2-ERA5",
```
NickGeneva: What are the odds that people will want to change `base_path` `forcing_subdir` or `ic_subdir`?

To clean up the constructor, could any of these be moved to class variables? This will a make the API doc a lot easier for people... Specific these string params seem strange to ask a user to provide... Always easier to add than subtract form APIs

I.e.

```python
class SampleDataSource:
    test = "test"
    def __init__(self):
        self.test = SampleDataSource.test

# 90% of people
ds = SampleDataSource()
print(ds.test)

# Power user
SampleDataSource.test = "test2"
ds = SampleDataSource()
print(ds.test)
````

File: earth2studio/data/ace.py [hunk 4]
```diff
@@ -0,0 +1,219 @@
...
+        # Map requested Earth2Studio variable names to FME names present in files
+        try:
+            var_list_fme = [ACELexicon[v][0] for v in var_list_e2s]
+        except KeyError as e:
+            raise KeyError(f"Unknown ACE2ERA5 variable id: {e}") from e
+
+        # Determine years to fetch
+        years = sorted({t.year for t in time_list})
+        paths: list[str] = []
+        for y in years:
```
NickGeneva: Tqdm this with the verbose option for consistency with other data sources

https://github.com/NVIDIA/earth2studio/blob/afa4105aa79f725e5eafe1a6cf8a0bcf7b2bd5c7/earth2studio/data/arco.py#L238

earth2studio/data/ace.py
```diff
@@ -0,0 +1,219 @@
...
+                filename = f"{self._forcing_subdir}/forcing_{y}.nc"
+                cache_name = "ace2era5"
+            else:
+                filename = f"{self._ic_subdir}/ic_{y}.nc"
+                cache_name = "ace2era5"
+
+            if self._local:
+                path = os.path.join(self._base_path, filename)
+            else:
+                pkg = Package(
```
NickGeneva: While convenient, this is going to say its downloading a package right. Can the hugging face file system just get used directly?

I also am not a fan of importing from the models sub module in here if possible :) (not always the case tho)

earth2studio/data/ace.py
```diff
@@ -0,0 +1,219 @@
...
+                    da.expand_dims("time")
+                    .assign_coords(time=time_list)
+                    .transpose("time", "lat", "lon")
+                )
+            elif "time" in da.dims and self._mode == "forcing":
+                # CO2 is time-only
+                da = da.expand_dims({"lat": self.lat, "lon": self.lon})
+                da = da.transpose("time", "lat", "lon")
+                if self._co2_fn is not None:
+                    co2_times = self._co2_fn(time_list)
```
NickGeneva: nit: maybe co2_data? co2_times sounds like time stamp values to me

earth2studio/models/px/ace2.py
```diff
@@ -0,0 +1,641 @@
...
+        - skt: Skin temperature
+        - z: Surface height of topography (m), invariant (from geopotential/g)
+        - land_abs: Land grid cell area fraction invariant
+        - ocean_abs: Ocean grid cell area fraction, NaN-filled to 0.0.
+        - sic_abs: Absolute sea-ice grid cell area fraction, NaN-filled to 0.0.
+        - global_mean_co2: Global mean atmospheric CO2 (ppm), computed via user-supplied time function.
+    dt : numpy.timedelta64, default 6h
+        Model timestep used to advance lead time coordinates.
+
+    References
```
NickGeneva, line 129: Thank you for adding references :+1:

earth2studio/models/px/ace2.py
```diff
@@ -0,0 +1,641 @@
...
+    per wall-clock day. ACE2-ERA5 was trained on the ERA5 dataset and requires
+    forcing data during rollout (see `forcing_data_source` parameter). This
+    wrapper makes use of the ``fme`` package to run model forward passes.
+
+    Parameters
+    ----------
+    checkpoint_path : str
+        Path to a serialized ACE2-ERA5 stepper checkpoint (e.g., tar file).
+    forcing_data_source : DataSource
+        Data source providing forcing data during rollout. Must provide all
```
NickGeneva: Perhaps the doc string can mention the use of the AI2 data source as a good go to? What do you think? Will more people just use that?

earth2studio/models/px/ace2.py
```diff
@@ -0,0 +1,641 @@
...
+
+    References
+    ----------
+    - ACE2-ERA5 paper: https://arxiv.org/abs/2411.11268v1
+    - ACE2 code: https://github.com/ai2cm/ace
+    """
+
+    def __init__(
+        self,
+        checkpoint_path: str,
```
NickGeneva: Typically this is the torch module object itself not the checkpoint location

So someone could manually load / override the loading and just pass in the object.

`stepper = load_stepper(checkpoint_path)`  gets moved into the load package function

earth2studio/models/px/ace2.py
```diff
@@ -0,0 +1,641 @@
...
+            _npdatetime64_to_cftime(times_state), dims=["sample", "time"]
+        )
+        hc_dims = ["lat", "lon"]
+        forcing_data = BatchData.new_on_device(
+            data=forcing_data,
+            time=time_da_forcing,
+            horizontal_dims=hc_dims,
+            labels=[set()],
+        )
+        state_data = BatchData.new_on_device(
```
NickGeneva, line 418: General question: for `BatchData.new_on_device` 

How does this know what device it needs to be on? Are you tracking what device the model is on?

earth2studio/data/ace.py
```diff
@@ -0,0 +1,219 @@
...
+
+    Parameters
+    ----------
+    base_path : str
+        HuggingFace repo ID or local directory path. Defaults to `allenai/ACE2-ERA5`.
+    mode : str
+        Either "forcing" or "initial_conditions". Controls which data tree and filenames are used.
+        Defaults to "forcing".
+    forcing_subdir : str, optional
+        Subdirectory for forcing data. Defaults to `forcing_data` containing files `forcing_YYYY.nc`.
```
NickGeneva: Nit: Please use the ", by default xyz" pattern here (end of doc lines of that parameter).

Use the autodoc extension in your cursor to generate the structure.
https://nvidia.github.io/earth2studio/userguide/developer/documentation.html#api-documentation

earth2studio/models/px/ace2.py
```diff
@@ -0,0 +1,641 @@
...
+        dt: np.timedelta64 = np.timedelta64(6, "h"),
+    ) -> PrognosticModel:
+        """Load ACE2-ERA5 prognostic model from a package.
+
+        Parameters
+        ----------
+        package : Package
+            Package to load the model checkpoint from.
+        forcing_data_source : DataSource, optional
+            External forcing data source, by default ACE2ERA5Data(mode="forcing").
```
NickGeneva: Nit Update to:

```
External forcing data source. All forcing variables described in the ACE2-ERA5 paper 
must be provided, by default ACE2ERA5Data(mode="forcing").
```

(default at end)

earth2studio/data/ace.py
pzharrington: Sort of goes back to the custom data use-case I mentioned in the other comment. I am fine with adopting this suggestion though to keep the constructor clean

earth2studio/models/px/ace2.py
pzharrington: Good catch, their setup is a bit wonky so this part should be clearer. Their `new_on_device` really just checks that all the tensors are on the current cuda device before returning the constructed `BatchData`. So a better pattern would be to check the device and then use `new_on_cpu` or `new_on_device` accordingly, will update

pzharrington: Per discussion on Slack, opted to keep this to `new_on_device` and have a warning in API docs on intended usage due to how ACE2 code interacts with `torch.cuda.current_device()`

earth2studio/data/ace.py
pzharrington: After more thinking I decided it's prob not worth the extra steps to support local paths/custom stuff. People in that scenario would likely have to write their own data source anyways in that scenario, and the CO2 override capability already allows for some basic level of customization

NickGeneva (approved):
LGTM, pushed some edits myself for little clean up and improvements to testing

I intentionally did not add the data source to the API docs... since seems theres more models in the pipeline.

I am a bit concerned with people getting confused that the ACE ERA data source is ERA5 in the standard form...

NickGeneva:
Here is the manual test code I used:

```python
import numpy as np

from datetime import datetime
from earth2studio.data import ACE2ERA5Data, fetch_data
from earth2studio.models.px import ACE2ERA5

prognostic = ACE2ERA5.load_model(ACE2ERA5.load_default_package())
ds = ACE2ERA5Data()
x, coords = fetch_data(
    ds, 
    np.array([datetime(2020,1,1)], dtype=np.datetime64),
    prognostic.input_coords()['variable'], 
    device="cuda"
)

coords_out = None
outputs = []
model = prognostic.create_iterator(x, coords)
for i, (x, coords) in enumerate(model):
    outputs.append(x)
    if i == 1:
        coords_out = coords
        break

idx = np.where(coords_out["variable"] == "t2m")[0][0]
fig, axes = plt.subplots(1, 5, figsize=(18, 4), constrained_layout=True)
axes = axes.flatten()
for i in range(len(outputs)):
    axes[i].imshow(outputs[i][0,0,idx].cpu().numpy(), origin='lower')

plt.savefig("ace.jpg")
```

Yielding results (can see diurnal cycle, looks reasonable"

image

NickGeneva:
From coverage runs:

```
earth2studio/data/ace2.py                   92      7     32      5    90%   90, 127-128, 161->171, 182, 198, 234, 249
earth2studio/models/px/ace2.py                      195     10     40      8    92%   178->186, 205-211, 270, 362, 467, 482-484, 542->541, 611->610
```

PR #557, opened 2025-11-18
Add TimeWindow temporal wrapper and CorrDiffCMIP6 diagnostic model

gertln:
# Earth2Studio Pull Request

## Description

1. TimeWindow data wrapper - A generic temporal data wrapper that fetches data at multiple time offsets
2. CorrDiffCMIP6 model wrapper - A CanESM5 CMIP6-specific variant of CorrDiff
3. diagnostic_from_data() - New workflow for applying diagnostic models directly to data without requiring a prognostic model

## Checklist


earth2studio/models/dx/corrdiff.py
```diff
@@ -721,6 +1230,400 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+        Returns
+        -------
+        torch.Tensor
+            Denormalized output tensor x * scale + center
+        """
+        print(1)
+        print(f"postprocess_output input shape: {x.shape}")
+        # crop to original extent
+        x = x[:, :, 23:-24, 48:-48]
+        print(f"postprocess_output after crop shape: {x.shape}")
```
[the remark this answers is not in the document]
gertln: fixed in https://github.com/NVIDIA/earth2studio/commit/78b69b8c83beb4ce094436d9cfe20ecbf3b969c7

earth2studio/data/__init__.py
```diff
@@ -36,6 +36,7 @@
 )
 from .rand import Random, Random_FX
 from .rx import CosineSolarZenith, LandSeaMask, SurfaceGeoPotential
+from .time_window import TimeWindow
```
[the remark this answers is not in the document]
gertln: fixed in https://github.com/NVIDIA/earth2studio/commit/78b69b8c83beb4ce094436d9cfe20ecbf3b969c7

gertln:
@greptile-apps

earth2studio/run.py
```diff
@@ -267,6 +267,113 @@ def diagnostic(
...
+        device=device,
+        interp_to=interp_to,
+        interp_method=interp_method,
+    )
+    logger.success(f"Fetched data from {data.__class__.__name__}")
+
+    # Remove lead_time dimension (diagnostic models usually operate on single lead)
+    if "lead_time" in coords:
+        x = x[:, 0]
+        del coords["lead_time"]
```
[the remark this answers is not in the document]
gertln: Diagnostic models that require temporal sequences (like CorrDiffCMIP6) can use the TimeWindow wrapper, which handles temporal offsets by:

1. Fetching data at multiple time offsets (t-1, t, t+1)
2. Concatenating along the variable dimension with suffixes (var_t-1, var_t, var_t+1)
3. Returning a single timestep with expanded variables
So when data reaches the diagnostic model it is a single timestep. The temporal information is encoded in the variable names, not in a lead_time dimension.

gertln:
@greptile-apps rescan

earth2studio/run.py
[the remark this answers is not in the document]
CharlelieLrt: Adding code to `run.py` is not allowed, this is an anti-pattern of the E2S package that should be avoided at all cost. If you do not need the function `diagnostic_from_data` I would remove it from there.
@NickGeneva please correct me if what Is wrong

earth2studio/data/time_window.py
```diff
@@ -0,0 +1,489 @@
...
+
+class TimeWindow:
+    """Wrapper for datasources that fetches data at multiple time offsets.
+
+    This wrapper takes an existing datasource and fetches data at multiple temporal
+    offsets relative to the requested time(s). The variable dimension is expanded
+    to include all offset versions with appropriate suffixes.
+
+    **Datasource contract**
+    -----------------------
```
CharlelieLrt, line 36: This is not a recognized section in the numpy-style docstring and will not render well in the docs

earth2studio/data/time_window.py
```diff
@@ -0,0 +1,489 @@
...
+      - ``time`` is a single ``datetime``, a list of ``datetime`` objects,
+        or a ``TimeArray``.
+      - ``variable`` is a string, list of strings, or ``VariableArray``.
+
+    - Returns an :class:`xarray.DataArray` with at least the following:
+
+      - A leading ``time`` dimension matching the requested times
+      - A ``variable`` dimension/coordinate whose entries match the *base*
+        variable names requested (e.g. ``\"tas\"``, ``\"u10m\"`` – **without**
+        any time suffixes such as ``\"_t-1\"``).
```
CharlelieLrt: I think it's fine not to remind all of this here but instead link the protocol: :class:`~earth2studio.data.DataSource`
Only keep here the contract requirements that are specific to the `TimeWindow` data source.

earth2studio/data/time_window.py
```diff
@@ -0,0 +1,489 @@
...
+            Expanded variable names with suffixes in deterministic order.
+        """
+        if self.group_by == "variable":
+            # Group by variable first: [v1_s1, v1_s2, ..., v2_s1, v2_s2, ...]
+            return [f"{v}{s}" for v in base_vars for s in self.suffixes]
+        # group_by == "offset"
+        # Group by offset first: [v1_s1, v2_s1, ..., v1_s2, v2_s2, ...]
+        return [f"{v}{s}" for s in self.suffixes for v in base_vars]
+
+    def _reorder_if_needed(
```
CharlelieLrt, line 176: The private method `_reorder_if_needed` is not called anywhere. If not useful, can it be removed?

earth2studio/data/time_window.py
```diff
@@ -0,0 +1,489 @@
...
+
+        result = xr.concat(offset_data_arrays, dim="variable")
+
+        # Use collected_names directly for ordering (they're already correct from _process_offset_data)
+        ordered_vars = [var for var in output_order if var in collected_names]
+        result = result.sel(variable=ordered_vars)
+
+        return result
+
+    async def fetch(
```
CharlelieLrt, line 349: I just realized your `CMIP6MultiRealm` data source is not defining an `async fetch` method. Is it on purpose? Could you explain why?

earth2studio/data/time_window.py
```diff
@@ -0,0 +1,489 @@
...
+    >>> # Create wrapper that fetches t-6h, t, and t+6h
+    >>> gfs = GFS()
+    >>> wrapped_gfs = TimeWindow(
+    ...     datasource=gfs,
+    ...     offsets=[timedelta(hours=-6), timedelta(hours=0), timedelta(hours=6)],
+    ...     suffixes=['_tm1', '_t', '_tp1']
+    ... )
+    >>>
+    >>> # Request data for a single time and multiple variables
+    >>> data = wrapped_gfs(datetime(2024, 1, 1), ['t2m', 'u10m'])
```
CharlelieLrt, line 119: Ok I understand this example. But how is the output ordered in a more complex case, if you request multiple data with `data = wrapped_gfs([datetime1, datetime2])`? Consider adding a second example if that would help understanding the behavior in this case

earth2studio/data/time_window.py
```diff
@@ -0,0 +1,489 @@
...
+from datetime import datetime, timedelta
+
+import xarray as xr
+
+from earth2studio.data.base import DataSource
+from earth2studio.data.utils import prep_data_inputs
+from earth2studio.utils.type import TimeArray, VariableArray
+
+
+class TimeWindow:
```
CharlelieLrt, line 28: Maybe a naive question, but why is this `TimeWindow` data source needed?
Intuitively, I would just have used the `CMIP6MultiReal` data source, and called it multiple times with the right datetimes (including the offsets) in the model wrapper. Would this approach work as well?

earth2studio/models/dx/corrdiff.py
```diff
@@ -52,13 +57,20 @@
...
     OptionalDependencyFailure("corrdiff")
     PhysicsNemoModule = None
     StackedRandomGenerator = None
     deterministic_sampler = None
 
 
+# Type alias for datasources that can be wrapped by TimeWindow.
+# They are expected to be callable like:
+#   datasource(time: datetime, variables: Sequence[str]) -> xarray.DataArray
+DataSource = Callable[[datetime, Sequence[str]], xr.DataArray]
```
CharlelieLrt, line 71: No objection to this, but any reason why you define a type alias instead of the protocol defined in `data/base.py`?

earth2studio/models/dx/corrdiff.py
```diff
@@ -118,6 +130,30 @@ class CorrDiff(torch.nn.Module, AutoModelMixin):
         Whether to use high-res mean conditioning, by default True
     seed : Optional[int], optional
         Random seed for reproducibility, by default None
+    time_window : dict | None, optional
```
CharlelieLrt, line 133: I strongly discourage any modification to the base CorrDiff wrapper, unless it is for adding a general purpose functionality, general purpose meaning a functionality potentially useful for many CorrDiff model wrapper subclasses. In this case, many of the modifications introduced to the `CorrDiff` wrapper are not really general purpose (e.g. the time window, the noise levels sigma, etc...).
Would it be possible to move these things to the `CorrDiffCMIP6` subclass instead? If you need to change/add things to the wrapper, it's preferable to do it by overriding in the subclass rather than modifying the base class.

That being said, maybe some of these functionalities you added to the base `CorrDiff` model wrapper are indeed general purpose. If you think so, we should have a discussion about adding them to the base wrapper. In any case, even if we decide it is indeed a good thing to add them to the base `CorrDiff` wrapper, that should be the subject of another PR. The current PR should only focus on the `CorrDiffCMIP6` wrapper.

earth2studio/models/dx/corrdiff.py
```diff
@@ -118,6 +130,30 @@ class CorrDiff(torch.nn.Module, AutoModelMixin):
...
+        For example, 0.05 allows output grid to extend 5% beyond input grid range.
+        Useful for Gaussian grids that don't include poles. Default is 0.0 (no extrapolation).
+    sigma_min : float | None, optional
+        Minimum noise level for diffusion process. If None, uses sampler-specific defaults:
+        deterministic sampler uses None (calculated internally), stochastic sampler uses 0.002.
+        By default None.
+    sigma_max : float | None, optional
+        Maximum noise level for diffusion process. If None, uses sampler-specific defaults:
+        deterministic sampler uses None (calculated internally), stochastic sampler uses 800.
+        By default None.
```
CharlelieLrt, line 156: I strongly suggest not exposing these parameters in the model wrapper, because they will be removed in future version of the samplers. IMO it also adds unnecessary complexity to the model wrapper

earth2studio/models/dx/corrdiff.py
```diff
@@ -159,18 +200,44 @@ def __init__(
...
+            )
+        else:
+            self.sigma_min = sigma_min
+
+        if sigma_max is None:
+            self.sigma_max: float | None = (
+                800.0 if sampler_type == "stochastic" else None
+            )
+        else:
+            self.sigma_max = sigma_max
```
CharlelieLrt, line 230: I would remove all of this. All of these low-level things _should_ be handled by the sampler shipped in physicsnemo

CharlelieLrt (changes requested):
A few comments about the PR:

- The `TimeWindow` data source is in relatively good shape. I left a few comments there, but those are mostly minor details.

- I'm 99% sure the `run.py` should never be modified. I'll need to confirm with @NickGeneva

- The PR introduces many changes to the existing `CorrDiff` base model wrapper, including some that are BC-breaking. This should be avoided at all cost. This PR should focus on the `CorrDiffCMIP6` and should circumvent code modifications to this new class. Changes to the base `CorrDiff` wrapper should be minimal. If there are more complex changes to upstream to the `CorrDiff` base model wrapper (e.g. new general purpose functionalities), we should first discuss about this and this should be the subject of a subsequent PR (so not high-priority for now).

- The wrapper `CorrDiffTaiwan` uses the most up-to-date APIs from physicsnemo (e.g. for the model, sampler, ...). It would be good to follow its logic in the new `CorrDiffCMIP6` wrapper.

- Please add to the PR (as a comment in the GitHub discussion), a short snippet and figures to illustrate outputs produced by the wrapper (see an example [there](https://github.com/NVIDIA/earth2studio/pull/533#issuecomment-3503539869) for another wrapper). This is used for validation if the model if revisited later on.

NickGeneva:
@CharlelieLrt you're correct, please do not add to the run.py. depending on what the workflow is, maybe this should be added as an example. Regardless, I think it would be best to push that to another pr for discussion and just focus on getting the wrapper added.

+1 it would be nice to see a code snippet in the PR that runs the model and also plots results. Please see other model PRs from me or Oliver for examples. Good for reference in the future 

Thanks guys!

earth2studio/data/time_window.py
gertln: The underlying intake-esgf library is synchronous and not thread-safe, so async wouldn't provide real parallelism here or would add a lot of additional complexity. We can revisit if intake-esgf adds native async support.

earth2studio/data/time_window.py
gertln: https://github.com/NVIDIA/earth2studio/pull/567

earth2studio/models/dx/corrdiff.py
gertln: Good point - fixed. Now importing DataSource from earth2studio.data

earth2studio/models/dx/corrdiff.py
gertln: https://github.com/NVIDIA/earth2studio/pull/566

earth2studio/models/dx/corrdiff.py
gertln: Why are these parameters removed in future versions of the samplers?

earth2studio/data/time_window.py
gertln: I removed it. Thanks.

PR #566, opened 2025-12-04
Improve CorrDiff base class

gertln:
# Earth2Studio Pull Request

## Description
This PR improves the CorrDiff base class with stricter validation, explicit configuration, and better extensibility for subclasses. These are general-purpose improvements that benefit all CorrDiff model wrappers.

#### 1. Explicit invariant configuration
Previously, `load_model` would silently load all variables from `invariants.nc` if the file existed, or silently fall back to no invariants if the file was missing. This implicit behavior made it hard to debug configuration issues.

**Now:**
- Invariants must be explicitly listed in `metadata.json` via `invariant_variables`
- If metadata specifies invariants but the file is missing → raises `FileNotFoundError`
- If file exists but requested variables are missing → raises `ValueError`
- If no invariants are specified in metadata, the model runs without them (explicit opt-out)

#### 2. Grid format validation
Added `_validate_grid_format()` to catch common configuration errors:
- Rejects rectilinear grids stored as 2D (should use efficient 1D storage)
- Validates 1D grids are monotonically increasing
- Provides clear error messages with remediation hints

#### 3. Configurable grid validation tolerances
New parameters in `__init__` and `load_model`:
- `grid_spacing_tolerance`: Relative tolerance for checking regular grid spacing (default: `1e-5`)
- `grid_bounds_margin`: Fraction of input grid range to allow for extrapolation (default: `0.0` = strict)

This allows models with slightly irregular grids or edge extrapolation to work without code changes.

#### 4. Inference context manager hook
Added `_inference_context()` method that subclasses can override to customize inference behavior (e.g., enable autocast, set memory format, disable gradients). The base implementation returns a no-op `nullcontext`.

#### 5. Backwards-compatible preprocessing
`preprocess_input()` now accepts both:
- 3D input `[C, H_in, W_in]` → returns `[1, C, H_out, W_out]`
- 4D input `[B, C, H_in, W_in]` → returns `[B, C, H_out, W_out]`

This maintains backwards compatibility while supporting more flexible single-sample use cases as well.

#### 6. Time-aware preprocessing for temporal features

`preprocess_input()` now extracts and stores the time dimension from input data, making it available for time-dependent feature computation (e.g., solar zenith angle). Previously, only spatial interpolation was performed; now subclasses can access `self._current_time` during inference to compute features that depend on the timestamp.

#### 7. Inference optimizations (following CorrDiffTaiwan patterns)

Added performance optimizations to `load_model` that match CorrDiffTaiwan:

- `profile_mode=False` on loaded models (disables profiling overhead)
- Models converted to `channels_last` memory format for better GPU/Tensor Core utilization
- `torch._dynamo` cache configuration for compilation readiness

These are safe optimizations that don't affect model outputs but can provide inference speedup on GPU.

#### 8. Other improvements
- Added `_load_json_from_package()` helper for consistent JSON loading
- Added duplicate variable detection in `load_model` with clear error message
- Added `_is_rectilinear()` helper to distinguish truly curvilinear grids from rectilinear grids stored as 2D

#### 9. Configurable sigma parameters for diffusion sampling
Added sigma_min and sigma_max parameters to both __init__ and load_model:
- Priority order: load_model argument > metadata.json > sampler defaults
- Tuning the noise schedule boundaries is often required to get optimal diffusion results for different domains/resolutions
- Exposing these parameters allows users and subclasses to adjust the diffusion sampling behavior without modifying code
- If not specified anywhere, the sampler's built-in defaults are used (backward-compatible)

### Test updates
- Fixtures now use truly curvilinear grids (with sinusoidal perturbation) instead of rectilinear meshgrids
- Added tests for both 3D and 4D preprocessing paths
- Updated `test_corrdiff_load_model_basic` to properly simulate missing invariants
- Adjusted exception tests to handle 1D vs 2D grid differences

## Checklist


earth2studio/models/dx/corrdiff.py
```diff
@@ -159,18 +171,26 @@ def __init__(
...
         self.residual_model = residual_model
         self.regression_model = regression_model
 
         # Store variable names
-        self.input_variables = input_variables
-        self.output_variables = output_variables
+        # Note: After preprocessing, actual channels = input_variables + invariant_variables
+        self.input_variables = list(input_variables)  # Weather variable channels
+        self.output_variables = list(output_variables)  # Output channels
+        self.invariant_variables = list(invariants.keys()) if invariants else []
```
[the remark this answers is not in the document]
gertln: I removed the redundant assignment

earth2studio/models/dx/corrdiff.py
```diff
@@ -118,6 +121,13 @@ class CorrDiff(torch.nn.Module, AutoModelMixin):
         Whether to use high-res mean conditioning, by default True
     seed : Optional[int], optional
         Random seed for reproducibility, by default None
+    grid_spacing_tolerance : float, optional
+        Relative tolerance for checking regular grid spacing. Allows for slight variations
+        in grid spacing (e.g., for Gaussian grids). Default is 1e-5 (0.001%).
```
CharlelieLrt: Can you specify in the docstrings what happens if the grid has variations in spacing that exceed `grid_spacing_tolerance`? Is it supposed to raise an error?

earth2studio/models/dx/corrdiff.py
```diff
@@ -159,18 +171,26 @@ def __init__(
         self.number_of_samples = number_of_samples
         self.number_of_steps = number_of_steps
         self.solver = solver
+        self.sampler_type = sampler_type
```
CharlelieLrt, line 185: I have been thinking about the `sigma_min` and `sigma_max` for the sampler. I think you can expose them here as parameters of the model wrapper (that are then passed. Since some tuning of these parameters is often required to get the best results, then it makes sense to expose them to all subclasses of the CorrDiff wrapper

earth2studio/models/dx/corrdiff.py
```diff
@@ -376,15 +559,58 @@ def load_default_package(cls) -> Package:
...
+            )
+
     @classmethod
     @check_optional_dependencies()
-    def load_model(cls, package: Package) -> DiagnosticModel:
+    def load_model(
+        cls,
+        package: Package,
+        grid_spacing_tolerance: float = 1e-5,
+        grid_bounds_margin: float = 0.0,
```
CharlelieLrt: Two comments about the arguments of `load_model`:
1. New arguments to `load_model` should be added only when absolutely necessary, so I think it would be better to find a way such that `grid_spacing_tolerance` and `grid_bounds_margin` are not added here. Also it seems a bit counterintuitive to have all other arguments infered from the package, while these two are the only ones that need to be specified to `load_model`. So, easiest solution would be to make them part of the package as well.
2. The `CorrDiffTaiwan` wrapper uses a `device` argument. Is there any reason why you didn't add it there? My understanding is that it is necessary to do the `.to(device)` **before** then to `.to(memory_format=torch.channels_last)` (because the latter is device-dependent), which is the reason why the `CorrDiffTaiwan` uses this `device` argument

earth2studio/models/dx/corrdiff.py
```diff
@@ -409,11 +635,32 @@ def load_model(cls, package: Package) -> DiagnosticModel:
...
-            metadata = json.load(f)
-        input_variables = metadata["input_variables"]
+        # Apply inference optimizations (following CorrDiffTaiwan patterns)
+        # Disable profiling mode for both models
+        residual.profile_mode = False
+        regression.profile_mode = False
+
+        # Convert to channels_last memory format for better GPU performance
+        residual = residual.to(memory_format=torch.channels_last)
+        regression = regression.to(memory_format=torch.channels_last)
```
CharlelieLrt, line 670: The `.to(device)` would need to be done before that

earth2studio/models/dx/corrdiff.py
```diff
@@ -702,8 +1060,16 @@ def __call__(
             device=x.device,
             dtype=torch.float32,
         )
+
+        # Extract time information if present in coords
+        time_array = coords.get("time", None)
+        if time_array is not None:
+            time_list = timearray_to_datetime(time_array)
+        else:
+            time_list = [None] * out.shape[0]
```
CharlelieLrt: I am not fond of this mechanism to pass the time as an item in `coords`, even though there might not be any better solution...
In any case case, the behavior should at least be documented in the docstring of `__call__`. That is, if there exist an item `time` in `coords` then this one will be used and passed to the `preprocess_input` method (if I didn't miss anything, it's the only place where it is used). Also, what is this `time` here? It's not a lead time, right? We are also risking confusion with the lead-time aware CorrDiff that takes as input the lead-time labels.

There is also the type and length requirement for `time`, right? We should raised error of `time` is in `coords` but does not have the right type/length.

I think overall this feature needs to be very clearly documented to clearly indicate how subclasses should use it.

CharlelieLrt: CI test failed because of type checking on this line `time_list = [None] * out.shape[0]`
`time_list` has to be a `list[datetime]`, so assigning a list full of `None` is going to fail

earth2studio/models/dx/corrdiff.py
gertln: Updated - docstring now specifies that ValueError is raised if tolerance is exceeded

earth2studio/models/dx/corrdiff.py
gertln: Done - `sigma_min `and `sigma_max `are now exposed as parameters on both `__init__ `and `load_model`. Priority order: load_model argument > metadata.json > sampler defaults.

earth2studio/models/dx/corrdiff.py
gertln: Both addressed:
- `grid_spacing_tolerance` / `grid_bounds_margin` now loaded from `metadata.json` (not `load_model` args)
- `device` argument added; `.to(device)` called before `.to(memory_format=torch.channels_last)`

earth2studio/models/dx/corrdiff.py
gertln: Done

earth2studio/models/dx/corrdiff.py
gertln: Addressed:
- **Documented in `__call__` docstring**: explains `coords["time"]` is optional validity time (not lead time), passed to `preprocess_input` as `valid_time`
- **Type/length validation**: raises `TypeError` for wrong dtype, `ValueError` if length != batch size
- **Renamed internally**: `time` → `valid_time` to avoid confusion with init_time/lead_time
- **Design note in code**: acknowledges this is a workaround since `CoordSystem` wasn't designed for per-sample metadata; points to `batch_func` / `handshake_dim` as the places to fix if this pattern becomes common

gertln: Fixed - added explicit type annotation `valid_time_list: list[datetime | None]` so the type checker accepts both branches.

gertln:
@greptile review

gertln:
> ### Additional Comments (1)
> 1. `earth2studio/models/dx/corrdiff.py`, line 191-192 ([link](/nvidia/earth2studio/blob/b1e5325a023b0d371f51c48abb3df72bb8e4db79/earth2studio/models/dx/corrdiff.py#L191-L192))
>    **style:** `self.invariant_center` and `self.invariant_scale` are stored but never accessed elsewhere in the code. They're concatenated into `in_center`/`in_scale` in `_register_buffers` (lines 477-478), which are then registered as buffers. Consider removing these unused instance attributes unless they're needed for inspection/debugging.
>    Note: If this suggestion doesn't match your team's coding style, reply to this and let me know. I'll remember it for next time!
> 
> 2 files reviewed, 1 comment
> 
> [Edit Code Review Agent Settings](https://app.greptile.com/review/github) | [Greptile](https://greptile.com?utm_source=greptile_expert&utm_medium=github&utm_campaign=code_reviews&utm_content=earth2studio_566)

Kept intentionally - subclasses may need access to invariant normalization stats separately.

CharlelieLrt (approved):
LGTM

PR #573, opened 2025-12-09
interpolate px models that take 2 time steps as input

mariusaurus:
# Earth2Studio Pull Request

## Description
The interpolation model determines the lead time in the output coords by adding lead times of input and output coords of the px model. If the px model takes two time steps as input, the interpolation model output coords are also of length 2, even though only a single time step is returned.
Here a fix is provided that takes the last element of the input element instead of the full list.

Closes: https://github.com/NVIDIA/earth2studio/issues/574

## Checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -98,5 +98,5 @@ coverage:
 include:
 
   - project: 'modulus/modulus-ci'
-    ref: ngeneva/e2s-rework
+    ref: main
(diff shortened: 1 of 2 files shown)


dallasfoster:
I think that this MR shows that we do not have good enough test coverage of the interpolation model as we should. Can I task you with adding the following changes to the test suite:

Currently test_interpmodafno.py uses Persistence for the px model and unfortunately we do not have the history implemented for Persistence yet so I created this issue: #574. 

In any case, I think we need some additional unit test before merging.

PR #597, opened 2025-12-19
CorrDiffCMIP6 integration

gertln:
# Earth2Studio Pull Request

## Summary
CorrDiffCMIP6 integration for CMIP6 downscaling: model wrapper + exports + numbered example + tests

## Changes
- Add `CorrDiffCMIP6` CMIP6-specific wrapper implementation in `earth2studio/models/dx/corrdiff.py`
  - CMIP6 preprocessing (time features + channel handling) and postprocessing
  - Optional multi-sample CPU streaming and optional tqdm progress for ensembles
- Export `CorrDiffCMIP6` via `earth2studio/models/dx/__init__.py`
- Add numbered example `examples/20_corrdiff_cmip6.py`
- Extend tests in `test/models/dx/test_corrdiff.py` with focused CorrDiffCMIP6 coverage
  - preprocessing + required time handling
  - multi-sample streaming
  - postprocess crop/clamp/flip behavior

## Notes / Follow-ups
- `CorrDiffCMIP6.load_default_package()` uses a placeholder URI until the final package URI is configured.

## Checklist

### DIFF
--- a/earth2studio/models/dx/corrdiff.py
+++ b/earth2studio/models/dx/corrdiff.py
@@ -30,6 +30,12 @@ import torch.nn.functional as F
 import xarray as xr
 import zarr
 
+try:
+    from tqdm.auto import tqdm
+except ImportError:  # pragma: no cover
+    tqdm = None
+
+from earth2studio.data import TimeWindow
 from earth2studio.models.auto import AutoModelMixin, Package
 from earth2studio.models.batch import batch_coords, batch_func
 from earth2studio.models.dx.base import DiagnosticModel
@@ -992,15 +998,17 @@ class CorrDiff(torch.nn.Module, AutoModelMixin):
         return x * self.out_scale + self.out_center
 
     def _inference_context(self) -> nullcontext:
-        """Return context manager for model inference.
+        """Return context manager to wrap model inference operations.
 
-        Base class returns nullcontext (no-op). Subclasses can override
-        to add autocast, profiling, or other context-dependent behavior.
+        This is a CorrDiff-internal hook (not a PyTorch `torch.nn.Module` method).
+        Subclasses can override it to change inference behavior (e.g., autocast,
+        profiling) without duplicating the full `_forward` implementation.
 
         Returns
         -------
         nullcontext
-            Context manager to wrap inference operations
+            Base class returns nullcontext (no-op). Subclasses can override to
+            return autocast, profiling contexts, or other context managers.
         """
         return nullcontext()
 
@@ -1226,6 +1234,15 @@ class CorrDiffCMIP6(CorrDiff):
     - Overrides default interpolation
     - Includes a time dimension in the coordinate system so timestamps are preserved through
       batching/compression and can be used for time-dependent features (e.g., solar zenith angle)
+
+    Note
+    ----
+    Unlike CorrDiffTaiwan which has fixed input/output variables, CorrDiffCMIP6
+    loads variables names from the model package. Input variables are
+    expanded based on the ``time_window`` configuration in metadata.json
+    (e.g., "t2m" → "t2m_t-6", "t2m_t-3", "t2m_t+0"). After loading, inspect
+    ``model.input_variables`` and ``model.output_variables`` for the actual
+    variable lists.
     """
 
     # Variables that must be non-negative (clipped to min=0 during postprocessing)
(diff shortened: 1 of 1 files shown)


gertln:
## Plots
The example `python examples/20_corrdiff_cmip6.py` writes two figures:

- `20_corrdiff_cmip6_quicklook.png`: 2×2 map comparison of **coarse CMIP6 input** vs **CorrDiffCMIP6 downscaled output** for
  - temperature (**CMIP6 `tas` vs CorrDiff `t2m`**)
  - 10 m wind speed (**ws10m = sqrt(u²+v²)**), computed from **CMIP6 `uas`/`vas`** and from downscaled **`u10m`/`v10m`**
- `20_corrdiff_cmip6_ws10m_ensemble_3x3.png`: 3×3 grid showing **7 stochastic members** of downscaled ws10m plus **ensemble mean** and **ensemble std**.

**Single-sample comparison**  
image


**Ensemble ws10m (members/mean/std)**  
image

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1214,866 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+        "q700",
+        "q850",
+        "q925",
+        "q1000",
+        "sst",
+        "d2m",
+    ]
+
+    @batch_func()
+    def __call__(
```
CharlelieLrt: It's a little strange to put the `__call__` before the `__init__`

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1214,866 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+            Input tensor [C, H, W]
+
+        Returns
+        -------
+        torch.Tensor
+            Input tensor unchanged [C, H, W]
+        """
+        return x
+
+    def input_coords(self) -> CoordSystem:
```
CharlelieLrt, line 1718: I noticed that you did not implement anything for `output_coords`, did you intend to inherit the one from the parent `CorrDiff` class?

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1214,866 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+    - Adds time-dependent features (solar zenith angle, hour of day) to inputs
+    - Uses relaxed grid validation tolerances suitable for Gaussian grids (1% spacing, 5% bounds)
+    - Overrides default interpolation
+    - Includes a time dimension in the coordinate system so timestamps are preserved through
+      batching/compression and can be used for time-dependent features (e.g., solar zenith angle)
+    """
+
+    # Variables that must be non-negative (clipped to min=0 during postprocessing)
+    # These represent physical quantities that cannot be negative (temperature, pressure, etc.)
+    _NONNEGATIVE_VARS = [
```
CharlelieLrt, line 1250: What about other variables? I am not seeing `VARIABLES` defined anywhere in the class like for the other wrappers?

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1214,866 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+        loop, passing the extracted timestamp as ``valid_time`` to ``preprocess_input``.
+        """
+        if "time" not in coords:
+            raise ValueError(
+                'CorrDiffCMIP6 requires a singleton "time" dimension in coords.'
+            )
+        if len(coords) != x.ndim:
+            raise ValueError(
+                'CorrDiffCMIP6 expects "time" as a tensor dimension (len(coords) must equal x.ndim).'
+            )
```
CharlelieLrt: I have been reading this multiple times, but this is still not really clear to me. The docstring would benefit from having a `Parameters` section the describe in more detail what `x` and `coords` should contain. Also I do not understand the "singleton time axis" requirement. Some explanations a bit more "high-level" would also help

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1214,866 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+            }
+        )
+        handshake_dim(coords_no_time, "lon", 3)
+        handshake_dim(coords_no_time, "lat", 2)
+        handshake_dim(coords_no_time, "variable", 1)
+        handshake_coords(coords_no_time, target_input_coords, "lon")
+        handshake_coords(coords_no_time, target_input_coords, "lat")
+        handshake_coords(coords_no_time, target_input_coords, "variable")
+
+        output_coords = OrderedDict(
```
CharlelieLrt: If I am not mistaken, this should be returned by `self.output_coords()`, right? (see the `CorrDiffTaiwan` wrapper for reference)

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1214,866 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+            out[i] = self._forward(x[i], valid_time)
+        return out, output_coords
+
+    def __init__(
+        self,
+        *args: Any,
+        time_feature_center: torch.Tensor | None = None,
+        time_feature_scale: torch.Tensor | None = None,
+        time_window: dict | None = None,
+        **kwargs: Any,
```
CharlelieLrt: Can we avoid the `**kwargs`, or or at least document what can be passed with a clear reference to the parent class?

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1214,866 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+
+        out_shape = tuple(len(v) for v in output_coords.values())
+        out = torch.empty(out_shape, device=x.device, dtype=torch.float32)
+        for i in range(out.shape[0]):
+            out[i] = self._forward(x[i], valid_time)
+        return out, output_coords
+
+    def __init__(
+        self,
+        *args: Any,
```
CharlelieLrt: To avoid this `*args`. For optional arguments I would say it's acceptable, but better to avoid for required positional arguments

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1214,866 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+        time_feature_center: torch.Tensor | None = None,
+        time_feature_scale: torch.Tensor | None = None,
+        time_window: dict | None = None,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(*args, **kwargs)
```
CharlelieLrt: The entire signature is so unclear at the end of the I am not even sure what can/must be passed to the constructor:
- All required args from the base `CorrDiff` wrapper
- Plus optionally `time_feature_center`, `time_feature_scale`, `time_window `
- Plus any kwarg accepted by the base `CorrDiff` wrapper
Is it correct?
All of this needs clear documentation from a user perspective + some cleanup

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1214,866 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+            {
+                "batch": np.empty(0),
+                "time": np.empty(0),
+                "variable": np.array(self.input_variables),
+                "lat": self.lat_input_numpy,
+                "lon": self.lon_input_numpy,
+            }
+        )
+
+    def _inference_context(self) -> torch.autocast:
```
CharlelieLrt, line 1779: What is this for? I just noticed that you introduced this `_inference_context` method in the base `CorrDiff` wrapper in a previous PR. This is overriding a semi-private method from `torch.nn.Module`, correct? What's the purpose of doing so?

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1214,866 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+
+        Raises
+        ------
+        ValueError
+            If valid_time is None (required for CorrDiffCMIP6)
+        """
+        if valid_time is None:
+            raise ValueError(
+                "CorrDiffCMIP6 requires valid_time for time-dependent features"
+            )
```
CharlelieLrt, line 1824: Why making it optional in the function definition then?

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1214,866 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+        x : torch.Tensor
+            Normalized output tensor to denormalize
+
+        Returns
+        -------
+        torch.Tensor
+            Denormalized output tensor x * scale + center
+        """
+        # 1) Crop padding added during preprocessing
+        x = x[:, :, 23:-24, 48:-48]
```
CharlelieLrt: Can you add a comment for the hard-coded indices?

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1214,866 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+        # Helper: generate one sample (shape [1, C, H, W] on model device)
+        def _generate_one(i: int) -> torch.Tensor:
+            seed0: int = (
+                int(self.seed)
+                if self.seed is not None
+                else int(np.random.randint(2**32))
+            )
+
+            image_res = None
+            if self.residual_model and self.inference_mode != "regression":
```
CharlelieLrt: I really don't like the option given by `self.inference_mode` to do either regression only or regression + diffusion. Unless there is a very strong and justified need to produce regression-only outputs (just for debugging purpose or simple curiosity is not sufficient), then it would be better if this option is removed all together (including from the parent class)

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1214,866 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+                    latents_shape=latents_shape,
+                )
+
+        # Decide where to store the stacked samples
+        out_device = (
+            torch.device("cpu") if self.stream_samples_to_cpu else image_lr.device
+        )
+
+        # Helper: generate one sample (shape [1, C, H, W] on model device)
+        def _generate_one(i: int) -> torch.Tensor:
```
CharlelieLrt: Why do we need to generate the samples one by one with a loop? Memory constraint? What's the max number of batched samples we can generate on an H100?

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1214,866 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+            if image_reg is None or image_res is None:
+                missing = []
+                if image_reg is None:
+                    missing.append("regression_model")
+                if image_res is None:
+                    missing.append("residual_model")
+                raise RuntimeError(
+                    "Missing outputs for inference_mode='both': expected outputs from "
+                    f"{', '.join(missing)}."
+                )
```
CharlelieLrt: I really don't like this, this is starting to look like the Physicsnemo CorrDiff example, that is awfully over-complicated because of this option of doing regression-only prediction

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1214,866 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+
+        it = range(1, self.number_of_samples)
+        if self.show_sample_progress and self.number_of_samples > 1:
+            try:
+                from tqdm.auto import tqdm
+            except ImportError as e:  # pragma: no cover
+                raise ImportError(
+                    "Progress bar requested (CorrDiffCMIP6.show_sample_progress=True) "
+                    "but tqdm is not installed. Install it with `pip install tqdm`."
+                ) from e
```
CharlelieLrt: @NickGeneva is this mechanism acceptable for optional dependencies? I see other wrappers usually do they import at the top of the module and set the dependency to `None` if the import fails... There is  also the `check_optional_dependencies` decorator that should check for those, correct?

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1214,866 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+        >>> # Basic usage (no time transformation)
+        >>> wrapped = model.create_time_window_wrapper(datasource)
+        >>>
+        >>> # With time normalization to noon
+        >>> def to_noon(dt):
+        ...     return dt.replace(hour=12, minute=0, second=0, microsecond=0)
+        >>> wrapped = model.create_time_window_wrapper(datasource, time_fn=to_noon)
+        """
+        # Lazy import to avoid circular dependency
+        from earth2studio.data import TimeWindow
```
CharlelieLrt: Move to top of module

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1214,866 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+        package : Package
+            Package containing model weights and configuration
+
+        Returns
+        -------
+        DiagnosticModel
+            Initialized CorrDiffCMIP6 model
+        """
+        # Load and validate metadata first (we need time_window for input expansion).
+        metadata = CorrDiff._load_json_from_package(package, "metadata.json")
```
CharlelieLrt: Why using `CorrDiff. _load_json_from_package` instead of just `cls. _load_json_from_package`?

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1214,866 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+        if time_window_raw is None:
+            raise ValueError(
+                "metadata.json is missing required 'time_window' configuration."
+            )
+        time_window = cls._validate_time_window_metadata(time_window_raw)
+        suffixes: list[str] = time_window["suffixes"]
+        group_by: str = time_window["group_by"]
+
+        # Load stats for time feature normalization (sza/hod).
+        stats = CorrDiff._load_json_from_package(package, "stats.json")
```
CharlelieLrt: Same here

CharlelieLrt (changes requested):
Overall, I think the model wrapper would benefit from a cleaner more streamlined implementation (e.g options that are maybe not 100% critical but that adds significant clutter to the code). It also needs more user-oriented/user-friendly documentation. If you provide this wrapper in its current state to some users I am not confident they would be able to figure out what it does and how to use it by themselves.

earth2studio/models/dx/corrdiff.py
gertln: Moved __init__ above __call__ in the latest commit

earth2studio/models/dx/corrdiff.py
gertln: added `output_coords`

earth2studio/models/dx/corrdiff.py
gertln: CorrDiffCMIP6 doesn’t have a fixed VARIABLES list in the code. It loads input_variables/output_variables from the model package metadata.

earth2studio/models/dx/corrdiff.py
gertln: Updated the docstrings with a complete parameters section and an explanation of the singleton time axis.

earth2studio/models/dx/corrdiff.py
gertln: Fixed - `__call__` now uses `self.output_coords(coords)` directly.

examples/20_corrdiff_cmip6.py
CharlelieLrt, line 1: We would prefer not to have any example for a model wrapper, unless the wrapper requires to be used in a highly unusual way that necessitates an example to figure out how to make it work.  So we would prefer to remove this file and copy paste the code used to generate the images in the GitHub conversation linked to the PR

earth2studio/models/dx/corrdiff.py
gertln: Simplified - CorrDiffCMIP6 now only allows `"regression"` or `"both"`, and `_forward` is linear without the nested branching.

earth2studio/models/dx/corrdiff.py
gertln: Updated to module‑level optional import (tqdm = None) with a simple runtime check

earth2studio/models/dx/corrdiff.py
gertln: Moved to top

earth2studio/models/dx/corrdiff.py
gertln: Switched to `cls._load_json_from_package(...)`

earth2studio/models/dx/corrdiff.py
gertln: Switched to cls._load_json_from_package(...)

earth2studio/models/dx/corrdiff.py
gertln: Done — removed `*args/**kwargs`, made all parameters explicit, and added full `__init__` docs.

earth2studio/models/dx/corrdiff.py
gertln: Removed `*args`

earth2studio/models/dx/corrdiff.py
gertln: It’s a CorrDiff‑specific hook (not a `torch.nn.Module` method) to wrap inference in a context manager (e.g., autocast/profiling) without duplicating `_forward`.

earth2studio/models/dx/corrdiff.py
gertln: Kept `valid_time: datetime | None` to match the base class signature, but `CorrDiffCMIP6 ` enforces it at runtime since time‑dependent features are required.

earth2studio/models/dx/corrdiff.py
gertln: Added comments and shared constants for the crop/pad indices.

earth2studio/models/dx/corrdiff.py
gertln: Kept regression‑only for prototyping and fast runs when interested in conditional mean only, but simplified CorrDiffCMIP6 to allow only `regression` or `both`.

earth2studio/models/dx/corrdiff.py
gertln: Kept per‑sample loop because `diffusion_step `enforces batch size 1.

earth2studio/models/dx/corrdiff.py
CharlelieLrt: IMO that's still way too complicated. I don't think there is any compelling use-case to do regression-only inference (prototyping is not one of them) and to have to deal with the added code complexity. My recommendation is to drop all together this option of regression-only vs. both and to just do the default which is both.

earth2studio/models/dx/corrdiff.py
CharlelieLrt: Right, But if it's just a `torch.autocast`, my recommendation is to delete this method from the wrapper (as well as from the base CorrDiff wrapper) and just replace  where it is called with `with torch.autocast()`

earth2studio/models/dx/corrdiff.py
CharlelieLrt: It is fine to override the signature and make the argument required (switch off ruff error with a # noqa comment)

earth2studio/models/dx/corrdiff.py
CharlelieLrt: Only `both` makes sense to me. All the others mode of inference should be deleted from this wrapper and the parent class. The benefit brought by the ability to prototype are too little in. comparison to the added code complexity. Also E2S is a production code library, so prototyping is not really in its scope (users should go to physicsnemo for that, or write their own code, which is not difficult to do)

earth2studio/models/dx/corrdiff.py
CharlelieLrt: I'm not totally sure I understand this correctly, but `diffusion_step ` does not enforce a batch size of 1. If you pass `rank_batches=[[0, 1, 2], [3, 4, 5, 6, 7, 8]], [9, 10]`, then `diffusion_step` should generate an ensemble of 11 predictions, and it should do so in a mini-batch style loop, with mini-batch sizes 3, 9, 2, respectively.

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1222,1044 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+        inference_mode: Literal["regression", "diffusion", "both"] = "both",
+        hr_mean_conditioning: bool = True,
+        seed: int | None = None,
+        grid_spacing_tolerance: float = 1e-5,
+        grid_bounds_margin: float = 0.0,
+        sigma_min: float | None = None,
+        sigma_max: float | None = None,
+        time_feature_center: torch.Tensor | None = None,
+        time_feature_scale: torch.Tensor | None = None,
+        time_window: dict | None = None,
```
CharlelieLrt, line 1333: IMO there are way too many arguments in this wrapper, which makes it unnecessarily over-complicated (and this applies also to the base parent wrapper). The abstraction is in fact so complicated, that the time a user would need to understand these arguments and how to use the wrapper, it would actually be faster for them to write their own inference code.

My suggestion is to drastically simplify the wrapper arguments (in the parent class as well), and only keep those that are absolutely essentials (e.g. `number_of_steps`, `sigma_min`, `signa_max` need to be tuned to generate good samples, but this is less sure for other parameters like `sampler_type`, `hr_mean_conditioning`)

Ideally we should thrive for an abstraction that could be understood by any lambda developer in less than 5 minutes.

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1222,1044 @@ def to(self, device: torch.device) -> "CorrDiff":
         return self
 
 
+class CorrDiffCMIP6(CorrDiff):
```
CharlelieLrt, line 1225: Can we reorder the methods in this class in a more logical way. Maybe first `__init__`, `load_model` (makes sense to have these two one after the other), then methods for calling (`_forward`, `__call__`), etc.,.. and in the end all the private methods used for validation and pre-processing

earth2studio/models/dx/corrdiff.py
```diff
@@ -1212,6 +1222,1044 @@ def to(self, device: torch.device) -> "CorrDiff":
...
+                f"solver must be either 'euler' or 'heun' but got {self.solver}"
+            )
+
+        # Preprocess input (CMIP6 requires valid_time)
+        image_lr = self.preprocess_input(x, valid_time)
+        image_lr = image_lr.to(torch.float32).to(memory_format=torch.channels_last)
+
+        # Regression model (mean)
+        image_reg = None
+        if self.regression_model:
```
CharlelieLrt, line 1912: All these checks on `self.regression_model` and `self.diffusion_model` throughout the wrapper implementation are useless. Both should always be provided in the package

CharlelieLrt (changes requested):

NickGeneva:
Closing with merge of #635 , thanks for the hard work here!

PR #614, opened 2026-01-09
IO Add Array Fix

NickGeneva:
# Earth2Studio Pull Request

## Description

Fixes instances when an array is written and it does not exist in the IO backend, but some coordinates do exist in the io backend.

Previously the coordinates in the IO would get overwritten by the array slice getting written, which we dont want... because slices get written not whole array.

E.g. writing a single time-step would overwrite the time coordinate for all arrays in the IO backend.

## Checklist

### DIFF
--- a/earth2studio/io/kv.py
+++ b/earth2studio/io/kv.py
@@ -110,7 +110,9 @@ class KVBackend:
 
         adjusted_coords, mapping = convert_multidim_to_singledim(coords)
 
-        self.coords = self.coords | adjusted_coords
+        for c, v in adjusted_coords.items():
+            if c not in self.coords:
+                self.coords[c] = v
 
         # Add any multidim coordinates that were expelled above
         for k in mapping:
--- a/earth2studio/io/xarray.py
+++ b/earth2studio/io/xarray.py
@@ -126,7 +126,9 @@ class XarrayBackend:
 
         adjusted_coords, mapping = convert_multidim_to_singledim(coords)
 
-        self.coords = self.coords | adjusted_coords
+        for c, v in adjusted_coords.items():
+            if c not in self.coords:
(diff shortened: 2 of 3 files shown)


NickGeneva:
Fixes the following edge case where some trys to write an array that does not exist:

```python
from pathlib import Path
import torch
import numpy as np
from earth2studio.io import NetCDF4Backend, ZarrBackend

io = NetCDF4Backend(file_name=Path("out.nc"), backend_kwargs={'mode': 'w'})
io.add_array({"lon": np.arange(10), "lat": np.arange(4)}, ["test"])

io.write(torch.zeros(2,4), {"lon": np.arange(2), "lat": np.arange(4)}, "test")
io.write(torch.ones(5,4), {"lon": np.arange(5), "lat": np.arange(4)}, "test2")

import xarray as xr

ds = xr.load_dataset("out.nc")

print(ds["test"])
print(ds["test2"])



io = ZarrBackend(file_name=Path("out.zarr"), backend_kwargs = {"overwrite": True})
io.add_array({"lon": np.arange(10), "lat": np.arange(4)}, ["test"])

io.write(torch.full((2,4), fill_value=0.5), {"lon": np.arange(2), "lat": np.arange(4)}, "test")
io.write(torch.ones(5,4), {"lon": np.arange(5), "lat": np.arange(4)}, "test2")

import xarray as xr

ds = xr.load_dataset("out.zarr")

print(ds["test"])
print(ds["test2"])
```

Previously this would just error:

```
ValueError: shape mismatch: objects cannot be broadcast to a single shape.  Mismatch is between arg 0 with shape (10, 4) and arg 1 with shape (5, 4).
```

but now properly adds in initialized arrays with partial values to the stores.

PR #626, opened 2026-01-15
Add StormScope GOES and MRMS models

pzharrington:
# Earth2Studio Pull Request

## Description
Adds StormScope GOES and MRMS models, with a simplified single-GPU notebook example. Introduces some new interpolation functionality for nearest-neighbor interp, helpful for getting multimodal data inputs onto the same regional grid.

## Checklist

## Dependencies
`natten` for the DiT architecture (MIT, https://github.com/SHI-Labs/NATTEN)

### DIFF
--- a/earth2studio/data/mrms.py
+++ b/earth2studio/data/mrms.py
@@ -117,7 +117,9 @@ class MRMS:
 
     async def _async_init(self) -> None:
         """Async initialization of S3 filesystem"""
-        self.fs = s3fs.S3FileSystem(anon=True, client_kwargs={}, asynchronous=True)
+        self.fs = s3fs.S3FileSystem(
+            anon=True, client_kwargs={}, asynchronous=True, skip_instance_cache=True
+        )
 
     def __call__(
         self,
--- a/earth2studio/models/px/stormscope.py
+++ b/earth2studio/models/px/stormscope.py
@@ -87,9 +87,11 @@ class StormScopeBase(torch.nn.Module, AutoModelMixin, PrognosticMixin):
     ----------
     model_spec : list[dict[str, Any]]
         Sequence of stage specifications. Each entry must contain:
+
         - 'model': physicsnemo.Module
         - 'sigma_min': float
         - 'sigma_max': float
+
         The sigma interval determines which expert is applied during denoising.
     means : torch.Tensor
         Per-variable mean used to normalize inputs. Expected shape [1, C, 1, 1].
--- a/earth2studio/data/gefs.py
+++ b/earth2studio/data/gefs.py
@@ -145,7 +145,9 @@ class GEFS_FX:
         ----
         Async fsspec expects initialization inside of the execution loop
         """
-        self.fs = s3fs.S3FileSystem(anon=True, client_kwargs={}, asynchronous=True)
+        self.fs = s3fs.S3FileSystem(
+            anon=True, client_kwargs={}, asynchronous=True, skip_instance_cache=True
+        )
 
     def __call__(
         self,
--- a/earth2studio/data/gfs.py
+++ b/earth2studio/data/gfs.py
@@ -155,7 +155,9 @@ class GFS:
         ----
         Async fsspec expects initialization inside of the execution loop
         """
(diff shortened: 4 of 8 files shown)


earth2studio/data/mrms.py
```diff
@@ -90,7 +90,7 @@ class MRMS:
 
     def __init__(
         self,
-        max_offset_minutes: float = 0,
+        max_offset_minutes: float = 10,
```
NickGeneva, line 95: nit: update the api doc string above

earth2studio/models/px/stormscope.py
```diff
@@ -0,0 +1,1622 @@
...
+            output data tensor and coordinate system dictionary.
+        """
+        yield from self._default_generator(x, coords)
+
+
+class StormScopeGOES(StormScopeBase):
+    """StormScope model forecasting GOES data on the HRRR grid.
+
+    This model supports multiple variants at different spatiotemporal resolutions:
+      - 6km resolution, 60 minute timestep
```
NickGeneva, line 969: I think a quirk of sphinx is you need blank lines separating a list, need to double check

NickGeneva:
Over all looks really good, I just need to cross verify the example runs reasonably.

pyproject.toml
```diff
@@ -269,6 +279,9 @@ docs = [
     "pydata-sphinx-theme==0.16.1"
 ]
 
+[tool.hatch.metadata]
```
NickGeneva: not familiar with this option, what was this added for?

pyproject.toml
```diff
@@ -79,6 +79,10 @@ perturbation = [
 statistics = [
     "nvidia-physicsnemo>=1.0.1,<2.0",
 ]
+utils = [
```
NickGeneva, line 81: Please add to [all] group if we keep this

earth2studio/utils/interp.py
```diff
@@ -196,3 +208,189 @@ def forward(self, values: Tensor) -> Tensor:
         f0 = torch.lerp(f00, f01, dj)
         f1 = torch.lerp(f10, f11, dj)
         return torch.lerp(f0, f1, i - i0)
+
+
+@check_optional_dependencies("utils")
+class NearestNeighborInterpolator(nn.Module):
```
NickGeneva, line 214: Will this be used anywhere outside StormScope? Maybe it makes more sense to just move this into the stormscope utils right now. I dont see any specific tests for this util?

NickGeneva: Its hard for me to envision this being used outside of the model atm... since its a module

pyproject.toml
pzharrington: Ah yeah, so it's technically not needed anymore here but I still think it could be useful in general? It allows you to in-line pin something in an optional deps group to a specific source/commit, without needing to go and separately define a source. e.g., something like 
```
[project.optional-dependencies]
weirdmodel = [
  "torch @ git+https://github.com/pytorch/pytorch.git@<COMMIT_SHA>",
]
```
and then other models could still just use regular torch in their deps groups. I was using it to version pin PNM to a specific commit for just the `stormscope` group in early testing

earth2studio/utils/interp.py
pzharrington: True I forgot to add tests for it. But I don't think it is necessarily useful only to StormScope (yes that is the only model using it as of now 😅), NN interpolation could be more broadly useful in that it's pretty flexible in handling cases including arbitrary/weird grids and/or irregular data. Where these `nn.Module`s should live is maybe the question since they seem sort of awkward here, but are not quite models themselves, but can still definitely serve an important purpose in a custom/complex workflow

pyproject.toml
NickGeneva: If its  not needed, then lets remove it, TOML has enough options in it as is.
If I remember correctly theres packaging implications for direct references (pypi doesnt like them)

Better to just lean on uv to deal with these issues with custom source like `nvidia-physicsenemo3`

pzharrington: Done

NickGeneva (approved):
LGTM! Awesome!

PR #628, opened 2026-01-16
Planetary Computer ECMWF Open Data IFS data source

swbg:
# Earth2Studio Pull Request

## Description
Adds a data source for the ECMWF Open Data IFS repository on Planetary Computer so that global weather models can be initialized from Planetary Computer.

## Checklist

### DIFF
--- a/earth2studio/data/planetary_computer.py
+++ b/earth2studio/data/planetary_computer.py
@@ -928,7 +928,7 @@ class PlanetaryComputerECMWFOpenDataIFS(_PlanetaryComputerData):
     ASSET_KEY = "data"
     SEARCH_KWARGS = {
         "query": {
-            "ecmwf:stream": {"eq": "oper"},
+            "ecmwf:stream": {"in": ["oper", "scda"]},
             "ecmwf:type": {"eq": "fc"},
             "ecmwf:step": {"eq": "0h"},
             "ecmwf:resolution": {"eq": "0.25"},


earth2studio/data/planetary_computer.py
```diff
@@ -495,13 +502,22 @@ def _locate_item(self, when: datetime) -> Any:
...
-                f"within ±{self._search_tolerance}"
-            ) from error
+            msg = f"No Planetary Computer item found for {when.isoformat()}"
+            if self._search_tolerance is not None:
+                msg += f" within ±{self._search_tolerance}"
+            raise FileNotFoundError(msg) from error
+
+        try:
+            _ = next(items)
+            logger.warning("Found more than one matching item, returning first match")
```
swbg, line 517: Just an idea, not sure.

NickGeneva:
Thanks @swbg , @loliverhennigh to have a look and test

earth2studio/data/planetary_computer.py
```diff
@@ -144,7 +148,7 @@ def __init__(
         lexicon: LexiconType,
         asset_key: str = "netcdf",
         search_kwargs: Mapping[str, Any] | None = None,
-        search_tolerance: timedelta = timedelta(hours=12),
+        search_tolerance: timedelta | None = timedelta(hours=12),
```
NickGeneva: I dont think supporting none is a requirement here. Adds unnecessary conditions that you then need to test. Better to just fix it to a timedelta datatype (e.g. you wouldnt have int | None here, youd just set the int to 0).

Maybe the tolerance should just be timedelta(hours=0) by default and people can change it otherwise? 

Does the following

```python
start = (when - self._search_tolerance).isoformat()
end = (when + self._search_tolerance).isoformat()
datetime_param = f"{start}/{end}"
````
        
not work if the delta is 0?

File: earth2studio/data/planetary_computer.py [hunk 3]
```diff
@@ -495,13 +502,22 @@ def _locate_item(self, when: datetime) -> Any:
...
         try:
-            return next(search.items())
+            item = next(items)
         except StopIteration as error:
-            raise FileNotFoundError(
-                f"No Planetary Computer item found for {when.isoformat()} "
-                f"within ±{self._search_tolerance}"
-            ) from error
+            msg = f"No Planetary Computer item found for {when.isoformat()}"
+            if self._search_tolerance is not None:
```
NickGeneva: If you make search_tol always a time-delta you can remove this conditon

earth2studio/data/planetary_computer.py
```diff
@@ -495,13 +502,22 @@ def _locate_item(self, when: datetime) -> Any:
...
+                msg += f" within ±{self._search_tolerance}"
+            raise FileNotFoundError(msg) from error
+
+        try:
+            _ = next(items)
```
NickGeneva: Will the actually perform a fetch when you iterate? Curious if this would download extra not needed data. Is this a generator? can you not len(items)? :thinking:

earth2studio/data/planetary_computer.py
```diff
@@ -871,3 +887,112 @@ def extract_variable_numpy(
             values = np.asarray(field.values).astype(np.float32)
             result = np.asarray(spec.modifier(values), dtype=np.float32)
             return result
+
+
+class PlanetaryComputerECMWFOpenDataIFS(_PlanetaryComputerData):
+    """IFS analysis data from the ECMWF Open Data repository.
```
NickGeneva, line 902: Million dollar question, can you run AIFS or AIFS ENS using this?

test/lexicon/test_planetary_computer_lexicon.py
```diff
@@ -85,3 +86,20 @@ def test_modis_fire_lexicon(variable: str) -> None:
     sample = np.random.rand(3, 3).astype(np.float32)
     out = modifier(sample.copy())
     assert np.allclose(out, sample)
+
+
+@pytest.mark.parametrize(
+    "variable",
+    ["u10m", "z500", "t850", "v100"],
+)
+def test_ifs_lexicon(variable: str) -> None:
```
NickGeneva: Nit: update test name to test_planetary_computer_ifx_lexicon

earth2studio/data/planetary_computer.py
```diff
@@ -871,3 +887,112 @@ def extract_variable_numpy(
...
+            # cfgrib remaps variable names that start with a number
+            "100u": "u100",
+            "100v": "v100",
+            "10u": "u10",
+            "10v": "v10",
+            "2t": "t2m",
+            "2d": "d2m",
+        }
+        with xr.open_dataset(
+            plan.local_path, engine="cfgrib", filter_by_keys={"shortName": var}
```
NickGeneva: Please use pygrib instead if possible, its much faster, see some of the other data sources for how to use pygrib and some of the error catch patterns

https://github.com/NVIDIA/earth2studio/blob/main/earth2studio/data/gfs.py#L389
https://github.com/NVIDIA/earth2studio/blob/main/earth2studio/data/ecmwf.py#L333

test/data/test_planetary_computer.py
```diff
@@ -239,3 +240,54 @@ def test_planetary_computer_modis_fire_cache(cache: bool) -> None:
...
+    data = ds(time=time, variable=variable)
+
+    times = list(time) if isinstance(time, (list, tuple)) else [time]
+    variables = list(variable) if isinstance(variable, (list, tuple)) else [variable]
+
+    assert data.shape == (len(times), len(variables), 721, 1440)
+    assert np.array_equal(data.coords["variable"].values, np.array(variables))
+
+
+@pytest.mark.slow
```
NickGeneva: To help keep testing overhead lower, I dont think the cache test is needed here since other planetary computer sub classes should check this, unless its changing something with how the cache is working

earth2studio/lexicon/planetary_computer.py
```diff
@@ -144,3 +144,114 @@ class MODISFireLexicon(metaclass=LexiconType):
     @classmethod
     def get_item(cls, val: str) -> tuple[str, Modifier]:
         return cls.VOCAB[val]
+
+
+class ECMWFOpenDataIFSLexicon(metaclass=LexiconType):
+    """Lexicon exposing ECMWF Open Data IFS variables."""
```
NickGeneva: is there any documentations on what variables are available that can be added for future reference here?

earth2studio/data/planetary_computer.py
swbg: We could do IFS_FX, the ocean model, etc., but I don't think AIFS is (officially) supported.

earth2studio/lexicon/planetary_computer.py
swbg: I got this by looking at one of the files. I could add an example URL if that's helpful.

earth2studio/data/planetary_computer.py
swbg: I am not sure. It is a generator and I thought maybe because it could return, e.g., 1M objects (the results are paged and I don't know if it would just retrieve all pages when calling len(items)). I have changed it now to never return more than 2 items (before it was 1 item per page) and then check len(items).

earth2studio/lexicon/planetary_computer.py
NickGeneva: Yeah an example URL would be good for future reference :+1:

earth2studio/data/planetary_computer.py
NickGeneva: I wanted to test this to see if it would work but I ran into an error:

```python
from earth2studio.data import PlanetaryComputerECMWFOpenDataIFS

ds = PlanetaryComputerECMWFOpenDataIFS()
# Access similar 
da = ds(datetime(2024,1,1), ["t2m"])
print(da)
```

```
File "/localhome/local-ngeneva/earth2studio/.venv/lib/python3.12/site-packages/xarray/core/indexing.py", line 1704, in _safe_setitem
    array[key] = value
    ~~~~~^^^^^
ValueError: could not broadcast input array from shape (451,900) into shape (721,1440)
```

swbg: Good catch, thanks. Until sometime early 2025, there are 2 versions available (0.4 degrees and 0.25 degrees). I have added an additional query/filter to only retrieve the items corresponding to 0.25 degrees. I have also added a time validation (there is no 0.25 degrees data available before 2024-03-01, same as for the IFS data source in ecwmf.py).

NickGeneva: Thanks @swbg 

Resolves the date issue, I am still seeing a time issue

```
from earth2studio.data import PlanetaryComputerECMWFOpenDataIFS

ds = PlanetaryComputerECMWFOpenDataIFS()
# Fails
da = ds(datetime(2024,5,1,6), ["t2m"])
print(da)
# Works
da = ds(datetime(2024,5,1,12), ["t2m"])
print(da)
# Fails
da = ds(datetime(2024,5,1,18), ["t2m"])
print(da)
```

The 6 and 18 hour query seems to just always fail, all the time with the

```
FileNotFoundError: No Planetary Computer item found for 2024-05-11T18:00:00+00:00 within ±0:00:00
```

I'm worried some part of the query `_locate_item` function is not working as intended. I see the the IFS data is offered at 6 hour intervals:

https://planetarycomputer.microsoft.com/dataset/ecmwf-forecast



Also MSFT seems super slow, I'm guessing if you have an API its faster? Is there a suggestion for users?

swbg: I was indeed only testing with 00z and 12z, which are available through the oper stream, but not 06z and 18z, which are available through scda. Timing seems ok to me, ca. 30 s for ca. 70 variables. May be faster when working on Azure.

NickGeneva: Nice, I can initialize AIFS now!

PR #683, opened 2026-02-10
[Service] Initial MR with a skeleton service.

gh0st-ryder:
# Earth2Studio Pull Request

## Description
* Adds a `service` directory where the E2S service code will be upstreamed.
* Adds the startup scripts that we run on Lepton and Azure
* Adds some code for the config, along with the default config. Adds unit tests for the config.
* Adds a skeleton main, that will be fleshed out in the following PRs.
* Adds a separate requirements and Makefile. A dockerfile will be added soon. The goal is to build a docker container for the service using this setup, for deployment on Lepton and Azure.

Note that no openapi spec is added yet. 

## Checklist

### DIFF
--- a/test/service/test_config.py
+++ b/test/service/test_config.py
@@ -34,6 +34,7 @@ mock_global_hydra = MagicMock()
 mock_global_hydra_instance = MagicMock()
 mock_global_hydra.instance.return_value = mock_global_hydra_instance
 
+# TODO: This leaks into other test envs. Danger!
 sys.modules["hydra"] = mock_hydra
 sys.modules["hydra.core"] = MagicMock()
 sys.modules["hydra.core.global_hydra"] = mock_global_hydra


service/inferenceserver/api_server/config.py
```diff
@@ -0,0 +1,497 @@
...
+class CORSConfig:
+    """CORS middleware configuration"""
+
+    allow_origins: list = field(default_factory=lambda: ["*"])
+    allow_credentials: bool = True
+    allow_methods: list = field(default_factory=lambda: ["*"])
+    allow_headers: list = field(default_factory=lambda: ["*"])
+
+
+@dataclass
```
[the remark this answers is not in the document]
gh0st-ryder: Mentioning for completeness here. This PR only brings in a few scripts, the config code, and a skeleton (empty) main.py. All other service code, including the openapi spec, the implementation, the code that interacts with redis, the queues, etc. will be upstreamed in the next MR.

I purposely kept this MR small so I could get past any initial hurdles with the precommit hooks, or being able to run the CI pipelines, etc. It is not intended to be a functional service yet. The next few series of MRs will flesh out the service.

service/inferenceserver/scripts/start_api_server.sh
```diff
@@ -0,0 +1,259 @@
...
+    sleep $HEALTH_RETRY_INTERVAL
+done
+
+# Invoke example_user_workflow to warm up the server
+echo "Invoking example_user_workflow to warm up the server..."
+curl -s -X POST "http://localhost:$API_PORT/v1/infer/example_user_workflow" \
+    -H "Content-Type: application/json" \
+    -d '{"parameters": {"task_name": "warmup", "num_iterations": 1, "delay_seconds": 0.1, "generate_output": false}}' \
+    && echo "" && echo "Example workflow invoked successfully!" \
+    || echo "Warning: Failed to invoke example workflow"
```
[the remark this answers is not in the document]
gh0st-ryder: That's correct. As mentioned previously, this PR isn't intended to implement the fully functional (or even basic) service yet. We will upstream that implementation gradually over the next few PRs.

test/service/test_config.py
```diff
@@ -0,0 +1,623 @@
...
+mock_hydra.compose = MagicMock()
+mock_hydra.initialize_config_dir = MagicMock()
+
+# Mock omegaconf
+mock_omegaconf = MagicMock()
+sys.modules["omegaconf"] = mock_omegaconf
+mock_omegaconf.OmegaConf = MagicMock()
+mock_omegaconf.OmegaConf.to_container = MagicMock(return_value={})
+
+# Import config after mocking dependencies (noqa: E402 - import after mocks is intentional)
```
NickGeneva, line 50: This is pretty gnarly... I think since this test is indeed for the api_server items we should keep it in the service folder. We just wont run it in the CI system.

pyproject.toml
```diff
@@ -336,6 +336,7 @@ exclude = []
 
 # Ignore `S101` (assertions) in all `test` files.
 "test/*.py" = ["S101"]
+"**/test/**/*.py" = ["S101", "S108"]
```
NickGeneva: Is this still needed? What is the cause of the S108 issue?

requirements.txt
```diff
@@ -5,6 +5,8 @@
 
 aiobotocore==3.1.1
```
NickGeneva, line 6: I dont think the requirements.txt should be getting changed here.
Perhaps these are left over?

tox-min.ini
```diff
@@ -44,7 +44,8 @@ commands =
         test/utils/test_coords.py \
         test/utils/test_imports.py \
         test/models/test_auto_models.py \
-        test/models/test_batch.py
+        test/models/test_batch.py \
+        test/service/test_config.py
```
NickGeneva: For now lets skip the adding stuff to the tox file and reserve this for tests that are in the package.

NickGeneva (commented):
General naming comment,

Thoughts on renaming "inferenceserver" folder to just "server"

Earth2studio is an inference package.

If you want to keep it we should rename it to "inference_server" (granted this then raises questions why have a folder inside called "api_server" since this already is inside a server folder.

Something is redundant here.

gh0st-ryder:
> General naming comment,
> 
> Thoughts on renaming "inferenceserver" folder to just "server"
> 
> Earth2studio is an inference package.
> 
> If you want to keep it we should rename it to "inference_server" (granted this then raises questions why have a folder inside called "api_server" since this already is inside a server folder.
> 
> Something is redundant here.

Yeah I can see the value in calling it "server", since the inference portion is kinda redundant. The general philosophy behind the organization is to have a "service" folder, with a "server" and a "client" inside it. The server itself has multiple folders such as the api_server itself, scripts, config, etc.

I'm OK changing "inferenceserver" to "server". But any opposition to doing it in one swoop after the follow-up PR gets checked in as well? As it stands, I have quite a bit of follow up code under review that all references "inferenceserver". Kind of a pain updating it currently. It would be easier to do it in one swoop after the dependent PRs are all merged. 

Thoughts?

test/service/test_config.py
gh0st-ryder: You bring up a good point. It is gnarly indeed. :-) Also I'm not sure the config test itself adds much value to be honest. Probably only adds to the brittleness of the testing, where it becomes a chore to fix up minor things. The follow up PR has several good tests for the rest of the server so those should suffice. 

Hmm... I'm thinking I might want to nuke this test file entirely. Thoughts?

My follow up PR brings in all the api_server unit tests at this main test/service level to enable testing in CI though. I do feel that has value to leave them at this level. So long as the tests add value and don't unnecessarily break things due to brittleness, they could live here I feel.

requirements.txt
gh0st-ryder: From earlier conversations, I was under the impression that we want to merge the service specific requirements.txt into this main one. Likewise for the service specific pyproject.toml. The follow-up PR does this. 

Do we want to keep them isolated within the service/server folder instead? If so, I can re-work the follow up PR and undo these changes here as well.

I'm not opposed to keeping them isolated. The service will have its own Dockerfile that installs the requirements it needs to the docker image. Makes sense to me that we don't mess with the dependencies at this level.

pyproject.toml
gh0st-ryder: Yeah good catch. This must have been left over from when I needed it before I modified the test. Looks like I don't require it anymore and I will remove it.

tox-min.ini
gh0st-ryder: Sounds good. I think if we nuke this particular test, this change goes away from this PR anyway.

Like I said earlier, I do feel there's value in the api_server tests making it to here. But I'll leave the final word to you to enforce, as the owner of the repo. :-)

requirements.txt
NickGeneva: Yeah, the pyproject edits in the toml do this.

> I'm not opposed to keeping them isolated. The service will have its own Dockerfile that installs the requirements it needs to the docker image.

Oh I see the need, yeah if you need a requirements.txt for the docker file, I would keep it in dedicated file in the service folder. The reason being this one will sometimes get automatically regenerated. So its not a safe space to keep deps for a docker file.

tox-min.ini
NickGeneva: Yeah, I'm thinking we will need a dedicated testing env or stage for the api server rather than adding it into other env for the package

NickGeneva (approved):

test/service/test_config.py
```diff
@@ -0,0 +1,615 @@
...
+mock_global_hydra.instance.return_value = mock_global_hydra_instance
+
+sys.modules["hydra"] = mock_hydra
+sys.modules["hydra.core"] = MagicMock()
+sys.modules["hydra.core.global_hydra"] = mock_global_hydra
+mock_hydra.compose = MagicMock()
+mock_hydra.initialize_config_dir = MagicMock()
+
+# Mock omegaconf
+mock_omegaconf = MagicMock()
```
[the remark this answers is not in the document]
NickGeneva: Added a comment here to fix in the future, valid concern.

test/service/test_config.py
NickGeneva: Lets ignore this for now, can address anything in the next PR.

PR #723, opened 2026-02-28
Split server

gh0st-ryder:
# Earth2Studio Pull Request

## Description
* Renames inferenceserver to server
* Splits serve/server into two parts: core changes (earth2studio/serve/server) and misc (serve/server)
* Tested on Lepton with the docker image by running the stormcast workflow.

## Checklist

### DIFF
--- a/test/serve/server/test_cleanup_daemon.py
+++ b/test/serve/server/test_cleanup_daemon.py
@@ -24,6 +24,7 @@ from earth2studio.serve.server.cleanup_daemon import (
     _delete_result_files,
     _process_expired_key,
     cleanup_expired_results,
+    main,
 )
 from earth2studio.serve.server.config import get_config
 from earth2studio.serve.server.workflow import WorkflowStatus
@@ -120,6 +121,102 @@ class TestDeleteResultFiles:
         # Verify Redis was queried
         mock_redis.get.assert_called_once()
 
+    def test_delete_result_files_zip_file_missing(self, mock_redis):
+        """When zip filename is in Redis but file does not exist, no unlink is called."""
+        result_id = "req_123"
+        mock_redis.get.return_value = "missing.zip"
+
+        mock_zip_path = MagicMock()
+        mock_zip_path.exists.return_value = False
+
+        mock_output_dir = MagicMock()
+        mock_output_dir.iterdir.return_value = []
+        mock_zip_dir = MagicMock()
+        mock_zip_dir.iterdir.return_value = []
+        mock_zip_dir.__truediv__ = lambda self, other: mock_zip_path
+
+        _delete_result_files(
+            mock_redis, result_id, result_id, mock_output_dir, mock_zip_dir
+        )
+
+        mock_zip_path.unlink.assert_not_called()
+
+    def test_delete_result_files_legacy_path_deletes_metadata_and_dirs(
+        self, mock_redis
+    ):
+        """Without workflow_name, still deletes metadata files and raw/zip dirs matching search_id."""
+        search_id = "exec_789"
+        result_id = search_id
+        mock_redis.get.return_value = None
+
+        mock_metadata_file = MagicMock()
+        mock_metadata_file.is_file.return_value = True
+        mock_metadata_file.is_dir.return_value = False
+        mock_metadata_file.name = f"metadata_{search_id}.json"
+        mock_metadata_file.unlink = MagicMock()
+
+        mock_raw_dir = MagicMock()
+        mock_raw_dir.is_dir.return_value = True
+        mock_raw_dir.name = f"output_{search_id}"
+        mock_raw_dir.__str__ = lambda self: f"output_{search_id}"
+
+        mock_zip_subdir = MagicMock()
+        mock_zip_subdir.is_dir.return_value = True
+        mock_zip_subdir.name = f"results_{search_id}"
+        mock_zip_subdir.__str__ = lambda self: f"results_{search_id}"
+
+        mock_output_dir = MagicMock()
+        mock_output_dir.iterdir.return_value = [mock_raw_dir]
+        mock_zip_dir = MagicMock()
+        mock_zip_dir.iterdir.return_value = [mock_metadata_file, mock_zip_subdir]
+
+        with patch(
+            "earth2studio.serve.server.cleanup_daemon.shutil.rmtree"
+        ) as mock_rmtree:
+            _delete_result_files(
+                mock_redis,
+                result_id,
+                search_id,
+                mock_output_dir,
+                mock_zip_dir,
+            )
+
+        mock_metadata_file.unlink.assert_called_once()
+        assert mock_rmtree.call_count == 2  # raw dir + zip_dir subdir
+        rmtree_calls = {str(c[0][0]) for c in mock_rmtree.call_args_list}
+        assert f"output_{search_id}" in rmtree_calls or mock_rmtree.call_count == 2
+
+    def test_delete_result_files_workflow_dir_not_exists(self, mock_redis):
+        """When workflow_name is set but workflow subdirectory does not exist, no rmtree on it."""
+        workflow_name = "my_wf"
+        execution_id = "exec_1"
+        combined_id = f"{workflow_name}:{execution_id}"
+        mock_redis.get.return_value = None
+
+        mock_output_dir = MagicMock()
+        mock_output_dir.iterdir.return_value = []
+        mock_workflow_dir = MagicMock()
+        mock_workflow_dir.exists.return_value = False  # workflow dir not present
+        mock_output_dir.__truediv__ = lambda self, other: mock_workflow_dir
+
+        mock_zip_dir = MagicMock()
+        mock_zip_dir.iterdir.return_value = []
+
+        with patch(
+            "earth2studio.serve.server.cleanup_daemon.shutil.rmtree"
+        ) as mock_rmtree:
+            _delete_result_files(
+                mock_redis,
+                combined_id,
+                execution_id,
+                mock_output_dir,
+                mock_zip_dir,
+                workflow_name=workflow_name,
+            )
+
+        mock_workflow_dir.iterdir.assert_not_called()
+        mock_rmtree.assert_not_called()
+
 
 class TestProcessExpiredKey:
     """Test suite for _process_expired_key function"""
@@ -204,6 +301,127 @@ class TestProcessExpiredKey:
         )
 
         # Verify - should not clean up
+        assert result is False
+        mock_delete_func.assert_not_called()
+        mock_redis.setex.assert_not_called()
+
+    def test_process_expired_key_key_missing(self, mock_redis, mock_delete_func):
+        """When Redis returns None for the key, return False without calling delete."""
+        mock_redis.get.return_value = None
+
+        result = _process_expired_key(
+            redis_client=mock_redis,
+            key="workflow_execution:wf1:exec_123",
+            current_time=datetime.now(timezone.utc),
+            results_ttl_hours=24,
+            retention_ttl=604800,
+            expected_status=WorkflowStatus.COMPLETED,
+            get_end_time_field="end_time",
+            delete_files_func=mock_delete_func,
+            log_prefix="workflow",
+        )
+
+        assert result is False
+        mock_delete_func.assert_not_called()
+        mock_redis.setex.assert_not_called()
+
+    def test_process_expired_key_wrong_status(self, mock_redis, mock_delete_func):
+        """When status is not expected (e.g. RUNNING), return False."""
+        current_time = datetime.now(timezone.utc)
+        old_time = (current_time - timedelta(hours=25)).isoformat()
+        mock_redis.get.return_value = json.dumps(
+            {"status": WorkflowStatus.RUNNING, "end_time": old_time}
+        )
+
+        result = _process_expired_key(
+            redis_client=mock_redis,
+            key="workflow_execution:wf1:exec_123",
+            current_time=current_time,
+            results_ttl_hours=24,
+            retention_ttl=604800,
+            expected_status=WorkflowStatus.COMPLETED,
+            get_end_time_field="end_time",
+            delete_files_func=mock_delete_func,
+            log_prefix="workflow",
+        )
+
+        assert result is False
+        mock_delete_func.assert_not_called()
+        mock_redis.setex.assert_not_called()
+
+    def test_process_expired_key_no_end_time(self, mock_redis, mock_delete_func):
+        """When end_time and completion_time are missing, return False."""
+        mock_redis.get.return_value = json.dumps({"status": WorkflowStatus.COMPLETED})
+
+        result = _process_expired_key(
+            redis_client=mock_redis,
+            key="workflow_execution:wf1:exec_123",
+            current_time=datetime.now(timezone.utc),
+            results_ttl_hours=24,
+            retention_ttl=604800,
+            expected_status=WorkflowStatus.COMPLETED,
+            get_end_time_field="end_time",
+            delete_files_func=mock_delete_func,
+            log_prefix="workflow",
+        )
+
+        assert result is False
+        mock_delete_func.assert_not_called()
+        mock_redis.setex.assert_not_called()
+
+    def test_process_expired_key_uses_completion_time_fallback(
+        self, mock_redis, mock_delete_func
+    ):
+        """When end_time is missing but completion_time is present, use it and expire."""
+        config = get_config()
+        current_time = datetime.now(timezone.utc)
+        old_time = (current_time - timedelta(hours=30)).isoformat()
+        mock_redis.get.return_value = json.dumps(
+            {
+                "status": WorkflowStatus.COMPLETED,
+                "completion_time": old_time,
+            }
+        )
+
+        result = _process_expired_key(
+            redis_client=mock_redis,
+            key="workflow_execution:wf1:exec_old",
+            current_time=current_time,
+            results_ttl_hours=24,
+            retention_ttl=config.redis.retention_ttl,
+            expected_status=WorkflowStatus.COMPLETED,
+            get_end_time_field="end_time",
+            delete_files_func=mock_delete_func,
+            log_prefix="workflow",
+        )
+
+        assert result is True
+        mock_delete_func.assert_called_once()
+        mock_redis.setex.assert_called_once()
+        updated = json.loads(mock_redis.setex.call_args[0][2])
+        assert updated["status"] == WorkflowStatus.EXPIRED
+
+    def test_process_expired_key_invalid_timestamp(self, mock_redis, mock_delete_func):
+        """When timestamp is invalid, log warning and return False."""
+        mock_redis.get.return_value = json.dumps(
+            {
+                "status": WorkflowStatus.COMPLETED,
+                "end_time": "not-a-valid-timestamp",
+            }
+        )
+
+        result = _process_expired_key(
+            redis_client=mock_redis,
+            key="workflow_execution:wf1:exec_123",
+            current_time=datetime.now(timezone.utc),
+            results_ttl_hours=24,
+            retention_ttl=604800,
+            expected_status=WorkflowStatus.COMPLETED,
+            get_end_time_field="end_time",
+            delete_files_func=mock_delete_func,
+            log_prefix="workflow",
+        )
+
         assert result is False
         mock_delete_func.assert_not_called()
         mock_redis.setex.assert_not_called()
@@ -449,3 +667,109 @@ class TestCleanupExpiredResults:
             assert (
                 mock_redis.setex.call_args[0][0] == "workflow_execution:wf_50h:exec_50h"
             )
+
+    def test_cleanup_expired_results_malformed_key_skipped(
+        self, mock_redis, cleanup_args
+    ):
+        """Keys with fewer than 3 parts (e.g. 'workflow_execution:incomplete') are skipped."""
+        current_time = datetime.now(timezone.utc)
+        old_time = (current_time - timedelta(hours=30)).isoformat()
+
+        mock_redis.keys.return_value = [
+            "workflow_execution:incomplete",  # only 2 parts
+            "workflow_execution:wf1:exec_old",
+        ]
+
+        def get_side_effect(key):
+            if key == "workflow_execution:wf1:exec_old":
+                return json.dumps(
+                    {"status": WorkflowStatus.COMPLETED, "end_time": old_time}
+                )
+            return None
+
+        mock_redis.get.side_effect = get_side_effect
+
+        with patch("earth2studio.serve.server.cleanup_daemon._delete_result_files"):
+            cleanup_expired_results(mock_redis, **cleanup_args)
+
+            # Only the valid key (exec_old) should be processed; malformed key is skipped
+        assert mock_redis.setex.call_count == 1
+        assert mock_redis.setex.call_args[0][0] == "workflow_execution:wf1:exec_old"
+
+
+class TestCleanupDaemonMain:
+    """Tests for main() daemon entrypoint."""
+
+    def test_main_redis_connection_failure_exits_with_one(self, tmp_path):
+        """When Redis connection or ping fails, main() calls sys.exit(1)."""
+        mock_config = MagicMock()
+        mock_config.paths.default_output_dir = str(tmp_path / "out")
+        mock_config.paths.results_zip_dir = str(tmp_path / "zip")
+        mock_config.server.results_ttl_hours = 24
+        mock_config.server.cleanup_watchdog_sec = 60
+        mock_config.redis.retention_ttl = 604800
+        mock_config.redis.host = "localhost"
+        mock_config.redis.port = 6379
+        mock_config.redis.db = 0
+        mock_config.redis.password = None
+        mock_config.redis.decode_responses = True
+        mock_config.redis.socket_connect_timeout = 5
+        mock_config.redis.socket_timeout = 5
+
+        mock_redis_instance = MagicMock()
+        mock_redis_instance.ping.side_effect = Exception("Connection refused")
+
+        with (
+            patch(
+                "earth2studio.serve.server.cleanup_daemon.get_config",
+                return_value=mock_config,
+            ),
+            patch(
+                "earth2studio.serve.server.cleanup_daemon.redis.Redis",
+                return_value=mock_redis_instance,
+            ),
+            patch("earth2studio.serve.server.cleanup_daemon.sys.exit") as mock_exit,
+        ):
+            mock_exit.side_effect = SystemExit(1)
+            with pytest.raises(SystemExit):
+                main()
+            mock_exit.assert_called_once_with(1)
+
+    def test_main_runs_one_cycle_then_handles_keyboard_interrupt(self, tmp_path):
+        """When time.sleep raises KeyboardInterrupt (e.g. SIGINT), main() exits gracefully and closes Redis."""
+        mock_config = MagicMock()
+        mock_config.paths.default_output_dir = str(tmp_path / "out")
+        mock_config.paths.results_zip_dir = str(tmp_path / "zip")
+        mock_config.server.results_ttl_hours = 24
+        mock_config.server.cleanup_watchdog_sec = 60
+        mock_config.redis.retention_ttl = 604800
+        mock_config.redis.host = "localhost"
+        mock_config.redis.port = 6379
+        mock_config.redis.db = 0
+        mock_config.redis.password = None
+        mock_config.redis.decode_responses = True
+        mock_config.redis.socket_connect_timeout = 5
+        mock_config.redis.socket_timeout = 5
+
+        mock_redis_instance = MagicMock()
+        mock_redis_instance.ping.return_value = True
+        mock_redis_instance.close = MagicMock()
+
+        with (
+            patch(
+                "earth2studio.serve.server.cleanup_daemon.get_config",
+                return_value=mock_config,
+            ),
+            patch(
+                "earth2studio.serve.server.cleanup_daemon.redis.Redis",
+                return_value=mock_redis_instance,
+            ),
+            patch("earth2studio.serve.server.cleanup_daemon.cleanup_expired_results"),
+            patch("earth2studio.serve.server.cleanup_daemon.signal.signal"),
+            patch(
+                "earth2studio.serve.server.cleanup_daemon.time.sleep",
+                side_effect=KeyboardInterrupt,
+            ),
+        ):
+            main()
+            mock_redis_instance.close.assert_called_once()
--- a/test/serve/server/test_cpu_worker.py
+++ b/test/serve/server/test_cpu_worker.py
@@ -20,7 +20,7 @@ import zipfile
 from dataclasses import dataclass, field
 from datetime import datetime
 from typing import Literal
-from unittest.mock import Mock
+from unittest.mock import Mock, patch
 
 import pytest
 import redis
@@ -144,7 +144,17 @@ pytest.importorskip("earth2studio.serve.server")
 sys.modules["earth2studio.serve.server.config"] = mock_config_module
 
 # Now import the module under test
-from earth2studio.serve.server.cpu_worker import create_results_zip  # noqa: E402
+from earth2studio.serve.server.cpu_worker import (  # noqa: E402  # noqa: E402
+    FileManifestEntry,
+    ResultMetadata,
+    build_file_manifest,
+    create_results_zip,
+    fail_workflow,
+    process_finalize_metadata,
+    process_object_storage_upload,
+    process_result_zip,
+)
+from earth2studio.serve.server.workflow import WorkflowResult  # noqa: E402
 
 
 class TestCreateResultsZip:
@@ -644,7 +654,7 @@ class TestCreateResultsZip:
         ]
 
         for f in required_fields:
-            assert f in metadata, f"Missing required field: {field}"
+            assert f in metadata, f"Missing required field: {f}"
 
         # Verify zip_created_at is a valid timestamp
         zip_created_at = metadata["zip_created_at"]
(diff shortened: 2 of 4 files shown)


earth2studio/serve/server/__init__.py
```diff
@@ -37,17 +30,21 @@
...
     if name == "Earth2Workflow":
-        from .e2workflow import Earth2Workflow
+        from earth2studio.serve.server import e2workflow
 
-        return Earth2Workflow
+        return e2workflow.Earth2Workflow
     raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
 
 
 __all__ = [
```
NickGeneva: Remove this, I don't do __all__ in this package because importing all should generally be considered bad practice

Lets remove these __all__ definitions to encourage explicit imports.

earth2studio/serve/server/config.py
```diff
@@ -176,8 +168,9 @@ def __init__(self) -> None:
     def _initialize_config(self) -> None:
         """Initialize configuration from Hydra"""
         try:
-            # Get config directory
-            config_dir = Path(__file__).parent / "conf"
+            # Config lives in serve/server/conf (unchanged location in repo)
```
NickGeneva, line 171: Im curious why this is a hard coded, why would this not be some sort of input argument / parameter / env var (not ideal)

earth2studio/serve/server/config.py
NickGeneva, line 166: I think this could be moved up into line 160.
Call `cls._instance._initialize_config()` in new and clean up the init

earth2studio/serve/server/cpu_worker.py
```diff
@@ -14,13 +14,6 @@
...
 # limitations under the License.
 
-"""
-CPU Worker Functions
-
-This module contains CPU-intensive functions that can be offloaded from the main worker process.
-
-"""
-
 __all__ = [
```
NickGeneva: Remove, avoid this in any files generally. Encourage explicit importing

earth2studio/serve/server/workflow.py
```diff
@@ -14,12 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""
```
NickGeneva, line 17: fyi, I deleted these, we dont add api docs to  the top of files like this in this package. 👍

earth2studio/serve/server/workflow.py
```diff
@@ -682,9 +676,12 @@ def discover_and_register_from_directories(
         Returns:
             tuple: (successful_imports, failed_imports) counts
         """
-        # Always include built-in workflows if requested
+        # Always include built-in workflows if requested (serve/server/example_workflows)
         if include_builtin:
-            builtin_workflows_dir = Path(__file__).parent.parent / "example_workflows"
+            _repo_root = Path(__file__).resolve().parent.parent.parent.parent
```
NickGeneva, line 768: Similar hard code question, why is this something written like this. Should be some parameter in like the config or something?

earth2studio/serve/server/workflow.py
NickGeneva, line 42: I'd be curious to learn more about the logging requirements... the rest of the package uses loguru for logging atm.

Depending on the complexity of whats needed may or not be possible to use here.

earth2studio/serve/server/worker.py
NickGeneva, line 47: Defining a ton of globals that are not constants is an anti-pattern which concerns me

earth2studio/serve/server/main.py
```diff
@@ -14,13 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
```
NickGeneva, line 15: Does the bottom of the file need to the

if __name___ == "__main__" etc?

`import uvicorn` should be moved up to the top

earth2studio/serve/server/object_storage.py
NickGeneva: Numpy style doc strings

earth2studio/serve/server/utils.py
NickGeneva: This is used a single time in this file, Just make this the actual type in the function for `current_stage` if its not used anywhere else.

earth2studio/serve/server/workflow.py
NickGeneva: Numpy style docs needed

earth2studio/serve/server/worker.py
NickGeneva: Check all functions are Numpy api docs and have parameter definitions if they are public

If they are private (lead with an underscore) feel free to not have an api doc or have just a small description like this

test/serve/server/test_workflow.py
```diff
@@ -24,25 +24,20 @@
...
+    WorkflowRegistry,
+    WorkflowResult,
+    WorkflowStatus,
+    parse_workflow_directories_from_env,
+    register_all_workflows,
+    workflow_registry,
+)
 
 # imitate API server environment (DANGER!!! REMOVE THIS)
 os.environ["EARTH2STUDIO_API_ACTIVE"] = "1"
```
NickGeneva: Can you update this to a pytest fixture or something that will set this for the test, and then unset it for the test? Setting it global like this will effect all the tests I think.

[I'm not 100% what the best way is myself...](https://stackoverflow.com/questions/77255758/how-can-i-mock-my-environment-variables-for-my-pytest)

earth2studio/serve/server/cleanup_daemon.py
NickGeneva: Why is this a global variable?

earth2studio/serve/server/cleanup_daemon.py
```diff
@@ -34,8 +27,8 @@
 
 import redis  # type: ignore[import-untyped]
 
-from api_server.config import get_config, get_config_manager
-from api_server.workflow import WorkflowStatus
+from earth2studio.serve.server.config import get_config, get_config_manager
+from earth2studio.serve.server.workflow import WorkflowStatus
 
 # Get configuration
 config = get_config()
```
NickGeneva: Dont initialize config global, initialize it inside of the main function then pass it into any down stream functions as needed to control scope

earth2studio/serve/server/config.py
NickGeneva: From what I can see this is not needed... just make the get_config_manager return the config object.

This is mixing globals and singletons which doesnt seem to make sense here because they are basically doing the same thing. The singleton (class instance) pattern is good, you should just rely on it.

NickGeneva, line 202: Is there a specific exception here you are expecting (like FileNotfound)? Catching all is not ideal if not needed.

NickGeneva, line 239: Why are you checking it see if the env set when you specify a default if its not set? Seems redundant.

NickGeneva, line 252: No default, what if a password is not provided

NickGeneva: This will return None if `REDIS_PASSWORD` is not set is that okay? Just confirming

NickGeneva, line 410: what is this doing?

NickGeneva, line 426: why are these checks needed? Isnt initialise_config always called on construction?

NickGeneva, line 392: why are these checks needed? Isnt initialise_config always called on construction?

earth2studio/serve/server/cpu_worker.py
```diff
@@ -75,7 +68,7 @@
 
 # Register custom workflows in the CPU worker process
 try:
-    from api_server.workflow import register_all_workflows
+    from earth2studio.serve.server.workflow import register_all_workflows
 
     register_all_workflows(redis_client)
```
NickGeneva, line 66: I'm wondering whats going on here, why is the register in the imports versus in some initialization / startup function? Isnt this called in the main?

earth2studio/serve/server/cpu_worker.py
NickGeneva, line 129: Still needed? Please clean up if not 👍

NickGeneva, line 123: This true, is this just for zip files?

NickGeneva, line 178: Needed still?

earth2studio/serve/server/cpu_worker.py
```diff
@@ -14,13 +14,6 @@
 # See the License for the specific language governing permissions and
```
NickGeneva, line 14: This file is called CPU worker but seems to be entirely just io functions? Or is the scope more?

earth2studio/serve/server/cpu_worker.py
NickGeneva, line 305: Seems this is used once, with the normal logger being used all other times. Whats the intent of this?

NickGeneva, line 497: Hmmm, why not `fail_workflow` in create_results_zip and raise the exception there so you get a trace of what occured

NickGeneva: log needed here, or will `fail_workflow` do the job, seems redundant

NickGeneva, line 836: Previously youve used logger.error.
Message different between this and fail_workflow, can one just use `fail_workflow`

earth2studio/serve/server/cleanup_daemon.py
NickGeneva, line 315: I dont think this should be here, left over?

NickGeneva: Maybe this needs a log in it to let the user its attempting to shut down? Could just loop and sleep with no feedback to the user right?

earth2studio/serve/server/cpu_worker.py
```diff
@@ -768,7 +761,7 @@ def process_finalize_metadata(
     Returns:
         Dict containing result info, None on critical failure
     """
-    from api_server.workflow import WorkflowStatus, workflow_registry
+    from earth2studio.serve.server.workflow import WorkflowStatus, workflow_registry
```
NickGeneva: This could be moved up top right? No real need to the lazy import here I think

earth2studio/serve/server/cpu_worker.py
NickGeneva, line 325: Hard coding compression level seems like a bad idea, but I dont mind since I prefer we view this method as a last resort (compared to using object/blob storage)

NickGeneva, line 328: Why this if statement again, didnt you just check this a few lines above?

NickGeneva, line 329: When would the be true? seems to me an "output_path" should always be a path... and the zip is called zip_filename or something the user specifies...

earth2studio/serve/server/__init__.py
gh0st-ryder: Done.

earth2studio/serve/server/config.py
gh0st-ryder: Yeah this has been a TODO issue on our list for re-work, though I'd like to keep it separate from the upstreaming process. For what it's worth, the server startup script does provide a way to override the config in case the user doesn't want the baked-in out-of-the-box experience.

earth2studio/serve/server/config.py
gh0st-ryder: Hmm, I thought new was for object creation and init for initialization? In that sense, _initialize_config() seems appropriately located? Am I missing some Python-ism here, or is it just preference? I'm happy to do it if you want, just trying to understand the rationale.

earth2studio/serve/server/cpu_worker.py
gh0st-ryder: Done.

earth2studio/serve/server/workflow.py
gh0st-ryder: Ack. Thanks!

earth2studio/serve/server/workflow.py
gh0st-ryder: See previous reason.

earth2studio/serve/server/workflow.py
gh0st-ryder: I don't think we gave the logging package much thought to be honest. Neither of us have much Python experience, so we went with what felt appropriate at the time. I'm happy to re-work the logging in a separate sweep later if you want us to stick to certain packages though.

earth2studio/serve/server/worker.py
gh0st-ryder: Yeah that's a very fair criticism. Let me put this on the list of things to re-work during the next sweep, and I will minimize the usage.

earth2studio/serve/server/main.py
gh0st-ryder: Done.

earth2studio/serve/server/object_storage.py
gh0st-ryder: Done.

earth2studio/serve/server/utils.py
gh0st-ryder: Done.

earth2studio/serve/server/workflow.py
gh0st-ryder: Done.

earth2studio/serve/server/worker.py
gh0st-ryder: Done.

test/serve/server/test_workflow.py
gh0st-ryder: Done.

earth2studio/serve/server/cleanup_daemon.py
gh0st-ryder: Reworked and removed.

earth2studio/serve/server/cleanup_daemon.py
gh0st-ryder: Done.

earth2studio/serve/server/config.py
gh0st-ryder: Done.

gh0st-ryder: This is pretty broadly used in the server as well as client to catch any unexpected failures. I'll argue that we can leave this be as is.

gh0st-ryder: Honestly, this whole section needs to be re-thought. I've brought it up before with the team. Currently we specify settings in the config, and have env overrides for them. I personally feel that only a few select env vars should be enabled, instead of letting us override them willy-nilly. Anyway, that's the reason why the default goes to the config.

gh0st-ryder: Yes, should be OK. The server endpoint will fail fast in that case, which would be expected.

gh0st-ryder: Not all log statements know or have the execution_id contextual information. e.g. logs before an exec_id is generated. In cases where it is missing, this adds the execution_id to the record, but leaves it blank. Basically this is in the broader context of as many log statements as possible carrying an "execution_id" where possible, so debugging is easy.

gh0st-ryder: That's a good question. To be frank, I've lost the context on this since the code was written a while ago. You may be right, but I'm inclined to leave this untouched for now on the off chance it breaks something unexpected. I realize that's a bit unsatisfactory, but it's true. :-P

gh0st-ryder: Same as above.

earth2studio/serve/server/cpu_worker.py
gh0st-ryder: Registration happens in main (main server thread), worker (GPU processes), as well as cpu_worker (CPU I/O processes). These are all separate rq processes, and therefore require initialization of the custom workflows.

earth2studio/serve/server/cpu_worker.py
gh0st-ryder: Yeah looks like some vestiges of legacy stuff that needs to go away. If you don't mind, I'd definitely want to pull this into the follow-up sweep PR. The reason being that this has a high potential to mess up the e2e functionality, so it will be easier to debug and test as a follow-on. I've added it to my list.

gh0st-ryder: That's correct I believe.

gh0st-ryder: Same comment as above for the legacy cleanup.

earth2studio/serve/server/cpu_worker.py
gh0st-ryder: That's correct. This is by design. We want to run the inference via worker.py (GPU workers) and I/O via cpu_worker.py.

earth2studio/serve/server/cpu_worker.py
gh0st-ryder: Yeah...uh... that seems to be an oversight. Everything in this function should be logging with that additional execution_id.
I fixed it, thanks for the catch!

gh0st-ryder: The answer is that this code was developed organically at a *later* point, once we decided to add the ability to skip zip creation (while using obj store for example). It was convenient to add it here at that point. Regardless, I think both ways work fine, so I'll leave this in place for now.

gh0st-ryder: Done.

gh0st-ryder: In this case I believe logger.exception adds some additional exception information/traceback.
I agree it is a bit redundant with the fail_workflow error logging, but I can't switch that to logger.exception since that isn't always the case when it is called. 
I think this should be infrequent enough that we can live with a bit of logging redundancy?

earth2studio/serve/server/cleanup_daemon.py
gh0st-ryder: Done.

gh0st-ryder: So this part executes so long as the user didn't signal shutdown. Basically the watchdog wakes up periodically and cleans up stale and expired results.
cleanup_expired_results itself will indeed log that it is starting a cleanup cycle.

earth2studio/serve/server/cpu_worker.py
gh0st-ryder: Done.

earth2studio/serve/server/cpu_worker.py
gh0st-ryder: Ack. Makes sense.

gh0st-ryder: Looks like the previous check only logs a warning, but still keeps the control flow moving. In this case, effectively if the output_path didn't exist, looks like it is skipping zip creation.

Basically this looks to be the result of growing the code organically. Likely the inefficiency was added during obj store support when the output path not existing wasn't a showstopper anymore. 

I realize it is inefficient here, but not necessarily wrong. I'm inclined to leave it in place for now, since a cleanup here may cascade into unintended functional effects in this already big PR unless we also test the obj store paths e2e.

gh0st-ryder: See above.

gh0st-ryder:
@greptile-ai

gh0st-ryder:
@greptile-ai

gh0st-ryder:
@greptile-ai

earth2studio/serve/server/main.py
NickGeneva, line 71: I have a concern about these globals, outside of my usual feedback.

When dealing with fastapi apps one needs to be careful with sharing values between endpoints. (multiple workers threads this could cause race conditions)

Using fast api apps state is likely what youre looking for:
https://stackoverflow.com/questions/71260288/how-to-share-variables-between-http-requests-in-fastapi

I would suggest using this

NickGeneva, line 253: would be nice if this can get placed behind a function so its not eagerly created on import... might not be easily possible with fast API however.

NickGeneva: Having it behind a function would also allow you to let the caller provide the config object rather than doing this repeat get_config() at the top of each file.

earth2studio/serve/server/main.py
```diff
@@ -265,7 +259,9 @@ async def health_check() -> dict[str, str]:
     """Health check endpoint using status script"""
     try:
         # Run the status script and check exit code
-        script_path = Path(__file__).parent.parent / "scripts" / "status.sh"
+        # Scripts live at repo serve/server/scripts, not inside the package
+        _repo_root = Path(__file__).resolve().parent.parent.parent.parent
```
NickGeneva, line 296: Similar hard code question, why is this here. I think this should be something configurable.

Putting that asside. Calling a bash script from a python script is just generally bad. Can you elaborate on what this is doing?

earth2studio/serve/server/main.py
```diff
@@ -875,8 +871,6 @@ async def stream_file() -> AsyncGenerator[bytes, None]:
 
 
 if __name__ == "__main__":
```
NickGeneva, line 1043: This main should be moved out to some where in the sever container or something since libraries provide functionality to get invoked. Users should not invoke the files directly

earth2studio/serve/server/main.py
NickGeneva: Some of these with actual parameters should have some better doc strings I'm guessing might show up...

Not sure if fast api needs something special for openapi spec

NickGeneva, line 387: Why do some of these have dual endpoints?

NickGeneva: you can turn this whole into on big function, could split some APIs out into routers to better organize... theres some options.

https://fastapi.tiangolo.com/tutorial/bigger-applications/#add-some-custom-tags-responses-and-dependencies

earth2studio/serve/server/object_storage.py
```diff
@@ -145,35 +173,43 @@ def generate_signed_url(self, remote_key: str, expires_in: int = 86400) -> str:
...
     pass
 
 
 class MSCObjectStorage(ObjectStorage):
     """
-    Object storage implementation using NVIDIA Multi-Storage Client with Rust backend.
-
-    MSC provides optimized parallel transfers and the Rust client bypasses Python's
-    GIL for significantly improved I/O performance (up to 12x faster).
+    Object storage using NVIDIA Multi-Storage Client (MSC) with Rust backend.
```
NickGeneva, line 201: Move the init parameters here.

E.g. https://github.com/NVIDIA/earth2studio/blob/main/earth2studio/models/px/aifs.py#L71

the inits should back no doc string

earth2studio/serve/server/object_storage.py
```diff
@@ -585,28 +688,36 @@ def get_object_storage(
...
-        ...     use_rust_client=True
-        ... )
+    Create an object storage instance.
+
+    Convenience factory to instantiate MSCObjectStorage (NVIDIA Multi-Storage
+    Client with Rust backend).
+
+    Parameters
+    ----------
+    provider : {"msc"}, optional
```
NickGeneva: type should be Literal["msc"]

earth2studio/serve/server/object_storage.py
NickGeneva: General question, is this function used somewhere?

I guess all this is doing is going from string to class which could be handled by the config parsing section.

But wasnt able to find this being used

earth2studio/serve/server/workflow.py
NickGeneva: The "Any" here should be pretty easy to be more specific looking at this function. Avoid Any when possible

NickGeneva: Might not be possible because of pydantic...

earth2studio/serve/server/workflow.py
```diff
@@ -78,14 +72,20 @@ def validate(
         It ensures that all required fields are present, types are correct, and any
```
NickGeneva, line 72: This just seems odd to do, is this really required 🤔 Doesnt pydantics `model_validate` do this type of thing already?

earth2studio/serve/server/main.py
shrek: For azure, there needed to be a common prefix for all endpoints and it couldnt be "/v1". So, the common prefix was made to be "/v1/infer". However, for backward compatibility, "/v1/workflows" was retained. Hence there are some dual endpoints. But, we could rationalize to keep a common prefix "/v1/infer" and remove the older paths as there is no need for backward compatibility.

earth2studio/serve/server/worker.py
NickGeneva, line 1: WorkflowRegistry should be refactored into a singleton pattern, so the global registry is stored in the class variable. So every use of the registry can import, initialize, then use it.

earth2studio/serve/server/workflow.py
NickGeneva, line 1: I think some refactoring could be done to improve the control of where custom workflows are... right nows its hard coded... I'm wondering if some configuration or something in the redis client could help control this?

Specifically what is used in `discover_and_register_from_directories`

earth2studio/serve/server/e2workflow.py
NickGeneva, line 1: Just commenting... didnt full review this one. this file needs some clean up but more importantly unifying with the main workflow. Having two is confusing.

I see why this one is added however, its interfacing better with the packages typical patterns.

earth2studio/serve/server/config.py
NickGeneva: What I'm referring to is the use of a conditional check 

if os.getenv("REDIS_HOST")

to then just call

self._config.redis.host = os.getenv(
                "REDIS_HOST", default=self._config.redis.host
            )

that later that handles the case os.getenv("REDIS_HOST") is None anyway. Its a redundant check I think to have these if statements.

NickGeneva: Got it, interesting... thanks for the clarification.

earth2studio/serve/server/cpu_worker.py
NickGeneva: Ah I see, perhaps there needs to be some worker init or start up function these workers could use?

But see the reason, thanks for the info

earth2studio/serve/server/cpu_worker.py
NickGeneva: Fine to defer, thanks for checking!

NickGeneva: can defer to later PR if you want.

NickGeneva: Gotcha, okay lets leave it for now. missed that its just a warning.

earth2studio/serve/server/worker.py
shrek: Makes sense, and will fix.

earth2studio/serve/server/e2workflow.py
shrek: The purpose for this class "Earth2Workflow" is to completely abstract the fact that a user script is running as part of the inference-server. A user by deriving from this class can "locally" test their inference scripts on a gpu outside of the inference server. ie, the same script works standalone. For example, I just tested one of the scripts locally on a gpu like this, and the same script could be used as an API.

```
from earth2studio.io import NetCDF4Backend

from foundry_fcn3 import FoundryFCN3Workflow

io = NetCDF4Backend("test.nc")
workflow = FoundryFCN3Workflow()
workflow(io)
```

However, for users, it could be confusing why there are 2 Workflow classes, and whats the distinction between them, and docs would help here to clarify.

earth2studio/serve/server/workflow.py
shrek: There is an environment variable WORKFLOW_DIR allowing users to specify which directories contain user workflow scripts. This helps users put their custom scripts in a folder that could be mounted onto a prebuilt container which is a scenario on some cloud deployments; for example, lepton has an NFS folder that is mounted onto the container.

earth2studio/serve/server/main.py
NickGeneva: I'm all for doing some clean up since this backwards compatibility is just for development, now is probably the best time but can be in another PR as well.

Up to you guys. :+1:

earth2studio/serve/server/workflow.py
NickGeneva: Ah okay, good that something exists as is.
However I still find the current builtin defaults a bit troublesome... 

https://github.com/gh0st-ryder/earth2studio/blob/4a3dae3eae424d68b8fb4541ce63ebd7f382dc30/earth2studio/serve/server/workflow.py#L763

It shouldnt ever be based on the file location... because install location of the package could change user to user (could be in docker, uv, conda etc)

If including a default is necessary, then it should be based on current working directory not the file. That or you force users to specify WORKFLOW_DIR / provide some sort of paramter

earth2studio/serve/server/main.py
gh0st-ryder: Added to the todo list to re-work using app.state.

gh0st-ryder: Hmm, I would push back on this suggestion. The fastapi documentation as well as numerous official examples suggest creating a single app global that is used as needed. It seems to be a common usage pattern. The doc you linked also has a main.py underneath that does the same.

earth2studio/serve/server/main.py
gh0st-ryder: Basically we run a status script once the server starts up. The purpose is basically to do a health check of the various components that are running, and report a red or green signal (along with details on the status of various components)

We also have a /health endpoint path via REST. We decided to piggyback off the available functionality in the shell script to furnish an answer to the health check query for the REST API. 

I do agree it isn't ideal for the Python function to call the shell script. Perhaps we could re-work this in the future to have a Python implementation, and call that in the shell script instead. Then the python endpoint response here could execute the python variant as well. I'll add that to our TODO list.

earth2studio/serve/server/main.py
gh0st-ryder: Hmm, I'm a bit unclear on where the suggestion to move it is?  Can you elaborate? Also not sure how well things would function on moving this out, given that the api_server is an individual process that starts up via this?

earth2studio/serve/server/main.py
gh0st-ryder: Done.

gh0st-ryder: Adding this to the list as discussed above.

earth2studio/serve/server/object_storage.py
gh0st-ryder: Done.

earth2studio/serve/server/object_storage.py
gh0st-ryder: Done.

earth2studio/serve/server/workflow.py
shrek: model validation would cover a majority of field validations, and perhaps even some relationship between fields. This method is a placeholder for workflows to do any additional validation checks - for example, involving a combination of fields or any other system specific validations that are outside the scope of model validations itself. granted, most cases will be covered by pydantic model validations.

earth2studio/serve/server/object_storage.py
gh0st-ryder: Deleted.

earth2studio/serve/server/workflow.py
gh0st-ryder: Done.

earth2studio/serve/server/worker.py
gh0st-ryder: Agree that this is a good idea. However this seems like a significant ask OTOH. I'll definitely put this on the follow-up list of asks.

earth2studio/serve/server/main.py
NickGeneva: Sounds good

earth2studio/serve/server/main.py
NickGeneva: In the container / whatever is running the server there needs to be a python file / script file that runs stuff anyway.

So things like this `if __name__ == "__main__":` end up being downstream in whatever is consuming the / starting the fast API.

For now lets leave this as is and revisit

earth2studio/serve/server/workflow.py
NickGeneva: Got it, thanks for clarification.

earth2studio/serve/server/e2workflow.py
NickGeneva: Yeah, we need to rework that then. Having two solutions / workflow base classes will be confusing...

Deferring to later PR

earth2studio/serve/server/workflow.py
NickGeneva: Deferring to later PR

shrek: makes sense, and agree. we will fix this.

NickGeneva (approved):
Lgtm, thanks for all the edits!

earth2studio/serve/server/config.py
NickGeneva: Explicit init works well and easier, lets keep it!

earth2studio/serve/server/main.py
```diff
@@ -265,7 +265,9 @@ async def health_check() -> dict[str, str]:
     """Health check endpoint using status script"""
     try:
         # Run the status script and check exit code
-        script_path = Path(__file__).parent.parent / "scripts" / "status.sh"
+        # Scripts live at repo serve/server/scripts, not inside the package
+        _repo_root = Path(__file__).resolve().parent.parent.parent.parent
+        script_path = _repo_root / "serve" / "server" / "scripts" / "status.sh"
         process = await asyncio.create_subprocess_exec(
             str(script_path),
```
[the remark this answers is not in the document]
NickGeneva: Hard codes to be addressed in future PR

earth2studio/serve/server/config.py
```diff
@@ -176,8 +168,9 @@ def __init__(self) -> None:
...
+            _repo_root = Path(__file__).resolve().parent.parent.parent.parent
+            config_dir = _repo_root / "serve" / "server" / "conf"
```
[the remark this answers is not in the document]
NickGeneva: Hard codes to be addressed in future PR

test/serve/server/conftest.py
```diff
@@ -0,0 +1,95 @@
...
+"""
+
+import sys
+from pathlib import Path
+
+# Repo root is parent of test/serve/server; ensure project is importable
+_repo_root = Path(__file__).resolve().parent.parent.parent
+if str(_repo_root) not in sys.path:
+    sys.path.insert(0, str(_repo_root))
+
```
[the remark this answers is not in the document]
NickGeneva: Hard codes to be addressed in future PR

PR #747, opened 2026-03-12
Clean up ECMWF data sources

swbg:
# Earth2Studio Pull Request

## Description
The ECMWF data sources had some additional, unused code and xarray coords were not cleanly removed. Changes:
* Rename _call to fetch and remove the previously unused fetch methods
* Rename "sample" to "ensemble" for consistency with other parts of Earth2Studio
* Rename fetch_wrapper to _download_wrapper, because it actually wraps the download method

Two questions:
* Should we remove IFS_ENS? I see the point of keeping IFS in addition to IFS_FX because IFS has been around for longer, but we introduced IFS_ENS alongside IFS_ENS_FX, which has the same functionality. We also don't have AIFS_ENS, just AIFS_ENS_FX
* The _ENS subclasses go out of their way to not allow fetching multiple ensemble members at once. Should we add this functionality back?

## Checklist

### DIFF
--- a/earth2studio/data/ecmwf.py
+++ b/earth2studio/data/ecmwf.py
@@ -177,13 +177,14 @@ class _ECMWFOpenDataSource(ABC):
 
         xr_array = loop.run_until_complete(
             asyncio.wait_for(
-                self.fetch(time, lead_time, variable), timeout=self.async_timeout
+                self._ecmwf_fetch(time, lead_time, variable),
+                timeout=self.async_timeout,
             )
         )
 
         return xr_array
 
-    async def fetch(
+    async def _ecmwf_fetch(
         self,
         time: datetime | list[datetime] | TimeArray,
         lead_time: timedelta | list[timedelta] | LeadTimeArray,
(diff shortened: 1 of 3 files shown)


NickGeneva:
> Should we remove IFS_ENS? I see the point of keeping IFS in addition to IFS_FX because IFS has been around for longer, but we introduced IFS_ENS alongside IFS_ENS_FX, which has the same functionality. We also don't have AIFS_ENS, just AIFS_ENS_FX

The non FX ones are useful when someone just wants the analysis states used by the forecast... in the case of when a model needs history the non FX version is useful. The FX is when the lead_time needs to reference the actual forecast steps.

> The _ENS subclasses go out of their way to not allow fetching multiple ensemble members at once. Should we add this functionality back?

Right now none of the other ensemble classes output multiple ensembles. I think I prefer to keep the consistency of always returning [time, variable, lat, lon] over adding in the potential for the method to return an additional ensemble dim.

But if this would be particularly useful, perhaps we can change the behaviors but other ENS models like GEFS would need a similar update.

earth2studio/data/ecmwf.py
```diff
@@ -140,17 +140,7 @@ def __call__(  # type: ignore[override]
         """Retrieve ECMWF data. The child class should override this"""
         pass
 
-    @abstractmethod
-    async def fetch(  # type: ignore[override]
```
NickGeneva, line 144: fetch should still be async here. Call is the synchronous. If there isnt really async implementation then this should then just pass through to the __call__ function.

NickGeneva (changes requested):
Thanks Stefan!

The updates to the xarray dim and also the rename of the _fetch_wrapper seem good. But the removal of the async fetch has some consistency issues with the interface for the data sources.

swbg:
I have renamed `async def _fetch` to `async def fetch` to comply with the data source protocol. This way, `fetch` is now also parallel in structure to other data sources (e.g., validating times, setting up the xarray array, creating tasks).

NickGeneva:
@swbg thanks. I did some further reversions because this results in the async fetch API for the models breaking and a test failing.

swbg:
I guess we went full circle here. The point I did not realize is that fetch needs to be overwritten to make sure forecast data sources take lead_time as an argument and regular data sources do not.

Issue #721, opened 2026-02-27
🚀[FEA]: Add torch.compile support for autoregressive inference rollouts in run.py


abhaygoudannavar:
### Is this a new feature, an improvement, or a change to existing functionality?

Improvement

### How would you describe the priority of this feature request

Medium

## Problem

The inference workflows in [earth2studio/run.py](cci:7://file://wsl.localhost/Ubuntu/home/abhay/earth2studio/earth2studio/run.py:0:0-0:0) ([deterministic](cci:1://file://wsl.localhost/Ubuntu/home/abhay/earth2studio/earth2studio/run.py:38:0-151:13), [diagnostic](cci:1://file://wsl.localhost/Ubuntu/home/abhay/earth2studio/earth2studio/run.py:155:0-276:13), [ensemble](cci:1://file://wsl.localhost/Ubuntu/home/abhay/earth2studio/earth2studio/run.py:280:0-438:13)) execute autoregressive rollouts using plain Python iteration:

```python
# run.py, lines 140-149 (deterministic workflow)
for step, (x, coords) in enumerate(model):
    x, coords = map_coords(x, coords, output_coords)
    io.write(*split_coords(x, coords))
    pbar.update(1)
    if step == nsteps:
        break
```

Each rollout step calls the same model forward pass with an identical computation graph, but PyTorch re-traces the graph on every iteration. For models like FCN3, SFNO, Pangu, and DLWP — which repeat the exact same operations at every autoregressive step — this creates unnecessary overhead from:

- Python overhead per iteration (interpreter, dynamic dispatch)
- Redundant CUDA kernel launches without graph capture
- No cross-step kernel fusion opportunities
## Proposed Solution
Add optional torch.compile support to the prognostic model forward pass within the inference loop. torch.compile(mode="reduce-overhead") uses CUDA Graphs under the hood and can yield 1.5–3x speedups on autoregressive rollouts by:

- Capturing the computation graph once and replaying it
- Eliminating Python overhead per step
- Enabling cross-kernel optimizations via Triton
## Implementation Approach
Add a compile parameter to deterministic(), diagnostic(), and ensemble() workflows in run.py:
``` python
def deterministic(
    time, nsteps, prognostic, data, io,
    output_coords=OrderedDict({}),
    device=None,
    verbose=True,
    compile: bool = False,  # NEW
) -> IOBackend:
```
- Wrap the model iterator with torch.compile when enabled:
``` python
if compile:
[...]

PR #812, opened 2026-04-13
🚀[FEA]: Add torch.compile support for inference workflows

manmeet3591:
This PR addresses [Issue #721](https://github.com/NVIDIA/earth2studio/issues/721) by adding optional `torch.compile` support to the core inference workflows.

### **Problem**
Inference workflows in `earth2studio/run.py` currently execute autoregressive rollouts using plain Python iteration. For models that repeat the exact same operations at every step (e.g., FCN3, SFNO), this results in significant Python overhead and redundant CUDA kernel launches.

### **Solution**
An optional `compile` parameter has been added to the following workflows:
- `deterministic()`
- `diagnostic()`
- `ensemble()`

When `compile=True`, the prognostic model's `_forward` method is wrapped with `torch.compile(mode="reduce-overhead")`. In the diagnostic workflow, the diagnostic model is also compiled.

### **Key Benefits**
- **CUDA Graph Capture**: Automatically utilizes CUDA Graphs to replay the computation graph, eliminating per-step launch overhead.
- **Speedup**: Can yield 1.5–3x speedups for common prognostic models during long rollouts.
- **Minimal Changes**: Opt-in feature that preserves existing eager mode behavior by default.

### **Testing**
Added `test/run/test_compile.py` to verify that the `compile` flag correctly triggers `torch.compile` with the expected parameters across all three workflows using mocks.

### DIFF
--- a/earth2studio/run.py
+++ b/earth2studio/run.py
@@ -45,6 +45,7 @@ def deterministic(
     output_coords: CoordSystem = OrderedDict({}),
     device: torch.device | None = None,
     verbose: bool = True,
+    compile: bool = False,
 ) -> IOBackend:
     """Built in deterministic workflow.
     This workflow creates a determinstic inference pipeline to produce a forecast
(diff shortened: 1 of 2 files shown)


NickGeneva:
Thanks for putting this together! I'm closing this as it a bit out of date and also just isn't in our roadmap right now, most of our performance hot spots are IO based and I dont want to merge without a more formal performance study. Will need to revisit.

PR #815, opened 2026-04-14
Hunt for Tropical Cyclones pt 2: reproducibility

mariusaurus:
# Earth2Studio Pull Request

## Description
This PR adds the option to reproduce individual mini-batches generated with the tc_tracking recipe when running through FCN3. Advantage is that one can store just the tracks and re-run interesting members if fields are of interest.

## Checklist


mariusaurus:
@greptileai

NickGeneva:
@greptile-apps

recipes/tc_tracking/src/modes/generate_tc_hunt_ensembles.py
```diff
@@ -232,8 +239,7 @@ def run_inference(
 
         # set random state or apply perturbation
         if ("model" not in cfg) or (cfg.model == "fcn3"):
```
NickGeneva, line 241: When would "model" be None in the config? Can we force it to be a requirement?

recipes/tc_tracking/src/modes/generate_tc_hunt_ensembles.py
```diff
@@ -232,8 +239,7 @@ def run_inference(
...
-            if hasattr(model, "set_rng"):
-                model.set_rng(seed=seed)  # type: ignore[attr-defined]
+            model.set_rng(seed=seed)
         elif cfg.model.lower().startswith("aifs"):
```
NickGeneva, line 243: We can probably just remove this elif

recipes/tc_tracking/src/modes/generate_tc_hunt_ensembles.py
```diff
@@ -276,8 +282,7 @@ def run_inference(
     # Consolidate metadata in zarr files
     if dist.rank == 0 and cfg.store_type == "zarr" and store is not None:
```
NickGeneva, line 283: when would store be None?

recipes/tc_tracking/src/modes/generate_tc_hunt_ensembles.py
```diff
@@ -395,3 +400,97 @@ def generate_ensemble(cfg: DictConfig) -> None:
...
+    Returns
+    -------
+    tuple[list[tuple[np.datetime64, np.ndarray, int]] | None, list[np.datetime64]]
+        Tuple of (work items for this rank, list of all initial-condition times)
+    """
+    ic_mems = OmegaConf.to_container(cfg.reproduce_members)
+
+    ics = []
+    for ii in range(len(ic_mems)):
+        ic_mems[ii][0] = np.datetime64(ic_mems[ii][0])
```
NickGeneva, line 430: just one line work here. `ics.append(np.datetime64(ic_mems[ii][0]))`

recipes/tc_tracking/src/modes/generate_tc_hunt_ensembles.py
```diff
@@ -395,3 +400,97 @@ def generate_ensemble(cfg: DictConfig) -> None:
...
+        ics.append(ic_mems[ii][0])
+
+        # expand to the full batch that contains the requested member
+        batch_id = ic_mems[ii][1] // cfg.batch_size
+        ic_mems[ii][1] = np.arange(
+            batch_id * cfg.batch_size,
+            min((batch_id + 1) * cfg.batch_size, cfg.ensemble_size),
+        )
+
+    ic_mems = remove_duplicates(ic_mems)
```
NickGeneva, line 440: should this be done before the loop above? Also is this needed? I'm wondering if letting the user run the same date / time / seed is fine.

recipes/tc_tracking/src/modes/generate_tc_hunt_ensembles.py
```diff
@@ -395,3 +400,97 @@ def generate_ensemble(cfg: DictConfig) -> None:
...
+    ic_mems.sort(key=lambda x: x[0])
+    ics = list(set(ics))
+
+    if not DistributedManager().distributed:
+        return ic_mems, ics
+
+    ic_mems_rank = distribute_runs(ic_mems)
```
NickGeneva, line 447: why is the distribute needed here, I thought all processes would get the same config. Is this only called by the rank 0 process?

recipes/tc_tracking/src/tc_hunt_utils.py
```diff
@@ -60,6 +60,45 @@ def set_initial_times(cfg: DictConfig) -> np.ndarray:
...
+        datetime64 objects.
+
+    Returns
+    -------
+    list[list]
+        De-duplicated list preserving original order.
+    """
+
+    def to_hashable(item: object) -> object:
+        if isinstance(item, np.ndarray):
```
NickGeneva, line 84: when would this be a numpy array? I'm trying to figure out why this has a lot of checks for just

```
reproduce_members: [
                    ['2024-09-24 00:00:00',  0, 4280400703],
                    ['2024-09-24 00:00:00',  8, 1612911750],
                    ['2024-09-24 00:00:00',  9, 1612911750],
                    ['2024-09-24 00:00:00', 12, 1863392111],
                   ]
```

Are we not able to assume just [datetime, int] for all in reproduce_members?

recipes/tc_tracking/test/test_tc_hunt.sh
```diff
@@ -23,6 +24,50 @@ if [ "$track_count" -eq 0 ]; then
...
+
+# Extract seeds for members 3 and 4 from filenames
+seed_mem3=$(printf '%s\n' "${track_files[@]}" | grep 'mem_0003' | grep -oP 'seed_\K\d+')
+seed_mem4=$(printf '%s\n' "${track_files[@]}" | grep 'mem_0004' | grep -oP 'seed_\K\d+')
+
+# Check if seeds in reproduce_helene.yaml match, update if not
+yaml_seed3=$(grep -oP "\['2024-09-24 00:00:00', 3, \K\d+" cfg/reproduce_helene.yaml)
+yaml_seed4=$(grep -oP "\['2024-09-24 00:00:00', 4, \K\d+" cfg/reproduce_helene.yaml)
+
+if [ "$seed_mem3" != "$yaml_seed3" ] || [ "$seed_mem4" != "$yaml_seed4" ]; then
```
NickGeneva, line 42: Seems part of this test script has some capability to inplace edit the config file which isnt really the best?

NickGeneva: Because we want the repo's default state to be passing the tests, not after running the script

recipes/tc_tracking/test/test_tc_hunt.sh
```diff
@@ -23,6 +24,50 @@ if [ "$track_count" -eq 0 ]; then
...
+echo ""
+
+# Get track files for members 2, 3, 4
+mapfile -t track_files < <(ls "$baseline_dir"/tracks_*.csv | sort | grep -E 'mem_000[234]' | xargs -n1 basename)
+
+# Extract seeds for members 3 and 4 from filenames
+seed_mem3=$(printf '%s\n' "${track_files[@]}" | grep 'mem_0003' | grep -oP 'seed_\K\d+')
+seed_mem4=$(printf '%s\n' "${track_files[@]}" | grep 'mem_0004' | grep -oP 'seed_\K\d+')
+
+# Check if seeds in reproduce_helene.yaml match, update if not
```
NickGeneva, line 38: I'm wondering if all these checks are needed. I think its better to just assume that the configs are the same. The purpose of the test should be to check the configs produce same results anyway. I thinking this could be removed / simplified.

Maybe at most I could see this producing a warning if they miss match.

recipes/tc_tracking/pyproject.toml
```diff
@@ -22,6 +22,12 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
```
NickGeneva, line 24: cross verify the main readme is updated with the new dep groups.

recipes/tc_tracking/src/modes/generate_tc_hunt_ensembles.py
```diff
@@ -232,8 +239,7 @@ def run_inference(
 
         # set random state or apply perturbation
         if ("model" not in cfg) or (cfg.model == "fcn3"):
-            if hasattr(model, "set_rng"):
-                model.set_rng(seed=seed)  # type: ignore[attr-defined]
+            model.set_rng(seed=seed)
```
NickGeneva, line 242: Confirming that other random seeds are also set, like torch and nump somewhere for all models.

recipes/tc_tracking/src/modes/generate_tc_hunt_ensembles.py
mariusaurus: enforced

recipes/tc_tracking/src/modes/generate_tc_hunt_ensembles.py
mariusaurus: yep, removed both if/elif and just left if hasattr... left a comment for users why setting rng in aifs is not possible

recipes/tc_tracking/src/modes/generate_tc_hunt_ensembles.py
mariusaurus: store is the output for field data. so users have the option of just storing the tracks, which are orders of magnitude smaller in data compared to fields. the reproduce functionality's main intend here is to be able to reporduce fields of tracks that look interesting.

recipes/tc_tracking/src/modes/generate_tc_hunt_ensembles.py
mariusaurus: does it though? the first line converts the time from config (likely string) to to numpy datetime in ic_mems. the second line stores the numpy time in list of ICs.

recipes/tc_tracking/src/modes/generate_tc_hunt_ensembles.py
mariusaurus: the reasoning here is that always a full batch has to be reproduced, but for the user it is a lot more intuitive to define which ensemble member to reproduce. the above loop iterates over ensemble members specified by the user and reconstruct the full batches in which these members were present. once batches are picked for each member, duplicate batches are removed. 
so if a user ran a batch size of 2 and in the config defines to reporduce members 0 and 1, batch 0 gets added twice. as a result, the exact same config would run twice, overwriting it's own data. wouldn't be catastrophic, but waste of compute.

recipes/tc_tracking/src/modes/generate_tc_hunt_ensembles.py
mariusaurus: maybe bit unfortunate name. the method which assigns runs to ranks is deterministic, so each rank just does it in isolation, but all are guaranteed to have the same result. renamed to `assign_runs_to_rank`

recipes/tc_tracking/src/tc_hunt_utils.py
mariusaurus: the format is `[ic: np.datetime, member_indices: np.array(int), random_seed: int]`. have simplified the whole thing to 
```
        hashable_key = (
            sublist[0],
            (tuple(sublist[1].tolist()), str(sublist[1].dtype), sublist[1].shape),
            sublist[2],
        )
```
and removed the function with all the checks. hope that addresses it?

recipes/tc_tracking/pyproject.toml
mariusaurus: it is updated, also separated FCN3 and AIFS based on your comment in PR one review. [all] and docker file are updated accordingly

recipes/tc_tracking/src/modes/generate_tc_hunt_ensembles.py
mariusaurus: these internal random states are independent from np/random/torch seeds. however, np/random/torch seeds impact generation of rng seed in generate ensemble mode. in reproduce member mode this has no effect as rng seeds are provided through config and not generated by script.
was this the question or did I misunderstand?

recipes/tc_tracking/test/test_tc_hunt.sh
mariusaurus: the random seeds for fcn3 (rng) produced in `generate_ensemble` are set randomly. setting the numpy/torch/random seeds address this on a given platform but wouldn't trust that this is true when executed on a different platform. will implement that the random seed is overwritten through the bash command rather than editing the yaml file. See conversation above

recipes/tc_tracking/test/test_tc_hunt.sh
mariusaurus: true, let's overwrite them through the run command rather than editing the yaml file

recipes/tc_tracking/src/tc_hunt_utils.py
NickGeneva: Yep!

recipes/tc_tracking/test/test_tc_hunt.sh
NickGeneva: Great thanks!

recipes/tc_tracking/src/modes/generate_tc_hunt_ensembles.py
NickGeneva: Ah yeah youre correct. since icmems is used later.

NickGeneva (approved):
Thanks for addressing / responding to the revisions!

PR #885, opened 2026-05-28
Update install docs: fix cbottle hash and add ORBIT-2 diagnostic

NickGeneva:
- Fix cbottle git hash and repo URL to match pyproject.toml (NickGeneva/cBottle@e48c7eb)
- Add ORBIT-2 (climate-learn) to the diagnostics install section


ktangsali (approved):
LGTM. Are your forks a temporary workaround? I would imagine whatever the fix is, would be nice to push it to upstream and then revert?

NickGeneva:
Yes forks are the temporary until the fixes get upstreamed into the research repos, PRs are open but its slow to get reviews

PR #888, opened 2026-05-29
NNJA Conv Obs fixes

aayushg55:
# Earth2Studio Pull Request

## Description

This PR fixes NNJA conventional observation decoding so `NNJAObsConv` better matches the GSI/UFS Replay conventional-observation convention.
### 1. Filter NNJA `pres` to GSI `ps`-like observations
NNJA `prepbufr::POB` was previously treated as `pres` for every PREPBUFR level. That is too broad: `POB` is often just vertical-coordinate metadata for upper-air, wind, satellite-wind, or scatterometer reports.
This now matches GSI’s `obstype="ps"` behavior more closely:
- keep pressure-like report types: `120`, `180`, `181`, `187`
- require `level_cat == 0`
- reject `POB < 500 hPa`
- keep sane pressure quality marks
- convert retained pressure observations from hPa to Pa for Earth2Studio output
This is source-level normalization.
Reference:
https://github.com/NOAA-EMC/GSI/blob/860d13740352004fca0136a8c3d0ac9dea30e0da/src/gsi/read_prepbufr.F90#L1898-L1904
### 2. Complete the PREPBUFR message-family table
`PREPBUFR_OBS_TYPES` now follows the NCEP PREPBUFR Table 1.a message families more completely.
This fixes skipped merged-PREPBUFR message groups, notably:
- `103: AIRCAR`, which carries MDCRS/ACARS aircraft observations and restores missing U/V report type `233`
- `106: PROFLR`, which restores rare profiler/PILOT wind type `229`
- additional Table 1.a families for completeness/provenance
Reference:
https://www.emc.ncep.noaa.gov/mmb/data_processing/prepbufr.doc/table_1.htm
### 3. Fix `prepbufr.acft_profiles` path and type normalization
`prepbufr.acft_profiles` uses a special NNJA S3 layout under `.../bufr/...`, unlike the standard merged `prepbufr` layout. The path builder now handles that source correctly.
That source also uses aircraft-profile-specific report codes in the `33x/43x/53x` ranges. GSI remaps those back to standard aircraft report types when reading U/V from the aircraft-profile file:
- `330/430/530 -> 230`
- `331/431/531 -> 231`
- `332/432/532 -> 232`
- `333/433/533 -> 233`
- `334/434/534 -> 234`
- `335/435/535 -> 235`
It normalizes the special `prepbufr.acft_profiles` source to the same report-type convention as standard merged PREPBUFR and UFS/GSI diagnostics.
Reference:
https://github.com/NOAA-EMC/GSI/blob/860d13740352004fca0136a8c3d0ac9dea30e0da/src/gsi/read_prepbufr.F90#L730-L745
Note: for most downstream users, standard merged `prepbufr` should remain the default. `prepbufr.acft_profiles` is a special aircraft-profile product; use it only when that profile structure is specifically needed.

## Checklist


aayushg55:
image
For conv_uv, now NNJA is correctly a superset of UFS replay. We see 233 (imortant) and other minor obs type (229) that were missing before.

image

image

image

earth2studio/data/nnja.py
```diff
@@ -121,6 +121,27 @@
...
+    433: 233,
+    533: 233,
+    334: 234,
+    434: 234,
+    534: 234,
+    335: 235,
+    435: 235,
+    535: 235,
+}
+
```
albertocarpentieri, line 144: why remapping the type in the data source output? If the remapping is for HealDA only, i think it should go in the model in order to keep the data source as generic as possible. @NickGeneva

aayushg55: This remap is not HealDA-specific -- it follows GSI’s handling of the `prepbufr.acft_profiles` product (not the main prepbufr files). However, this remapping is only appropriate if `NNJAObsConv` is trying to expose GSI/PREPBUFR-compatible report types.

GSI treats these files specially:

- Standard prepbufr is the main merged conventional obs file.
- When aircraft_t_bc is enabled, GSI skips aircraft-profile-type observations from the normal prepbufr path.
- It then reads the aircraft profile file separately (acft_profl_file) and remaps the profile-specific 33x/43x/53x codes to standard aircraft report types (`23x`). You can think of the prepbufr as being the full homogenized set of obs, and the `prepbufr.acft_profiles` as being a more detailed version focused just on the aircraft data.

The tradeoff is that this collapses profile-stage information. If we want `NNJAObsConv(source="prepbufr.acft_profiles")` to preserve raw profile codes for users interested in the flight-level/ascending/descending distinction, then the better API would be to add another column.

aayushg55:
@greptile-apps

earth2studio/lexicon/nnja.py
```diff
@@ -113,6 +121,19 @@ def mod(df: pd.DataFrame) -> pd.DataFrame:
...
+                    obs_type = pd.to_numeric(df["type"], errors="coerce")
+                    level_cat = pd.to_numeric(df["level_cat"], errors="coerce")
+                    quality = pd.to_numeric(df.get("quality"), errors="coerce")
+                    pressure_obs = (
+                        obs.ge(_PRESSURE_MIN_HPA)
+                        & obs_type.isin(_PRESSURE_REPORT_TYPES)
+                        & df["class"].isin(_PRESSURE_CLASSES)
+                        & level_cat.eq(0)
+                        & (quality.isna() | ((quality >= 0) & (quality <= 15)))
+                    )
```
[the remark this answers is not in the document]
aayushg55: Good catch that the previous expression was not a meaningful “good quality” filter. The intent here is to match GSI’s read-stage validity guard, not to select actively assimilated observations. GSI rejects quality marks outside the valid PrepBUFR range with if(qm > 15 .or. qm < 0) cycle, so this has been changed to quality.between(0, 15) and the comment now clarifies that it admits valid PrepBUFR marks, including monitor/non-use marks, rather than filtering to used observations. A stricter assimilated-like filter would need to be a separate option because GSI use also depends on usage flags, thinning, and setup/gross-check logic.

earth2studio/data/nnja.py
albertocarpentieri: I would rather add more options, same for the pressure filtering which should probably be outside the datasource. I think the datasource should be open for other preprocessing / models as well.

aayushg55: We may want to add another raw_report_type field (for the unmapped aircraft metadata).

For the pressure obs, I am not sure it makes sense to keep the original semantics. Because pressure is metadata for all conv obs, we were creating fake pressure observations from things like AMV height assignments, aircraft/upper-air coordinates, scatterometer metadata, etc. I don’t think we want to reexpose the metadata of existing obs as a separate variable/obs stream. It is the equivalent of exposing height metadata as its own obs stream. I suppose we could have separate pressure and ps in the lexicon, but again the first is not always an actual obs. In contrast, the surface pressure obs are actual observations.

NickGeneva: @aayushg55 I'm still not following the remapping, here. Based on your comment this suggests that we are missing a needed column of sorts to then do further filtering that then follow GSI's process. The data source should really just report the data and do unit conversions for consistency across the package.

NickGeneva: > For the pressure obs, I am not sure it makes sense to keep the original semantics. Because pressure is metadata for all conv obs, we were creating fake pressure observations from things like AMV height assignments, aircraft/upper-air coordinates, scatterometer metadata, etc. 

I agree, the kinda fake pressure obs are maybe not good to expose if we dont need them. If it sufficient to have most obs just report pressure as a meta data and not the obs value itself thats  prefered imo

earth2studio/data/nnja.py
```diff
@@ -131,6 +152,7 @@
         E2STUDIO_SCHEMA.field("elev"),
         # NNJA stores PrepBUFR report-type code as uint16 (numeric)
         pa.field("type", pa.uint16(), nullable=True),
+        pa.field("level_cat", pa.uint16(), nullable=True),
```
NickGeneva, line 161: whats level_cat?

earth2studio/data/utils_bufr.py
```diff
@@ -109,18 +109,30 @@
     "VOB": OBS_VOB,
 }
 
-# PrepBUFR section-1 dataCategory -> NCEP message-type class string
-# Ref: NCEP PrepBUFR documentation and embedded DX Table A.
+# PrepBUFR section-1 dataCategory -> NCEP message/subset family.
+# Ref: NCEP PREPBUFR Table 1.a / prepobs_prep.bufrtable Table A.
+# https://www.emc.ncep.noaa.gov/mmb/data_processing/prepbufr.doc/table_1.htm
```
NickGeneva, line 114: thank you for the doc link

earth2studio/data/nnja.py
NickGeneva: Oh okay so the remaps are condensing multiple aircraft obs into a couple of bins.... I see. So if i switch back to prepbufr will all the original tags be preserved?

earth2studio/data/nnja.py
```diff
@@ -1006,7 +1037,25 @@ class NNJAObsConv(_NNJAObsBase):
     ----------
     source : {"prepbufr", "convbufr", "prepbufr.acft_profiles"}, optional
         Which encoding family of the NNJA conventional archive to read,
-        by default ``"prepbufr"``.
+        by default ``"prepbufr"``. These sources are different stages of the
+        NCEP observation-processing pipeline, not independent replacement
+        datasets:
+
+        - ``"convbufr"`` points at raw dump streams grouped by family, such as
```
NickGeneva, line 1050: is this one actually used?

earth2studio/data/nnja.py
NickGeneva: I think this link is wrong and needs to be: https://registry.opendata.aws/noaa-reanalyses-pds/

earth2studio/data/nnja.py
NickGeneva: Nvm i see it in the bucket

earth2studio/data/nnja.py
aayushg55: To clarify, all of this obs type remapping is only happening for this special "prepbufr.acft_profiles" source that is basically a subset of the data in the standard prepbufr files we would normally read from. TLDR: I was trying to find where the missing aircraft obs were, and tried looking at that special aircraft-only source, but originally that source wasn't actually supported to read from. The aircraft obs exist in these files, but come with a more detailed report code, which the GSI normalizes to the 2xx range obs type. But in regard to the missing obs, the actual issue was that we had a limited PREPBUFR_OBS_TYPES table, so we dropped many obs.

Now, if we want to support the "prepbufr.acft_profiles" source, the question becomes what do we put in the obs type column.

1. Put the unmapped report code as is in "type." Now the type column has GSI-like type values for the standard prepbufr but not for prepbufr.acft_profiles data. This is not inherently problematic, just overloading the type column. The user could, in theory, be careful and remap to GSI-like type values if desired.
2. Remap to the GSI codes (the code in this commit). Now the type column is more conistent but we have lost the original codes.
3. Add a new "raw_type" (or similarly named) column to designate "type" as being the GSI-like code and the other as being the raw internal type (not sure if any other source would benefit from such a designation or only these aircraft files).

earth2studio/data/nnja.py
aayushg55: Added a schema comment for this. `level_cat` is the decoded PrepBUFR `CAT` descriptor, i.e. the level category. We carry it because the `pres` modifier uses `CAT == 0` to identify station-level pressure rows rather than treating every POB pressure coordinate as a pressure observation.

earth2studio/data/nnja.py
aayushg55: Fixed, thanks. The NNJA docstring now points to https://registry.opendata.aws/noaa-reanalyses-pds/.

earth2studio/data/nnja.py
aayushg55: Based on this discussion, I kept the default `prepbufr.acft_profiles` output behavior mapped to the standard GSI/PREPBUFR aircraft `23x` report types in the `type` column, but added an internal toggle around that remap. It is not exposed as a public user option. This keeps the current default GSI-compatible output while leaving us a narrow internal path to preserve the raw aircraft-profile `33x`/`43x`/`53x` codes if we decide we need that later.

NickGeneva (approved):

PR #890, opened 2026-05-29
Fix batch_func validation for invalid output coords

### LINKED ISSUE #889: batch_func lets invalid output coords through and then crashes later
Version: current `main` snapshot

Installation method: source

I noticed a bug in `earth2studio/models/batch.py` while reviewing the batching path.

`batch_func._compress_batch` currently checks this:

```python
if (
    next(iter(input_coords)) != "batch"
    and next(iter(model.output_coords(input_coords))) != "batch"
):
    raise ValueError(
        "Model coordinate systems not compatible with batch processing"
    )
```

This only raises when both coordinate systems are wrong. If `input_coords` is valid but `output_coords` is not, the model still gets through the guard.

That later blows up in `_decompress_batch`, which does:

```python
del out_coords["batch"]
```

So instead of failing early with the compatibility error, it can fail later with a `KeyError: 'batch'`.\n\nExpected behavior:\n\nThe compatibility check should reject the model if either `input_coords` or `output_coords` does not start with `batch`.\n\nI think this should be enough to fix it:\n\n```python\nif (\n    next(iter(input_coords)) != "batch"\n    or next(iter(model.output_coords(input_coords))) != "batch"\n):\n```\n\nIt would also be good to add a regression test for the case where the output coords are missing `batch` so this fails cleanly and early.

fallintoplace:
Closes #889

## Description
- reject invalid `output_coords` earlier in `batch_func._compress_batch`
- add a regression test for models whose output coordinates do not start with `batch`

## Why
The existing guard only raised when both `input_coords` and `output_coords` were invalid. If `input_coords` was valid but `output_coords` was not, the call got past the compatibility check and later failed in `_decompress_batch` with `KeyError: 'batch'` instead of the intended early `ValueError`.\n\n## Impact\nThis makes the batching path fail fast with the compatibility error for misconfigured models and keeps the failure mode predictable.\n\n## Validation\n- `uv run pytest 'test/models/test_batch.py::test_invalid_output_coords_raise' 'test/models/test_batch.py::test_invalid_ordering_raises[cpu]'`\n- `uv run ruff check earth2studio/models/batch.py test/models/test_batch.py`

### DIFF
--- a/earth2studio/models/batch.py
+++ b/earth2studio/models/batch.py
@@ -101,7 +101,7 @@ class batch_func:
         input_coords = model.input_coords()
         if (
             next(iter(input_coords)) != "batch"
-            and next(iter(model.output_coords(input_coords))) != "batch"
+            or next(iter(model.output_coords(input_coords))) != "batch"
         ):
             raise ValueError(
                 "Model coordinate systems not compatible with batch processing"
--- a/test/models/test_batch.py
+++ b/test/models/test_batch.py
@@ -35,6 +35,7 @@ def PhooModel():
             return OrderedDict(
                 [
                     ("batch", np.empty(1)),
+                    ("lead_time", np.zeros(1)),
                     ("variable", np.array(["a", "b"])),
                     ("x", np.ones(1)),
                 ]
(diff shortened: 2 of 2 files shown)


NickGeneva (approved):
LGTM, nice spot. Thanks!

PR #908, opened 2026-06-05
Add offline mode to CDS data source

### LINKED ISSUE #905: 🚀[FEA]: Offline inference on CDS datasource

New Feature

Hi,
We're working on a new system that the computing nodes has no internet connection.
We tried to cache the CDS data on the login node and share the cache with computing nodes.
However, the datasource still tries to connect to copernicus.eu server, leading to https timeout failure.
This method works on an older system that the nodes have internet - we download the cache on the login node to save download time (and money!) on nodes.
So:
1. Is the CDS server connection must happen even when the cache already exists?
2. Is there some better way suggested to tackle this issue? (we're still trying to pack cache into Zarr offline input)

NickGeneva:
## Summary

Adds an `offline` parameter to the `CDS` data source class that enables inference on computing nodes without internet access by using pre-populated cache only.

## Motivation

Closes #905

Users working in air-gapped environments (e.g., HPC computing nodes) cannot connect to the CDS server. If they pre-populate the cache on a login node with internet access, they should be able to run inference on compute nodes without any network connectivity. Previously, the `CDS` class always created a `cdsapi.Client` during initialization, which requires valid credentials and could attempt network operations.

## Changes

- Added `offline: bool = False` parameter to `CDS.__init__`
- When `offline=True`:
  - The `cdsapi.Client` is **not** created (no credentials/network needed)
  - Cache hits return data normally (no behavior change)
  - Cache misses raise `FileNotFoundError` with a helpful message explaining how to populate the cache
  - `cache=False` is auto-corrected to `cache=True` with a warning (offline mode requires cache)
- Added runtime guard before `self.cds_client.retrieve()` to satisfy mypy type narrowing

## Usage

```python
from earth2studio.data import CDS

# On login node (with internet): populate cache
ds = CDS(cache=True)
data = ds(time, variables)  # Downloads and caches

# On compute node (no internet): use cached data
ds = CDS(offline=True)
data = ds(time, variables)  # Reads from cache only
```

## Tests

Added 5 new unit tests for offline mode:
- `test_cds_offline_init` - verifies no cdsapi.Client is created
- `test_cds_offline_cache_override` - verifies cache=False is corrected
- `test_cds_offline_cache_miss_raises` - verifies FileNotFoundError on miss
- `test_cds_offline_cache_hit` - verifies successful read from cache
- `test_cds_offline_multiple_variables_partial_miss` - verifies error on partial cache

All tests pass locally. All pre-commit hooks (black, ruff, mypy, etc.) pass.

### DIFF
--- a/earth2studio/data/cds.py
+++ b/earth2studio/data/cds.py
@@ -72,6 +72,11 @@ class CDS:
         Cache data source on local memory, by default True
     verbose : bool, optional
         Print download progress, by default True
+    offline : bool, optional
+        When True, only read from cache and never connect to the CDS server.
+        Useful for environments without internet access where the cache has been
+        pre-populated. Raises FileNotFoundError if requested data is not in cache.
+        By default False.
 
     Warning
     -------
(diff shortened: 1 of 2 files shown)


NickGeneva:
Closing in favor of a better approach: lazy client initialization instead of an explicit offline toggle.

PR #913, opened 2026-06-09
HealDA IO Improvement: UFS observation fetch via Obstore 

negin513:
# Earth2Studio Pull Request

## Description

This PR improves end-to-end HealDA performance by swapping the UFS observation fetch from s3fs to [obstore](https://github.com/developmentseed/obstore). This PR speeds up the object-store reads behind the UFS observation sources.

This PR targets the **S3 reads** portion of the UFS observation fetch (`UFSObsConv` / `UFSObsSat` --> the sources HealDA reads). `s3fs`/fsspec is GIL-bound and caps concurrent S3 reads at ~36 MB/s regardless of concurrency; obstore's native (Rust) async client lifts that ceiling, so the many small per-file GSI reads run with real parallelism.

**Changes:** `UFSObsConv` / `UFSObsSat` `_fetch_remote_file` now uses `obstore.get_range_async` / `get_async` against an anonymous, per-bucket-cached `S3Store` (`pool_max_idle_per_host = max_workers`). `_async_init` removed (lazy store creation). s3fs plumbing dropped from the UFS path.

image

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/earth2studio/blob/main/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.
- [x] The [CHANGELOG.md](https://github.com/NVIDIA/earth2studio/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/earth2studio/issues) is linked to this pull request.
- [x] Assess and address Greptile feedback (AI code review bot for guidance; use discretion, addressing all feedback is not required).

## Dependencies

Adds `obstore>=0.8` as a dependency. For now, s3fs/fsspec/aiohttp are still retained.

PR #914, opened 2026-06-09
HealDA IO Improvement: Parallel decode + Obstore

negin513:
# Earth2Studio Pull Request

## Description

This PR adds the second HealDA data-path improvement: parallelizing the NetCDF→DataFrame decode of the UFS GSI observation files. After #913, the remaining cost is the per-file HDF5→pandas **decode**, which is CPU- and GIL-bound and runs single-threaded.

`UFSObsConv` / `UFSObsSat` read hundreds of `diag_*.nc4` files per analysis. This PR:

- decodes files across a **persistent spawn-based process pool** (new `decode_workers` parameter, default `"auto"` = `min(16, CPUs)`; spawn is safe after CUDA init, unlike fork)
- decodes each file **once per (file, window) group** — `diag_conv_uv` serves both `u` and `v`, `diag_conv_gps` two retrieval types — instead of once per variable
- **assembles and time-masks the frames inside the workers**, so the per-row work (column casts, string broadcasts, boolean masking) parallelizes instead of serializing in the parent; broadcast columns travel as categoricals and are restored to object dtype after the final concat
- replaces the `pd.to_timedelta` hour conversion with a **bit-exact numpy path** (~19% of per-file decode)
- **falls back to serial decode** with an actionable warning if the workers cannot start (e.g. a calling script without an `if __name__ == "__main__":` guard); the pool is rebuilt on the next call, so transient worker crashes never permanently disable parallel decode

### Performance

Warm-cache 48 h HealDA fetch (2024-01-01T00:00, HealDA field subset), measured with the stage-attributed harness from the review thread. Apple-silicon laptop, so parallel numbers are conservative — the gain scales with worker count:

| | conv | sat |
|---|---|---|
| before (this PR as first pushed), serial | 2.43 s | 19.7 s |
| **now, serial** | **1.03 s (2.4x)** | **9.4 s (2.1x)** |
| before, auto workers (warm pool) | 0.49 s | 8.3 s |
| **now, auto workers (warm pool)** | **0.40 s** | **6.3 s** |

Correctness: outputs verified **bit-identical** (values, row order, dtypes) against the previous implementation on real S3 data for both sources; row counts bit-match the review benchmarks (43,268,516 sat / 6,179,731 conv).

### Known follow-ups (out of scope here)

- Spawn workers pay a full `earth2studio.data` import (~3.5 s each) on pool startup; moving the decode worker into a leaf module would cut 3–6 s off first fetch
- Merge request windows per cycle so overlapping requested times don't decode the same file twice
- Unify the spawn/fork + pool-lifetime policy with the `utils_ncep.py` decode pools (per review discussion)
- Overlap fetch and decode by submitting each file as it lands
- Half-open `(min, max]` window selection belongs with #967 / #1012

## Checklist


NickGeneva:
Temporary closing this, since the obs sources are going through some refactors will need to revisit. Can re-open when needed

negin513:
@aayushg55 and @albertocarpentieri for vis.

aayushg55:
@negin513 It looks like this currently exactly matches main with no changes? Is that inteneded?

aayushg55:
## The harness

Run it with `PYTHONPATH` pointing at the checkout you want to measure, so the same file can time
any revision:

```bash
# one NUMA node, so worker count and memory bandwidth stay comparable between runs
PYTHONPATH=/path/to/earth2studio numactl --cpunodebind=1 --membind=1 \
  python ufs_stage_bench.py --label pr914 --kind sat --fields healda \
  --decode-workers 16 --passes 2 --reap-pool
```

`--reap-pool` exists because of how the kernel accounts child CPU. `getrusage(RUSAGE_CHILDREN)`
reports CPU for children that have exited *and* been waited for; a worker that is still alive
contributes nothing. Your pool deliberately keeps its workers alive between fetches, so without
this flag the worker CPU column reads 0.00 s, which is how I first misread it. The flag calls
`shutdown(wait=True)` on the pool after each pass and clears the attribute so the next pass builds
a fresh one, which makes the workers exit, be reaped, and land in the accounting.

The side effect is that every pass then pays pool startup, so `--reap-pool` is what produced the
"first fetch" rows above and leaving it off produced the "later fetches" rows. My per-call pool
exits inside its own `with` block, so its workers are always reaped and the flag changes nothing
for it. `--profile` adds a cProfile of the parent process on the last pass.

<details>
<summary><code>ufs_stage_bench.py</code></summary>

```python
"""Stage-attributed fetch benchmark for any ``earth2studio.data.ufs`` revision.

Runs against main, PR #914 or another branch unchanged: the stage timers are
installed by wrapping functions in this process, so nothing in the library is
edited. Work done inside a decode worker is invisible to those wrappers, which is
the point — what they report is the work that stayed in the parent.

CPU seconds come from ``getrusage``: ``RUSAGE_CHILDREN`` counts what the worker
processes burned, including their interpreter startup, but only once they have
been reaped — a persistent pool reports zero until ``--reap-pool`` shuts it down.
"""

import argparse
import cProfile
import functools
import inspect
import io
import pstats
import resource
import time
from collections import defaultdict
from datetime import timedelta
from typing import Any, Callable

import numpy as np

import earth2studio.data.ufs as ufs_module
from earth2studio.data import UFSObsConv, UFSObsSat

CONV_VARIABLES = ["u", "v", "t", "q", "pres"]
SAT_VARIABLES = ["atms", "mhs", "amsua", "amsub", "iasi", "crisfsr", "airs"]

# What HealDA-v2 actually requests. Without these every field in the schema is
# read, which is a heavier workload than any model asks for.
CONV_FIELDS = ["time", "lat", "lon", "observation", "variable", "type", "elev", "pres"]
SAT_FIELDS = [
    "time",
    "lat",
    "lon",
    "observation",
    "variable",
    "sensor_index",
    "satellite",
    "scan_angle",
    "satellite_za",
    "solza",
]

STAGES: defaultdict[str, float] = defaultdict(float)
CALLS: defaultdict[str, int] = defaultdict(int)


def _wrap(fn: Callable[..., Any], stage: str) -> Callable[..., Any]:
    def wrapper(*args: Any, **kwargs: Any) -> Any:
        start = time.perf_counter()
        try:
            return fn(*args, **kwargs)
        finally:
            STAGES[stage] += time.perf_counter() - start
            CALLS[stage] += 1

    # Keep the wrapped name and module so a wrapper that lands on a module-level
    # decode entry still pickles by reference when a pool submits it
    return functools.wraps(fn)(wrapper)


def _patch_attr(owner: Any, name: str, stage: str) -> bool:
    """Time ``owner.name`` in this process, preserving staticmethod binding."""
    raw = (
        owner.__dict__.get(name)
        if inspect.isclass(owner)
        else getattr(owner, name, None)
    )
    if raw is None:
        return False
    if isinstance(raw, staticmethod):
        setattr(owner, name, staticmethod(_wrap(raw.__func__, stage)))
    else:
        setattr(owner, name, _wrap(raw, stage))
    return True


def _defining_class(cls: type, name: str) -> type | None:
    return next((base for base in cls.__mro__ if name in base.__dict__), None)


def install_timers(source_cls: type) -> list[str]:
    """Wrap the parent-side stages that exist on this revision."""
    installed = []
    compile_owner = _defining_class(source_cls, "_compile_dataframe")
    if compile_owner and _patch_attr(
        compile_owner, "_compile_dataframe", "compile_total"
    ):
        installed.append("compile_total")
    owner = _defining_class(source_cls, "_add_task_columns")
    if owner and _patch_attr(owner, "_add_task_columns", "add_task_columns"):
        installed.append("add_task_columns")
    # Decode entry point: named differently across revisions, and only visible
    # here when it runs in the parent (serial mode)
    for candidate in ("_decode_gsi_group", "_decode_gsi_file"):
        if _patch_attr(ufs_module, candidate, "decode_in_parent"):
            installed.append(f"decode_in_parent ({candidate})")
            break
    if _patch_attr(ufs_module.pd, "concat", "concat"):
        installed.append("concat")
    return installed


def build_source(
    kind: str, window_hours: int, decode_workers: str | None, fields: str
) -> Any:
    cls = UFSObsConv if kind == "conv" else UFSObsSat
    kwargs: dict[str, Any] = {}
    if (
        decode_workers is not None
        and "decode_workers" in inspect.signature(cls).parameters
    ):
        kwargs["decode_workers"] = (
            decode_workers if decode_workers == "auto" else int(decode_workers)
        )
    tolerance = (timedelta(hours=-(window_hours - 3)), timedelta(hours=3))
    return cls(time_tolerance=tolerance, verbose=False, **kwargs), kwargs


def cpu_seconds() -> tuple[float, float]:
    me = resource.getrusage(resource.RUSAGE_SELF)
    kids = resource.getrusage(resource.RUSAGE_CHILDREN)
    return (me.ru_utime + me.ru_stime, kids.ru_utime + kids.ru_stime)


def main() -> None:
    parser = argparse.ArgumentParser()
    parser.add_argument("--label", required=True)
    parser.add_argument("--kind", choices=["conv", "sat"], default="sat")
    parser.add_argument("--time", default="2024-01-01T00:00")
    parser.add_argument("--window-hours", type=int, default=48)
    parser.add_argument("--decode-workers", default=None)
    parser.add_argument(
        "--fields",
        choices=["healda", "all"],
        default="all",
        help="healda = only the columns HealDA-v2 requests; all = whole schema",
    )
    parser.add_argument("--passes", type=int, default=2)
    parser.add_argument("--profile", action="store_true", help="cProfile the parent")
    parser.add_argument(
        "--reap-pool",
        action="store_true",
        help="shut a persistent pool down after each pass, so getrusage sees "
        "the worker CPU; also makes every pass pay pool startup",
    )
    args = parser.parse_args()

    variables = CONV_VARIABLES if args.kind == "conv" else SAT_VARIABLES
    fields = None
    if args.fields == "healda":
        fields = np.array(CONV_FIELDS if args.kind == "conv" else SAT_FIELDS)
    source, kwargs = build_source(
        args.kind, args.window_hours, args.decode_workers, args.fields
    )
    installed = install_timers(type(source))
    print(f"[{args.label}] module {ufs_module.__file__}")
    print(f"[{args.label}] fields={args.fields}, workers={kwargs.get('decode_workers')}")
    print(f"[{args.label}] timers {installed}")

    analysis_time = np.array([np.datetime64(args.time)])
    for pass_index in range(args.passes):
        STAGES.clear()
        CALLS.clear()
        cpu_before = cpu_seconds()
        start = time.perf_counter()
        if args.profile and pass_index == args.passes - 1:
            profiler = cProfile.Profile()
            profiler.enable()
            df = source(analysis_time, variables, fields)
            profiler.disable()
        else:
            profiler = None
            df = source(analysis_time, variables, fields)
        wall = time.perf_counter() - start
        if args.reap_pool:
            pool = getattr(source, "_decode_pool", None)
            if pool is not None:
                pool.shutdown(wait=True)
                source._decode_pool = None
        cpu_after = cpu_seconds()

        compile_total = STAGES.get("compile_total", 0.0)
        attributed = sum(
            value for stage, value in STAGES.items() if stage != "compile_total"
        )
        print(
            f"[{args.label}] pass {pass_index}: wall {wall:6.2f} s  "
            f"compile {compile_total:6.2f} s  fetch+other {wall - compile_total:6.2f} s  "
            f"({len(df):,} rows)"
        )
        for stage in ("decode_in_parent", "add_task_columns", "concat"):
            if stage in STAGES:
                print(
                    f"    parent {stage:18s} {STAGES[stage]:6.2f} s  "
                    f"({CALLS[stage]} calls)"
                )
        print(
            f"    parent unattributed in compile "
            f"{compile_total - attributed:6.2f} s  (worker wait + IPC + frame build)"
        )
        print(
            f"    cpu: parent {cpu_after[0] - cpu_before[0]:6.2f} s  "
            f"children {cpu_after[1] - cpu_before[1]:6.2f} s  "
            f"total/wall {(sum(cpu_after) - sum(cpu_before)) / wall:5.2f}x",
            flush=True,
        )
        if profiler is not None:
            stream = io.StringIO()
            pstats.Stats(profiler, stream=stream).sort_stats("cumulative").print_stats(
                18
            )
            print("\n".join(stream.getvalue().splitlines()[4:30]), flush=True)
        del df


if __name__ == "__main__":
    main()
```

</details>

Sample output, so the columns above are readable — this is your branch at 16 workers on the
satellite leg with a fresh pool:

```
[pr914-w16] fields=all, workers=16
[pr914-w16] timers ['compile_total', 'add_task_columns', 'decode_in_parent (_decode_gsi_file)', 'concat']
[pr914-w16] pass 0: wall  16.71 s  compile  16.31 s  fetch+other   0.40 s  (43,268,516 rows)
    parent add_task_columns     0.33 s  (117 calls)
    parent concat               0.30 s  (1 calls)
    parent unattributed in compile  15.68 s  (worker wait + IPC + frame build)
    cpu: parent   6.32 s  children 173.04 s  total/wall 10.73x
```

A stage that reports zero calls is one that no longer runs in the parent, which is why the same
run against my branch prints only `concat`.

aayushg55:
## One thing to keep out of this PR

My branch also filters observations with a half-open window, `(min, max]` instead of
`[min, max]`, so that adjacent requests tile without returning the same observation twice. That
is a behaviour change and it belongs with #967 and #1012, not here — but it does affect the
numbers above, so flagging it for interpretation.

What changes concretely. Task planning floors the window's lower bound to a cycle label and then
steps forward in 6-hour cycles, so a closed `[min, max]` window over 48 hours admits **9** cycles
where a half-open `(min, max]` window admits **8** — the extra one is the cycle whose only
qualifying rows sit exactly on `min`. Every task in that cycle is a file that gets opened,
decoded in full, and then almost entirely discarded by the row filter:

| | closed `[min, max]` | half-open `(min, max]` |
|---|---|---|
| satellite tasks / files opened | 216 | 192 |
| conventional tasks / distinct files | 45 / 36 | 40 / 32 |
| satellite rows returned | 43,268,516 | 43,268,476 |
| conventional rows returned | 6,179,731 | 6,164,780 |

So it is 11% more files opened and fully decoded to gain 40 satellite rows and 14,951
conventional rows. Nothing is saved *within* a file — the per-file cost is identical — the whole
difference is 24 satellite files and 4 conventional files never being read at all. And the rows
it gains are the ones you do not want when windows tile: an observation exactly on a boundary is
returned by both the window that ends there and the window that starts there.

So of my 1.46x at one worker, about 1.30x is the cheaper per-file decode measured above and about
1.12x is simply not decoding that extra cycle. Only the 1.30x is on the table in this PR; the
rest belongs in #967.

@NickGeneva Would be nice to solidify the file selection

earth2studio/data/ufs.py
aayushg55: > `values = pd.to_timedelta(values, unit="h") + task.datetime_file`

This one line is 19% of a file's decode time, and replacing it is bit-exact.

I ran both branches' decode functions over the same 12 cached files in-process, carrying your output through the parent-side steps so both end at the same finished frame — identical 5,059,154 rows confirms the same files:

| | per file | of which `pd.to_timedelta` |
|---|---|---|
| this PR | 522.1 ms | 97.8 ms, 18.7% |
| mine | 408.9 ms | — |

113.2 ms per file separates the two branches and 97.8 ms of it is this line, so it's most of the per-file gap. Over the 104 files of a serial satellite fetch that's about 10 s.

```python
_NS_PER_HOUR = 3_600_000_000_000


def _hours_since_to_datetime(values: np.ndarray, origin: datetime) -> np.ndarray:
    """Convert fractional hours since ``origin`` to ``datetime64[ns]``.

    Bit-exact with ``pd.to_timedelta(values, unit="h") + origin``; the rounding
    to 12 decimal places is what reproduces the pandas result.
    """
    hours = values.astype(np.float64, copy=False)
    whole_hours = hours.astype(np.int64)
    fractional_ns = (np.round(hours - whole_hours, 12) * _NS_PER_HOUR).astype(np.int64)
    offset_ns = whole_hours * _NS_PER_HOUR + fractional_ns
    return np.datetime64(origin, "ns") + offset_ns.astype("timedelta64[ns]")
```

Splitting whole hours from the fraction and rounding to 12 places is what makes it exact. The obvious one-liner `(values * 3.6e12).astype("int64")` differs from pandas by up to 2 ns on 144,151 of 200,012 test values, which matters only because the row filter can flip a boundary observation. The split version I checked against pandas on every value including boundaries: 0 ns difference everywhere.

earth2studio/data/ufs.py
```diff
@@ -225,51 +334,50 @@ def _compile_dataframe(
...
+            for (task, local_path, column_map), future in zip(jobs, futures):
+                try:
+                    decoded.append(future.result())
+                except Exception as exc:  # pragma: no cover - defensive
+                    logger.error("Failed to read {}: {}", local_path, exc)
+                    raise exc
 
+        frames: list[pd.DataFrame] = []
+        for (task, local_path, column_map), data in zip(jobs, decoded):
+            df = pd.DataFrame(data)
```
aayushg55: > `df = pd.DataFrame(data)`

This loop is the reason the pool stops paying off. Workers return `dict[str, pa.Array]` and then the parent runs `pd.DataFrame(data)`, the rename, `_add_task_columns` and the time mask once per task, serially: **4.4 s of parent CPU** on the satellite leg. Two symptoms follow from it.

The warm-pool wall time bottoms out near that floor — 9.1 s per fetch with the pool left alive, against 6.4 s of parent CPU. And adding workers past 16 makes it *worse*, because they contend for a parent that can't keep up: **16.3 s at 16 workers, 17.7 s at 32, 20.0 s at 48**.

If the worker returns the finished frame instead of the column dict, parent CPU drops to about 2 s and wall time follows. That keeps everything else about your design, including Arrow across the boundary for the columns themselves.

One thing worth knowing before optimizing this: I replayed these steps in isolation and they only cost about 1.5 s, so `pd.DataFrame` on a dict of Arrow arrays is *not* the problem — it's cheap and returns numpy dtypes without copying. The expensive part is one line further down; see the next comment.

earth2studio/data/ufs.py
aayushg55: > `df["satellite"] = task.satellite`

This is the single most expensive step in the parent-side assembly. It broadcasts one string into a ~370k-row object column, once per file, 104 times for a satellite fetch. It's most of the 4.4 s in the previous comment, and it's why the isolated replay of `pd.DataFrame` + rename + mask came to only 1.5 s.

Moving frame assembly into the worker fixes this along with everything else. If you'd rather not move assembly, this line alone is worth attention — a dictionary-encoded or categorical column, or attaching it after the concat rather than per task.

earth2studio/data/ufs.py
```diff
@@ -58,6 +62,75 @@ class _GSIAsyncTask:
...
+    This is the per-process unit of work for parallel decode: it must stay a
+    module-level function taking only picklable arguments (in particular,
+    ``task`` must not carry a lexicon closure), and it returns Arrow arrays,
+    which serialize between processes near zero-copy.
+    """
+    with h5netcdf.File(local_path, "r") as ds:
+        data: dict[str, pa.Array] = {}
+        channel_index_raw: np.ndarray | None = None
+        for name, dset in ds.variables.items():
+            if name not in column_map:
```
aayushg55: > `if name not in column_map:`

Related but separate: each *task* decodes a file, and several tasks can want the same file, so some files are opened and fully read more than once. `diag_conv_uv` holds both u and v, and `diag_conv_gps` holds two retrieval types, so on this request it's **45 conventional tasks over 36 distinct files** — 9 redundant full decodes.

Grouping tasks by `(file, window)` and fanning out the observation column inside the worker gets those back. The columns other than `observation` are identical across a group, so it's one read and a per-task column selection.

earth2studio/data/ufs.py
```diff
@@ -195,13 +271,46 @@ def _handle_missing_file(self, key: str) -> None:
...
+
+        Uses the 'spawn' start method: safe after CUDA init and available on
+        all platforms. Workers are spawned on demand up to max_workers, so
+        sizing the pool to the configured limit (rather than this call's file
+        count) costs nothing up front.
+        """
+        if self._decode_pool is None:
+            pool = concurrent.futures.ProcessPoolExecutor(
+                max_workers=self._decode_worker_limit(),
+                mp_context=multiprocessing.get_context("spawn"),
```
aayushg55, line 351: > `mp_context=multiprocessing.get_context("spawn"),`

Your choice is the defensible one and mine is the weaker one on paper — fork after CUDA is initialised is undefined, and my pool forks. I don't think it should be settled inside this PR though, because there are now three decode paths in the package answering the same question differently:

| | `gdas.py` / `nnja.py` via `utils_ncep.py` | this PR | mine |
|---|---|---|---|
| knob | `decode_workers: int = 8` | `int \| "auto" = "auto"`, capped at `min(16, cpus)` | `decode_workers: int = 8` |
| pool lifetime | per call, in a `with` block | persistent per instance | per call, in a `with` block |
| start method | platform default, so fork on Linux | spawn | fork |
| per-worker setup | `initializer` loads BUFR tables | pickled with each task | `initializer` stashes the schema |

The hazard you're avoiding is already in the tree: a run builds the model on the GPU and then fetches observations, so the four `ProcessPoolExecutor` pools in `utils_ncep.py` — including the IR sounder decode — fork with CUDA live today. Fixing it in one source leaves it in the other two while paying the cost here.

And the cost is mostly one import, not the start method. 16 workers, one task each so every worker really starts:

| pool | until all 16 are up | CPU summed over the 16 |
|---|---|---|
| spawn, workers import `earth2studio.data.ufs` | 6.4 s | 115 s |
| spawn, workers import only h5netcdf/numpy/pandas/pyarrow | 0.4 s | 14 s |
| fork | 0.05 s | ~0 s |

`earth2studio.data.ufs` takes about 5 s to import because it pulls in torch, and with spawn every worker pays it. So the fix worth having is making the worker's import cheap — move the decode body into a module that doesn't import torch, which today means not pickling the source class into the worker. Then spawn costs 0.4 s per pool instead of 6.4 s, and your persistent pool makes even that one-time.

You can read what that buys off the summary table: a pool whose startup is nearly free behaves like one that's already warm, so your first fetch lands near your "later fetches" row — about 4.6 s instead of 19.6 s on the HealDA field set.

That split also has a benefit independent of start method. My fork pool is safe today only because nothing in the worker touches CUDA and nothing in it logs, and nothing enforces that. If the decode body lives in a module that *cannot* import torch, that becomes a property you can test instead of a convention to remember.

Suggestion: one shared helper owning start method, worker count and the parallel threshold, used by both `utils_ncep.py` and `ufs.py`, rather than deciding it per source.

earth2studio/data/ufs.py
```diff
@@ -58,6 +62,75 @@ class _GSIAsyncTask:
     satellite: str | None = None
 
 
+# Cap for automatic decode-worker selection. Decode throughput plateaus once it
+# drops below the network floor (~16 workers in practice), and more workers than
+# files is wasted anyway (the pool spawns processes on demand).
+_DECODE_WORKERS_CAP = 16
```
aayushg55, line 71: > `_DECODE_WORKERS_CAP = 16`

The cap is the right number, but not for the reason the comment gives. It says throughput plateaus once decode drops below the network floor; what I measured is that past 16 it doesn't plateau, it **regresses** — 16.3 s at 16 workers, 17.7 s at 32, 20.0 s at 48 — and the mechanism is likely related to the serial parent-side assembly in comment 2, not the network. 
Note: my comment  and L65-68 would likely be deprecated after we improve perf through the suggested changes

aayushg55 (changes requested):
Thanks @negin513. I had a branch of my own going based on the same profiling, so I benchmarked your branch vs `main` vs [mine](https://github.com/aayushg55/earth2studio/tree/ufs-perf) to see where they agree and where they don't. Short version: we can get some more perf with some tweaks. I've left them as inline comments. Happy to merge in my changes if that is helpful.

**How I measured.** One script that runs unmodified against any revision of `earth2studio/data/ufs.py` — it installs its timers by wrapping functions in the calling process, so no library file is edited. A 48-hour window at `2024-01-01T00:00`, conventional plus all seven satellite sensors: 6.2M conventional rows, 43.3M satellite rows, 252 files. Page cache warm, runs interleaved between revisions so none gets a colder cache, pinned to one NUMA node (70 cores). Every number is the mean of two passes that agreed within 4%. Harness posted below.

Your pool is created on first use and kept on the instance, so a **first fetch** pays to start 16 workers and have each import `earth2studio.data.ufs`, while **later fetches** through the same source object reuse live workers. My pool is per-call, so one number covers both.

On the field set HealDA actually requests:

| | conv | sat | total |
|---|---|---|---|
| `main`, serial | 4.02 s | 28.22 s | 32.24 s |
| this PR, 16 workers, first fetch | 8.06 s | 11.56 s | 19.62 s |
| this PR, 16 workers, later fetches | 0.46 s | 4.10 s | 4.56 s |
| this PR, 8 workers, later fetches | 0.69 s | 5.57 s | 6.26 s |
| mine, 16 workers | 0.31 s | 2.29 s | 2.60 s |
| mine, 8 workers | 0.35 s | 2.82 s | 3.17 s |

Requesting the whole schema instead, which is the case that separates the two designs most:

| | conv | sat | total |
|---|---|---|---|
| `main` | 10.21 s | 53.04 s | 63.25 s |
| this PR, 1 worker | 10.22 s | 52.52 s | 62.74 s |
| this PR, auto (16 workers) | 8.98 s | 16.34 s | 25.32 s |
| mine, 1 worker | 5.52 s | 36.19 s | 41.71 s |
| mine, 16 workers | 0.59 s | 3.98 s | 4.56 s |

More columns means a bigger per-task frame, and the frame is assembled in the parent, so reading more than you need punishes this structure specifically.

Where the time goes at 16 workers. Parent CPU is the floor on wall time since it can't be parallelised away; worker CPU is summed across all 16, so it's the cost to the machine rather than to the clock:

| leg | | wall | parent CPU | worker CPU |
|---|---|---|---|---|
| sat | this PR | 16.7 s | 6.4 s | 174 s |
| sat | mine | 4.1 s | 2.1 s | 38 s |
| conv | this PR | 9.4 s | 1.2 s | 134 s |
| conv | mine | 0.6 s | 0.3 s | 6.5 s |

The three things I'd change, each its own thread below: the per-file decode `pd.to_timedelta` (19% of a file), the frame assembly happens serially in the parent (which is why more workers eventually makes it slower), and conventional files are decoded once per variable rather than once per file. The spawn vs fork choice for the multiprocess pool also has some nuance and whichever we lean towards should apply for all the obs sources.

And fyi there's also an open cycle-selection bug (in all of our sat/conv data sources) that pays an extra 11% of wasted decode.

earth2studio/data/ufs.py
aayushg55: Sorry about this one — my original comment suggested a categorical column as one option for the `satellite`/`variable` broadcast. Turns out it doesn't help anymore, now that assembly moved into the worker.

Here's why: a categorical is cheap to *pickle* back from a worker, which mattered when the broadcast column was being built in the parent. But now each file gets its own tiny categorical (`categories=["metop-a"]` for one file, `categories=["n16"]` for the next, etc). `pd.concat` can only merge categoricals cheaply when they all share the *same* categories — with 117 different ones, it falls back to materializing every row as a real Python string instead. That fallback is now the single biggest cost in the whole fetch: about 5.3s of the 7s warm-pool sat leg, all in one `pd.concat` call in the parent.

The fix is simpler than the thing it's replacing: just assign the plain string again, same as before assembly moved into the worker.

```python
# _decode_gsi_group, per-task loop
df["variable"] = task.e2s_obs_name          # was: _repeated_string_column(...)

# UFSObsSat._add_task_columns
if task.satellite is not None:
    df["satellite"] = task.satellite        # was: _repeated_string_column(...)
```

`_repeated_string_column` and the post-concat `.astype(object)` restore loop can both go.

Tested this against the 48h HealDA sat fetch, warm pool: 7.16s → 2.17s, same 43,268,516 rows. The pickle savings the categorical was for turned out to be tiny next to everything else already crossing the pipe with each frame, so it wasn't worth what it cost at concat time.

aayushg55 (approved):
Thanks! Just one small change and should be good to go.

negin513:
/okay to test [649ca42](https://github.com/NVIDIA/earth2studio/pull/914/commits/649ca42995fb35d82db5f3c7c3465b6e28d96289)

negin513:
Addressed in https://github.com/NVIDIA/earth2studio/pull/914

PR #925, opened 2026-06-17
Add DLESyM-v0-ISCCP-ERA5 model

pzharrington:
# Earth2Studio Pull Request

## Description
Adds the DLESyM-v0-ISCCP-ERA5 model

## Checklist


pzharrington:
@greptileai

pzharrington:
Validation:

1. Comparison against reference pipeline -- RMSE between the two grows at expected rates and saturates/asymptotes by ~30 days across variables. Chaotic dynamics in atmosphere amplify tiny numerical differences and leads to decorrelation over the span of roughly a month; SST is much slower to drift due to slower dynamics and more smoothing in the model. Both saturate below sqrt(2) climatology reference due to model smoothing. Not able to be bit-identical for two reasons listed below: 
   a. Inference-side bug in the upstream pipeline. The atmos SST coupling in `coupled_forecast.py` reads SST at the start of the input history window (-18h) rather than the integration/initial time (0h). This contradicts both Cresswell-Clay et al. 2024 (which specifies the initial SST S_0, held constant through the rollout) and the `ConstantCoupler`'s own docstring ("coupled field at integration time"). Avoiding propagating the bug from upstream reference is a better choice.
   b. float32 accumulation drift caused by the insolation implementation difference (`physicsnemo` vs upstream `insolation` agree only to ~3e-7). Sticking with `phyiscsnemo` implementation is more consistent w/ other models and more maintainable.
image

2. Long rollout stability -- model runs stably to 100 years without producing any values at any intermediate step which exceed the bounds of what is physical/expected by climatology.
image

Reference scripts:
<details>
<summary>Comparison to reference pipeline (click to expand)</summary>

```python
"""
sanity_check_dlesym_parity.py
=============================================================
Multi-coupled-step parity check: DLESyMv0_ISCCP_ERA5 (E2S) vs the upstream
``scripts/coupled_forecast.py``.  Both are initialized from the same
``ref_dlesym/example_data`` zarrs at 2017-01-01T00, using native HEALPix and
``use_ttr=False`` so the TTR transform and lat/lon regridding are off the
critical path.  Defaults to a 60-day rollout so the per-variable RMSE can be
seen to asymptote toward climatology.  Output is:

  <out-dir>/rmse_vs_leadtime.png           — normalized RMSE (RMSE/σ_ref) vs lead
  <out-dir>/spatial_<var>_lt<N>h.png       — 3-panel (E2S | ref | diff)
  <out-dir>/precip_tp06_ic.png             — tp06 map from DLESyMv0_ISCCP_ERA5Precip

Pass criteria (printed by ``_evaluate_parity``):
  (a) first coupled step (≤96h): per-variable nRMSE < ``--step1-tol``
  (b) final lead time: |Δmean|/σ_ref and |Δstd|/σ_ref < ``--dist-tol``

Note on first-step parity: the production wrapper couples the atmosphere on SST
at integration time (0h), while the upstream ConstantCoupler reads SST at the
atmos input-window start on its first step.  This makes the first coupled step's
atmos predictions differ by a small (physically tiny) amount.  Pass
``--match-ref-coupling`` to emulate the upstream quirk and recover a
bit-identical first step.

Usage::

    DLESYM_V0_ISCCP_ERA5_PACKAGE_PATH=/path/to/pkg \\
    python sanity_check_dlesym_parity.py \\
        [--upstream-dir ref_dlesym] \\
        [--days 60 | --steps N] \\
        [--match-ref-coupling] \\
        [--device cuda:0] \\
        [--out-dir parity_out]

To skip re-running the reference (already have the NetCDF files)::

    python sanity_check_dlesym_parity.py \\
        --ref-atmos parity_out/ref_atmos.nc \\
        --ref-ocean parity_out/ref_ocean.nc

Prerequisites
-------------
  - ``DLESYM_V0_ISCCP_ERA5_PACKAGE_PATH`` set to the built model package.
  - ``ref_dlesym/`` cloned and ``git lfs pull`` run inside it.
  - Earth2Studio with the ``dlesym`` extra installed (``uv run python`` or
    ``.venv/bin/python``).
"""

import argparse
import math
import os
import subprocess
import sys
from pathlib import Path

import numpy as np
import torch
import xarray as xr
from omegaconf import OmegaConf

try:
    import cartopy.crs as ccrs

    _HAS_CARTOPY = True
except Exception:  # pragma: no cover - cartopy is optional for plotting
    _HAS_CARTOPY = False

# --------------------------------------------------------------------------- #
# Constants
# --------------------------------------------------------------------------- #
# Upstream channel name → Earth2Studio variable name.
_VAR_RENAME = {"t2m0": "t2m", "tcwv0": "tcwv", "ws10": "ws10m", "olr": "rlut"}
_VAR_RENAME_INV = {v: k for k, v in _VAR_RENAME.items()}
INIT_TIME = np.datetime64("2017-01-01T00:00")


def _e2s_to_up(name: str) -> str:
    return _VAR_RENAME_INV.get(name, name)


# --------------------------------------------------------------------------- #
# Package / config loading
# --------------------------------------------------------------------------- #
def _load_pkg():
    from earth2studio.models.auto import Package

    path = os.environ.get("DLESYM_V0_ISCCP_ERA5_PACKAGE_PATH")
    if not path:
        raise SystemExit("DLESYM_V0_ISCCP_ERA5_PACKAGE_PATH is not set")
    pkg = Package(path, cache_options={"same_names": True})
    cfg = OmegaConf.load(pkg.resolve("config.yaml"))
    return pkg, cfg


# --------------------------------------------------------------------------- #
# IC construction from example zarrs
# --------------------------------------------------------------------------- #
def _read_example_state(
    example_dir: Path, cfg, times: np.ndarray
) -> tuple[torch.Tensor, list[str]]:
    """Return (state (T, V, 12, ns, ns), list of E2S variable names)."""
    atmos_vars = list(cfg.io.atmos_variables)
    ocean_vars = list(cfg.io.ocean_variables)
    variables = atmos_vars + ocean_vars

    ds_a = xr.open_zarr(example_dir / "hpx64_9varCoupledAtmos-sst.zarr")
    ds_o = xr.open_zarr(example_dir / "hpx64_1varCoupledOcean-z1000-ws10-olr.zarr")

    def _pull(ds: xr.Dataset, var: str) -> np.ndarray:
        up = _e2s_to_up(var)
        for arr, cdim in [("inputs", "channel_in"), ("targets", "channel_out")]:
            if arr in ds and cdim in ds[arr].dims:
                chans = [str(c) for c in ds[cdim].values]
                if up in chans:
                    da = ds[arr].isel({cdim: chans.index(up)}).sel(
                        time=times, method="nearest"
                    )
                    return np.asarray(da.values, dtype=np.float32)
        raise KeyError(f"{var!r} (upstream {up!r}) not found in dataset")

    fields = []
    for v in variables:
        ds_primary = ds_a if v in atmos_vars else ds_o
        try:
            fields.append(_pull(ds_primary, v))
        except KeyError:
            ds_fallback = ds_o if ds_primary is ds_a else ds_a
            fields.append(_pull(ds_fallback, v))

    state = np.stack(fields, axis=1)  # (T, V, 12, ns, ns)
    return torch.from_numpy(state), variables


def _patch_coupling_to_window_start(model) -> None:
    """Force the FIRST step's atmos SST coupling to the input-window start.

    The upstream ``ConstantCoupler`` reads SST at the *start* of the atmos input
    window (earliest atmos input time, e.g. -18h) rather than the integration
    time (0h) on the first coupled step (``j==0``, non-coupled mode) -- a
    reference quirk.  Our shipped wrapper correctly uses 0h, so the first step's
    atmos predictions differ slightly from the reference unless we emulate the
    quirk.  This patch applies the window-start index ONLY on the first
    invocation of ``_make_atmos_coupling`` (restoring the normal index right
    after), so steps 2+ still couple on the fresh ocean output.  Test instrument
    only; never touches the production wrapper.
    """
    in_coords = model.input_coords()
    leads = list(in_coords["lead_time"])
    start_t = model.atmos_input_times[0]  # earliest atmos input time, e.g. -18h
    start_idx = leads.index(start_t)
    orig_idx = list(model.atmos_coupled_input_lt_idx)
    n = len(orig_idx)
    orig_make = model._make_atmos_coupling
    state = {"first": True}

    def _patched_make(x, coords):
        if state["first"]:
            state["first"] = False
            model.atmos_coupled_input_lt_idx = [start_idx] * n
            try:
                return orig_make(x, coords)
            finally:
                model.atmos_coupled_input_lt_idx = orig_idx
        return orig_make(x, coords)

    model._make_atmos_coupling = _patched_make
    print(
        f"[patch] atmos SST coupling forced to window-start lead "
        f"{int(np.asarray(start_t).astype('timedelta64[h]').astype(int))}h "
        f"for the FIRST step only (reference-matching; test only)"
    )


def _build_ic(
    pkg, cfg, example_dir: Path, device: str, match_ref_coupling: bool = False
) -> tuple:
    from earth2studio.models.px import DLESyMv0_ISCCP_ERA5

    model = DLESyMv0_ISCCP_ERA5.load_model(pkg, use_ttr=False).to(device)
    if match_ref_coupling:
        _patch_coupling_to_window_start(model)
    in_coords = model.input_coords()
    full_lts = in_coords["lead_time"]
    state, variables = _read_example_state(example_dir, cfg, INIT_TIME + full_lts)
    x = state[None, None].to(device)  # (1, 1, LT, V, 12, ns, ns)
    coords = OrderedDict(
        {
            "batch": np.array([0]),
            "time": np.array([INIT_TIME]),
            "lead_time": full_lts,
            "variable": np.array(variables),
            "face": np.arange(12),
            "height": np.arange(int(cfg.data.nside)),
            "width": np.arange(int(cfg.data.nside)),
        }
    )
    return model, x, coords, variables


# --------------------------------------------------------------------------- #
# E2S rollout
# --------------------------------------------------------------------------- #
def _run_e2s(model, x, coords, steps: int) -> dict:
    """Run E2S for ``steps`` coupled steps.

    Returns ``{(valid_time, variable): field(12, ns, ns)}`` in physical units.
    """
    out: dict = {}
    it = model.create_iterator(x, coords)

    # IC yield: capture lead_time=0 slice
    ic, ic_c = next(it)
    zero = np.timedelta64(0, "h")
    ic_leads = list(ic_c["lead_time"])
    if zero in ic_leads:
        li0 = ic_leads.index(zero)
        t0 = np.datetime64(ic_c["time"][0])
        for vi, v in enumerate(list(ic_c["variable"])):
            out[(t0, v)] = ic[0, 0, li0, vi].cpu().numpy()

    for step in range(steps):
        y, yc = next(it)
        vlist = list(yc["variable"])
        lt_last = yc["lead_time"][-1]
        print(
            f"[e2s] step {step + 1:2d}  lead[-1]={lt_last.astype('timedelta64[h]')}"
        )
        for li, lt in enumerate(yc["lead_time"]):
            vt = np.datetime64(yc["time"][0]) + lt
            for vi, v in enumerate(vlist):
                f = y[0, 0, li, vi]
                if not torch.isfinite(f).all():
                    print(f"  !! non-finite: {v} at {vt}")
                out[(vt, v)] = f.cpu().numpy()
    return out


# --------------------------------------------------------------------------- #
# Upstream reference (subprocess)
# --------------------------------------------------------------------------- #
def _run_reference(
    upstream_dir: Path, out_dir: Path, steps: int, device: str
) -> tuple[Path, Path]:
    out_dir.mkdir(parents=True, exist_ok=True)
    gpu = "-1" if device == "cpu" else device.split(":")[-1]
    lead_hours = steps * 96
    cmd = [
        sys.executable,
        "scripts/coupled_forecast.py",
        "--atmos-model-path", "models/dlwp",
        "--ocean-model-path", "models/dlom",
        "--lead-time", str(lead_hours),
        "--forecast-init-start", str(INIT_TIME.astype("datetime64[D]")),
        "--forecast-init-end", str(INIT_TIME.astype("datetime64[D]")),
        "--freq", "D",
        "--data-directory", "example_data",
        "--atmos-dataset-name", "hpx64_9varCoupledAtmos-sst",
        "--ocean-dataset-name", "hpx64_1varCoupledOcean-z1000-ws10-olr",
        "--output-directory", str(out_dir),
        "--atmos-output-filename", "ref_atmos",
        "--ocean-output-filename", "ref_ocean",
        "--gpu", gpu,
    ]
    env = dict(os.environ)
    env["PYTHONPATH"] = str(upstream_dir) + os.pathsep + env.get("PYTHONPATH", "")
    print(f"[reference] running upstream coupled_forecast.py ...")
    subprocess.run(cmd, cwd=upstream_dir, check=True, env=env)
    return out_dir / "ref_atmos.nc", out_dir / "ref_ocean.nc"


def _load_reference(atmos_nc: Path, ocean_nc: Path) -> dict:
    out: dict = {}
    for nc in (atmos_nc, ocean_nc):
        ds = xr.open_dataset(nc)
        init = np.datetime64(ds["time"].values[0])
        for var in ds.data_vars:
            da = ds[var]
            if "step" not in da.dims:
                continue
            e2s = _VAR_RENAME.get(var, var)
            for si, step in enumerate(da["step"].values):
                lt_h = np.timedelta64(int(step), "ns").astype("timedelta64[h]")
                vt = np.datetime64(init + lt_h)
                out[(vt, e2s)] = np.asarray(
                    da.isel(time=0, step=si).values, dtype=np.float32
                )
    return out


# --------------------------------------------------------------------------- #
# Plotting helpers
# --------------------------------------------------------------------------- #
def _build_ll_regridder(nside: int, device: str):
    """Build an ``earth2grid`` HEALPix→lat/lon regridder (same grid as the model).

    Returns ``(regrid, lat, lon)`` where ``regrid`` maps a flat HEALPix field
    ``(npix,)`` to an equiangular ``(721, 1440)`` lat/lon grid, ``lat`` runs
    90→-90 and ``lon`` runs 0→360.
    """
    import earth2grid

    hpx = earth2grid.healpix.Grid(
        level=int(np.log2(nside)),
        pixel_order=earth2grid.healpix.HEALPIX_PAD_XY,
    )
    ll = earth2grid.latlon.equiangular_lat_lon_grid(721, 1440)
    regrid = earth2grid.get_regridder(hpx, ll).to(torch.float32).to(device)
    lat = np.linspace(90, -90, 721, endpoint=True)
    lon = np.linspace(0, 360, 1440, endpoint=False)
    return regrid, lat, lon


def _hpx_to_ll(regrid, field: np.ndarray, device: str) -> np.ndarray:
    """Regrid one HEALPix field ``(12, ns, ns)`` to lat/lon ``(721, 1440)``."""
    t = torch.as_tensor(np.asarray(field), dtype=torch.float32, device=device).reshape(
        1, -1
    )
    with torch.no_grad():
        out = regrid(t).reshape(721, 1440)
    return out.detach().cpu().numpy()


def _make_map_axes(nrows: int, ncols: int, figsize: tuple):
    """Create a figure/axes grid, using a PlateCarree projection if cartopy is
    available so coastlines can be overlaid."""
    if _HAS_CARTOPY:
        return plt.subplots(
            nrows,
            ncols,
            figsize=figsize,
            subplot_kw={"projection": ccrs.PlateCarree()},
        )
    return plt.subplots(nrows, ncols, figsize=figsize)


def _draw_map(ax, data, lat, lon, vmin, vmax, cmap):
    """Draw ``data`` (721, 1440) on ``ax`` with coastlines when available."""
    if _HAS_CARTOPY:
        im = ax.pcolormesh(
            lon,
            lat,
            data,
            transform=ccrs.PlateCarree(),
            vmin=vmin,
            vmax=vmax,
            cmap=cmap,
            shading="auto",
        )
        ax.coastlines(linewidth=0.4, color="k")
        ax.set_global()
    else:
        im = ax.imshow(
            data,
            extent=[0, 360, -90, 90],
            origin="upper",
            aspect="auto",
            vmin=vmin,
            vmax=vmax,
            cmap=cmap,
            interpolation="nearest",
        )
    return im


def _plot_comparison(
    ours: dict,
    ref: dict,
    regrid,
    lat: np.ndarray,
    lon: np.ndarray,
    device: str,
    out_dir: Path,
) -> None:
    out_dir.mkdir(parents=True, exist_ok=True)

    # ── normalized RMSE vs lead time ─────────────────────────────────────────
    # Normalize each variable's RMSE by the reference field's spatial σ at that
    # lead time so all variables share one axis and the asymptote toward
    # climatology is visible: two independent draws from the same distribution
    # give RMSE → √2·σ, i.e. nRMSE → √2 ≈ 1.41.
    by_var: dict[str, list] = {}
    for (vt, var), fo in ours.items():
        if (vt, var) in ref:
            by_var.setdefault(var, []).append((vt, fo, ref[(vt, var)]))

    fig, ax = plt.subplots(figsize=(10, 5))
    for var, entries in sorted(by_var.items()):
        entries = sorted(entries)
        lts = [
            int((vt - INIT_TIME).astype("timedelta64[h]").astype(int))
            for vt, _, _ in entries
        ]
        nrmses = [
            float(
                np.sqrt(((fo.astype(np.float64) - fr.astype(np.float64)) ** 2).mean())
            )
            / max(float(fr.astype(np.float64).std()), 1e-12)
            for _, fo, fr in entries
        ]
        ax.plot(lts, nrmses, marker="o", ms=3, label=var)
    ax.axhline(
        np.sqrt(2), color="k", ls="--", lw=1, label="√2 (decorrelated)"
    )
    ax.set_xlabel("Lead time (h)")
    ax.set_ylabel("Normalized RMSE (RMSE / σ_ref)")
    ax.set_title(
        f"DLESyMv0_ISCCP_ERA5 vs upstream coupled_forecast  init={INIT_TIME}"
    )
    ax.legend(fontsize=7, ncol=2)
    ax.grid(True, alpha=0.3)
    fig.tight_layout()
    p = out_dir / "rmse_vs_leadtime.png"
    fig.savefig(p, dpi=150)
    plt.close(fig)
    print(f"[plot] {p}")

    # ── spatial diff at last lead time per variable ──────────────────────────
    all_vts = sorted({vt for vt, _ in ours})
    if not all_vts:
        return
    last_vt = all_vts[-1]
    lt_h = int((last_vt - INIT_TIME).astype("timedelta64[h]").astype(int))

    for var in sorted({v for _, v in ours if (last_vt, v) in ref}):
        fo = ours[(last_vt, var)].astype(np.float64)
        fr = ref[(last_vt, var)].astype(np.float64)
        diff = fo - fr

        go = _hpx_to_ll(regrid, fo, device)
        gr = _hpx_to_ll(regrid, fr, device)
        gd = _hpx_to_ll(regrid, diff, device)

        both = np.stack([go, gr])
        vmin = float(np.nanpercentile(both, 2))
        vmax = float(np.nanpercentile(both, 98))
        dabs = max(float(np.nanpercentile(np.abs(gd), 99)), 1e-12)

        fig, axes = _make_map_axes(1, 3, (16, 4))
        for ax, data, title, cmap, lo, hi in (
            (axes[0], go, "E2S", "viridis", vmin, vmax),
            (axes[1], gr, "reference", "viridis", vmin, vmax),
            (axes[2], gd, "diff (E2S – ref)", "RdBu_r", -dabs, dabs),
        ):
            im = _draw_map(ax, data, lat, lon, lo, hi, cmap)
            ax.set_title(title)
            fig.colorbar(im, ax=ax, shrink=0.6)

        rmse = float(np.sqrt(((fo - fr) ** 2).mean()))
        ref_std = float(fr.std())
        nrmse = rmse / max(ref_std, 1e-12)
        # The diff colorbar is auto-scaled to ±dabs (99th pct of |diff|), so a
        # negligible residual still fills the colormap.  nRMSE (RMSE as a % of
        # the field's own spatial σ) is the scale-free read on how big the diff
        # really is relative to the signal.  At long lead the forecasts have
        # decorrelated toward climatology, so nRMSE → ~141 % (√2) is expected.
        fig.suptitle(
            f"{var}  lead={lt_h}h  RMSE={rmse:.3e}  max|d|={np.abs(diff).max():.3e}  "
            f"nRMSE={100 * nrmse:.1f}% of field σ  "
            f"[diff cbar autoscaled to ±{dabs:.2e}]",
            fontsize=9,
        )
        fig.tight_layout()
        p = out_dir / f"spatial_{var}_lt{lt_h:04d}h.png"
        fig.savefig(p, dpi=120)
        plt.close(fig)
        print(f"[plot] {p}")


# --------------------------------------------------------------------------- #
# Precip diagnostic
# --------------------------------------------------------------------------- #
def _run_precip(
    pkg, cfg, example_dir: Path, device: str, out_dir: Path, regrid, lat, lon
) -> None:
    """Run the precip diagnostic on the IC state (t=-ptimes[0], t=0) and plot."""
    from earth2studio.models.dx import DLESyMv0_ISCCP_ERA5Precip

    precip = DLESyMv0_ISCCP_ERA5Precip.load_model(pkg, use_ttr=False).to(device)
    pvars = list(precip.input_coords()["variable"])
    ptimes = precip.input_times  # e.g. [timedelta64(-6,'h'), timedelta64(0,'h')]

    ic_times = INIT_TIME + ptimes
    ic_state, ic_vars = _read_example_state(example_dir, cfg, ic_times)
    order = [ic_vars.index(v) for v in pvars]
    xp = ic_state[:, order][None, None].to(device)  # (1, 1, LT, Vp, 12, ns, ns)

    p_coords = OrderedDict(
        {
            "batch": np.array([0]),
            "time": np.array([INIT_TIME]),
            "lead_time": ptimes,
            "variable": np.array(pvars),
            "face": np.arange(12),
            "height": np.arange(int(cfg.data.nside)),
            "width": np.arange(int(cfg.data.nside)),
        }
    )

    tp, _ = precip(xp, p_coords)
    tp_np = tp[0, 0, 0, 0].cpu().numpy()  # (12, ns, ns)
    tp_flat = tp_np.ravel()

    tp_min = float(tp_flat.min())
    tp_mean = float(tp_flat.mean())
    tp_max = float(tp_flat.max())
    n_neg = int((tp_flat < -1e-4).sum())
    n_extreme = int((tp_flat > 0.1).sum())  # > 100 mm / 6h

    print(f"\n[precip] tp06:  min={tp_min:+.3e}  mean={tp_mean:+.3e}  max={tp_max:+.3e}  m/6h")
    print(f"         negative pixels: {n_neg}  |  >100 mm/6h pixels: {n_extreme}")
    if n_extreme:
        print(f"  [WARNING] {n_extreme} pixels > 100 mm/6h — check scaling")
    elif n_neg > 100:
        print(f"  [WARNING] {n_neg} pixels with negative precip")
    else:
        print("  [OK] tp06 range looks physical")

    # Spatial plot
    grid = _hpx_to_ll(regrid, tp_np, device)
    vmax = max(float(np.nanpercentile(np.abs(grid), 99.5)), 1e-6)

    fig, ax = _make_map_axes(1, 1, (9, 4))
    im = _draw_map(ax, grid, lat, lon, 0, vmax, "Blues")
    ax.set_title(
        f"DLESyMv0_ISCCP_ERA5Precip — tp06 at IC ({INIT_TIME})\n"
        f"min={tp_min:.3e}  mean={tp_mean:.3e}  max={tp_max:.3e}  m/6h"
    )
    fig.colorbar(im, ax=ax, label="tp06 (m / 6 h)", shrink=0.7)
    fig.tight_layout()
    p = out_dir / "precip_tp06_ic.png"
    fig.savefig(p, dpi=120)
    plt.close(fig)
    print(f"[plot] {p}")


# --------------------------------------------------------------------------- #
# Parity evaluation
# --------------------------------------------------------------------------- #
def _evaluate_parity(
    ours: dict, ref: dict, step1_tol: float, dist_tol: float
) -> bool:
    """Two-part sanity criterion appropriate for a long coupled rollout.

    (a) **First coupled step** (valid times within +96h): the deterministic
        models should agree closely this early, so require per-variable
        ``nRMSE = RMSE / σ_ref < step1_tol``.  Pattern correlation is reported
        alongside.  (With ``--match-ref-coupling`` the first step is bit-exact,
        so nRMSE → 0.)

    (b) **Final lead time**: by 60 days the forecasts have decorrelated toward
        climatology, so a pointwise RMSE is meaningless.  Instead require the
        E2S and reference fields to occupy the same physical distribution:
        ``|Δmean| / σ_ref`` and ``|Δstd| / σ_ref`` both below ``dist_tol``.
    """
    from collections import defaultdict

    common = set(ours) & set(ref)
    if not common:
        print("[parity] no overlapping keys — variable name mismatch?")
        print(f"  ours sample:  {sorted(ours)[:3]}")
        print(f"  ref  sample:  {sorted(ref)[:3]}")
        return False

    # ── (a) first coupled step ───────────────────────────────────────────────
    step1_cut = INIT_TIME + np.timedelta64(96, "h")
    s1: dict[str, list] = defaultdict(list)
    for (vt, var), fo in ours.items():
        if (vt, var) in ref and INIT_TIME < vt <= step1_cut:
            s1[var].append((fo, ref[(vt, var)]))

    print(
        f"\n[parity:a] first coupled step (≤96h) — close agreement  "
        f"(tol: nRMSE < {step1_tol:.0%})"
    )
    print(f"{'var':12s} {'nRMSE':>10s} {'corr':>9s} {'pass':>6s}")
    print("-" * 41)
    pass_a = True
    for var in sorted(s1):
        fo = np.stack([a for a, _ in s1[var]]).astype(np.float64)
        fr = np.stack([b for _, b in s1[var]]).astype(np.float64)
        rmse = float(np.sqrt(((fo - fr) ** 2).mean()))
        sd = float(fr.std())
        nrmse = rmse / max(sd, 1e-12)
        corr = float(np.corrcoef(fo.ravel(), fr.ravel())[0, 1])
        ok = nrmse < step1_tol
        pass_a = pass_a and ok
        print(f"{var:12s} {nrmse:9.2%} {corr:9.5f} {'OK' if ok else 'FAIL':>6s}")

    # ── (b) final-state distribution ─────────────────────────────────────────
    last_vt = max(vt for vt, _ in ours)
    lt_h = int((last_vt - INIT_TIME) / np.timedelta64(1, "h"))
    print(
        f"\n[parity:b] final lead {lt_h}h — distribution match "
        f"(decorrelated → climatology; tol: |Δμ|/σ, |Δσ|/σ < {dist_tol:.0%})"
    )
    hdr = (
        f"{'var':12s} {'mean_e2s':>11s} {'mean_ref':>11s} {'std_e2s':>11s} "
        f"{'std_ref':>11s} {'Δμ/σ':>8s} {'Δσ/σ':>8s} {'pass':>6s}"
    )
    print(hdr)
    print("-" * len(hdr))
    pass_b = True
    final_vars = sorted({v for (vt, v) in ours if vt == last_vt and (last_vt, v) in ref})
    for var in final_vars:
        fo = ours[(last_vt, var)].astype(np.float64)
        fr = ref[(last_vt, var)].astype(np.float64)
        mo, mr = float(fo.mean()), float(fr.mean())
        so, sr = float(fo.std()), float(fr.std())
        dmean = abs(mo - mr) / max(sr, 1e-12)
        dstd = abs(so - sr) / max(sr, 1e-12)
        ok = dmean < dist_tol and dstd < dist_tol
        pass_b = pass_b and ok
        print(
            f"{var:12s} {mo:11.3e} {mr:11.3e} {so:11.3e} {sr:11.3e} "
            f"{dmean:7.2%} {dstd:7.2%} {'OK' if ok else 'FAIL':>6s}"
        )
    print("-" * len(hdr))

    print(
        f"[parity] (a) first-step agreement: {'PASS' if pass_a else 'FAIL'}   "
        f"(b) final distribution: {'PASS' if pass_b else 'FAIL'}"
    )
    return pass_a and pass_b


# --------------------------------------------------------------------------- #
# Main
# --------------------------------------------------------------------------- #
def main() -> int:
    p = argparse.ArgumentParser(description="DLESyMv0_ISCCP_ERA5 parity sanity check")
    p.add_argument("--upstream-dir", default="ref_dlesym")
    p.add_argument(
        "--days",
        type=int,
        default=60,
        help="forecast length in days (default 60; RMSE should asymptote to climatology)",
    )
    p.add_argument(
        "--steps",
        type=int,
        default=None,
        help="coupled steps (96 h each); overrides --days when set",
    )
    p.add_argument("--device", default="cuda:0")
    p.add_argument("--out-dir", default="parity_out")
    p.add_argument(
        "--step1-tol",
        type=float,
        default=0.05,
        help=(
            "Pass criterion (a): max per-variable nRMSE (RMSE/σ_ref) over the "
            "first coupled step (≤96h). Default 0.05 (5%% of field σ)."
        ),
    )
    p.add_argument(
        "--dist-tol",
        type=float,
        default=0.10,
        help=(
            "Pass criterion (b): max |Δmean|/σ_ref and |Δstd|/σ_ref at the final "
            "lead time. Default 0.10 (10%% of field σ)."
        ),
    )
    p.add_argument(
        "--match-ref-coupling",
        action="store_true",
        help=(
            "Emulate the upstream first-step SST-coupling quirk (window-start "
            "lead instead of 0h) so the first coupled step is bit-identical to "
            "the reference. Off by default: the production wrapper uses 0h, "
            "which is the source of the small nonzero first-step RMSE."
        ),
    )
    p.add_argument("--ref-atmos", default=None, help="pre-computed reference atmos.nc")
    p.add_argument("--ref-ocean", default=None, help="pre-computed reference ocean.nc")
    args = p.parse_args()

    steps = args.steps if args.steps is not None else math.ceil(args.days * 24 / 96)

    device = args.device
    if device.startswith("cuda") and not torch.cuda.is_available():
        print(f"[warn] {device} not available, falling back to cpu")
        device = "cpu"

    upstream_dir = Path(args.upstream_dir).resolve()
    example_dir = upstream_dir / "example_data"
    # Resolve to absolute so the reference subprocess (cwd=upstream_dir) writes
    # to the right location when --output-directory is passed as a relative path.
    out_dir = Path(args.out_dir).resolve()
    out_dir.mkdir(parents=True, exist_ok=True)
    sys.path.insert(0, str(upstream_dir))

    pkg, cfg = _load_pkg()
    model, x, coords, _ = _build_ic(
        pkg, cfg, example_dir, device, match_ref_coupling=args.match_ref_coupling
    )
    regrid, lat, lon = _build_ll_regridder(int(cfg.data.nside), device)
    if not _HAS_CARTOPY:
        print("[warn] cartopy not available — maps drawn without coastlines")

    print(
        f"\n[e2s] running {steps} coupled step(s) "
        f"(~{steps * 96 / 24:.0f} days) from {INIT_TIME} ..."
    )
    ours = _run_e2s(model, x, coords, steps)

    if args.ref_atmos and args.ref_ocean:
        atmos_nc, ocean_nc = Path(args.ref_atmos), Path(args.ref_ocean)
        print(f"\n[reference] loading from {atmos_nc}, {ocean_nc}")
    else:
        atmos_nc, ocean_nc = _run_reference(upstream_dir, out_dir, steps, device)
    ref = _load_reference(atmos_nc, ocean_nc)

    ok = _evaluate_parity(ours, ref, args.step1_tol, args.dist_tol)
    _plot_comparison(ours, ref, regrid, lat, lon, device, out_dir)
    _run_precip(pkg, cfg, example_dir, device, out_dir, regrid, lat, lon)

    print(f"\n[done] outputs → {out_dir.resolve()}")
    return 0 if ok else 1


if __name__ == "__main__":
    sys.exit(main())
```

</details>
  
<details>
<summary>Long rollout test (click to expand)</summary>

```python
"""
sanity_check_dlesym_longrollout.py
====================================
100-year long-rollout stability check for the **pure Earth2Studio**
DLESyMv0_ISCCP_ERA5 pipeline.

This exercises the end-to-end Earth2Studio implementation: the lat/lon model
variant ``DLESyMv0_ISCCP_ERA5LatLon`` is initialized from an Earth2Studio data
source (ARCO ERA5) with ``use_ttr=True`` (the default), so the ERA5 ``ttr``
input, the TTR→OLR transform, the HEALPix regridding round-trip and the coupled
atmosphere/ocean rollout are all on the critical path.  The goal is to show the
long-rollout behavior of the pure-E2S model is stable over a multi-decade run.

It runs the model for ``--years`` years (default 100, ~9125 coupled steps of
96 h each) and checks every output step for:

  - Non-finite values (immediate HARD flag, model output is broken).
  - Hard violations: any pixel exceeds ±``SIGMA_HARD`` σ of the variable's
    training-climatological mean (from the package config.yaml scaling stats).
    These would represent clearly unphysical blow-ups.
  - Warn violations: any pixel exceeds ±``SIGMA_WARN`` σ.  These appear during
    normal S2S drift; they are logged but do not count as failures.

At the end (or on ``--plot-every N`` if requested), a panel of global maps for
all variables is saved.  The precip diagnostic is also run on the first coupled
output to confirm it produces physical tp06 values.

Usage::

    DLESYM_V0_ISCCP_ERA5_PACKAGE_PATH=/path/to/pkg \\
    python sanity_check_dlesym_longrollout.py \\
        [--years 100] \\
        [--init 2021-06-15] \\
        [--device cuda:0] \\
        [--out-dir longrollout_out] \\
        [--plot-every 0] \\
        [--no-ttr]

The script is designed for unattended overnight runs: it prints a progress line
every ~1 % of steps and writes the final-step plot when done.

Typical wall-clock time on a single A100: ~3–5 h for 100 years.

Prerequisites
-------------
  - ``DLESYM_V0_ISCCP_ERA5_PACKAGE_PATH`` set to the built model package.
  - Network access to the ARCO ERA5 store for the initial condition.
  - Earth2Studio with the ``dlesym`` extra installed.
"""

import argparse
import math
import os
import time
from pathlib import Path

import numpy as np
import torch
from omegaconf import OmegaConf

try:
    import cartopy.crs as ccrs

    _HAS_CARTOPY = True
except Exception:  # pragma: no cover - cartopy is optional for plotting
    _HAS_CARTOPY = False

# --------------------------------------------------------------------------- #
# Constants
# --------------------------------------------------------------------------- #
# Upstream channel name → Earth2Studio variable name (config scaling keys may
# use either namespace; this maps any upstream names to E2S names).
_VAR_RENAME = {"t2m0": "t2m", "tcwv0": "tcwv", "ws10": "ws10m", "olr": "rlut"}

SIGMA_WARN = 10   # per-pixel |z-score| > this triggers a warning entry
SIGMA_HARD = 15   # per-pixel |z-score| > this is a hard violation


# --------------------------------------------------------------------------- #
# Package / config
# --------------------------------------------------------------------------- #
def _load_pkg():
    from earth2studio.models.auto import Package

    path = os.environ.get("DLESYM_V0_ISCCP_ERA5_PACKAGE_PATH")
    if not path:
        raise SystemExit("DLESYM_V0_ISCCP_ERA5_PACKAGE_PATH is not set")
    pkg = Package(path, cache_options={"same_names": True})
    cfg = OmegaConf.load(pkg.resolve("config.yaml"))
    return pkg, cfg


# --------------------------------------------------------------------------- #
# Physical bounds from training scaling stats
# --------------------------------------------------------------------------- #
def _build_bounds(cfg) -> dict[str, tuple[float, float, float, float]]:
    """Return {e2s_var: (warn_lo, warn_hi, hard_lo, hard_hi)} from config stats."""
    bounds: dict[str, tuple[float, float, float, float]] = {}
    for v_up, stats in cfg.data.scaling.items():
        v = _VAR_RENAME.get(v_up, v_up)
        m = float(stats["mean"])
        s = float(stats["std"])
        bounds[v] = (
            m - SIGMA_WARN * s,
            m + SIGMA_WARN * s,
            m - SIGMA_HARD * s,
            m + SIGMA_HARD * s,
        )
    return bounds


# --------------------------------------------------------------------------- #
# Per-step physicality check
# --------------------------------------------------------------------------- #
def _check_step(
    y: torch.Tensor,
    y_coords: dict,
    bounds: dict,
    step_num: int,
    violations: list[str],
    hard_violations: list[str],
) -> None:
    """Check the last-lead-time slice of y for out-of-bounds or non-finite values.

    ``y`` is on the lat/lon grid with layout (batch, time, lead_time, variable,
    lat, lon)."""
    vlist = list(y_coords["variable"])
    field = y[0, 0, -1]  # (V, lat, lon) — last lead time (current valid time)
    vt = np.datetime64(y_coords["time"][0]) + y_coords["lead_time"][-1]

    for vi, var in enumerate(vlist):
        f = field[vi]

        # Non-finite check
        if not torch.isfinite(f).all():
            n_bad = int((~torch.isfinite(f)).sum())
            msg = (
                f"step {step_num:6d} | {str(vt):22s} | {var:12s} | "
                f"NON-FINITE ({n_bad}/{f.numel()} pixels)"
            )
            print(f"  !! {msg}")
            hard_violations.append(msg)
            continue

        if var not in bounds:
            continue

        fmin = float(f.min())
        fmax = float(f.max())
        wlo, whi, hlo, hhi = bounds[var]

        if fmin < hlo or fmax > hhi:
            msg = (
                f"step {step_num:6d} | {str(vt):22s} | {var:12s} | "
                f"HARD  range=[{fmin:+.3e}, {fmax:+.3e}]  "
                f"hard_bounds=[{hlo:+.3e}, {hhi:+.3e}]"
            )
            print(f"  !! {msg}")
            hard_violations.append(msg)
        elif fmin < wlo or fmax > whi:
            msg = (
                f"step {step_num:6d} | {str(vt):22s} | {var:12s} | "
                f"warn  range=[{fmin:+.3e}, {fmax:+.3e}]  "
                f"warn_bounds=[{wlo:+.3e}, {whi:+.3e}]"
            )
            violations.append(msg)


# --------------------------------------------------------------------------- #
# Plotting
# --------------------------------------------------------------------------- #
def _make_map_axes(nrows: int, ncols: int, figsize: tuple):
    """Create a figure/axes grid, using a PlateCarree projection if cartopy is
    available so coastlines can be overlaid."""
    if _HAS_CARTOPY:
        return plt.subplots(
            nrows,
            ncols,
            figsize=figsize,
            subplot_kw={"projection": ccrs.PlateCarree()},
        )
    return plt.subplots(nrows, ncols, figsize=figsize)


def _draw_map(ax, data, lat, lon, vmin, vmax, cmap):
    """Draw ``data`` (lat, lon) on ``ax`` with coastlines when available."""
    if _HAS_CARTOPY:
        im = ax.pcolormesh(
            lon,
            lat,
            data,
            transform=ccrs.PlateCarree(),
            vmin=vmin,
            vmax=vmax,
            cmap=cmap,
            shading="auto",
        )
        ax.coastlines(linewidth=0.4, color="k")
        ax.set_global()
    else:
        im = ax.imshow(
            data,
            extent=[float(lon.min()), float(lon.max()), -90, 90],
            origin="upper",
            aspect="auto",
            vmin=vmin,
            vmax=vmax,
            cmap=cmap,
            interpolation="nearest",
        )
    return im


def _plot_fields(
    y: torch.Tensor,
    y_coords: dict,
    out_path: Path,
    title: str,
    hard_violations: list[str] | None = None,
) -> None:
    """Panel of global maps (one per variable) for the last lead time of y."""
    vlist = list(y_coords["variable"])
    lat = np.asarray(y_coords["lat"])
    lon = np.asarray(y_coords["lon"])
    n = len(vlist)
    ncols = min(4, n)
    nrows = math.ceil(n / ncols)

    fig, axes = _make_map_axes(nrows, ncols, (5 * ncols, 3 * nrows))
    axes_flat = np.array(axes).flatten()

    field = y[0, 0, -1].cpu().numpy()  # (V, lat, lon)
    for vi, var in enumerate(vlist):
        f = field[vi]  # (lat, lon)
        lo = float(np.nanpercentile(f, 2))
        hi = float(np.nanpercentile(f, 98))
        if lo == hi:
            hi = lo + 1.0
        im = _draw_map(axes_flat[vi], f, lat, lon, lo, hi, "viridis")
        axes_flat[vi].set_title(var, fontsize=8)
        fig.colorbar(im, ax=axes_flat[vi], shrink=0.6)

    for ax in axes_flat[n:]:
        ax.axis("off")

    hard_flag = ""
    if hard_violations:
        hard_flag = f"  !! {len(hard_violations)} HARD violation(s) !!"
    fig.suptitle(title + hard_flag, fontsize=10)
    fig.tight_layout()
    fig.savefig(out_path, dpi=100)
    plt.close(fig)
    print(f"[plot] {out_path}")


# --------------------------------------------------------------------------- #
# Precip diagnostic sanity check (run once on the first coupled output)
# --------------------------------------------------------------------------- #
def _run_precip_check(
    model, precip, y: torch.Tensor, y_coords: dict, out_dir: Path
) -> None:
    """Diagnose precip from a coupled lat/lon output, mirroring the example.

    Takes the last two atmosphere lead times of ``y`` as the [-6h, 0h] history
    window, regrids them onto HEALPix, runs the diagnostic, and regrids tp06
    back to lat/lon for plotting and range checks."""
    precip_vars = list(precip.input_coords()["variable"])
    y_vars = list(y_coords["variable"])
    var_order = [y_vars.index(v) for v in precip_vars]

    precip_in = y[:, :, -2:][:, :, :, var_order]
    precip_coords = y_coords.copy()
    precip_coords["lead_time"] = y_coords["lead_time"][-2:]
    precip_coords["variable"] = np.array(precip_vars)

    precip_in_hpx = model.to_hpx(precip_in)
    precip_coords_hpx = model.coords_to_hpx(precip_coords)
    tp, _ = precip(precip_in_hpx, precip_coords_hpx)
    tp_ll = model.to_ll(tp)

    field = tp_ll[0, 0, 0, 0].cpu().numpy()  # (lat, lon)
    flat = field.ravel()
    tp_min, tp_mean, tp_max = float(flat.min()), float(flat.mean()), float(flat.max())
    n_neg = int((flat < -1e-4).sum())
    n_extreme = int((flat > 0.1).sum())

    print(f"\n[precip] tp06:  min={tp_min:+.3e}  mean={tp_mean:+.3e}  "
          f"max={tp_max:+.3e}  m/6h")
    if n_extreme:
        print(f"  [WARN] {n_extreme} pixels > 100 mm/6h")
    elif n_neg > 100:
        print(f"  [WARN] {n_neg} negative-precip pixels")
    else:
        print("  [OK] tp06 range is physical")

    lat = np.asarray(y_coords["lat"])
    lon = np.asarray(y_coords["lon"])
    grid = np.clip(field, 0.0, None)
    vmax = max(float(np.nanpercentile(grid, 99.5)), 1e-6)
    fig, ax = _make_map_axes(1, 1, (9, 4))
    im = _draw_map(ax, grid, lat, lon, 0, vmax, "Blues")
    ax.set_title(
        f"Precip sanity — tp06 (first coupled output)\n"
        f"min={tp_min:.3e}  mean={tp_mean:.3e}  max={tp_max:.3e}  m/6h"
    )
    fig.colorbar(im, ax=ax, label="tp06 (m / 6 h)", shrink=0.7)
    fig.tight_layout()
    p = out_dir / "precip_ic_sanity.png"
    fig.savefig(p, dpi=120)
    plt.close(fig)
    print(f"[plot] {p}")


# --------------------------------------------------------------------------- #
# Main
# --------------------------------------------------------------------------- #
def main() -> int:
    p = argparse.ArgumentParser(
        description="DLESyMv0_ISCCP_ERA5 pure-E2S long-rollout stability check"
    )
    p.add_argument("--years", type=int, default=100)
    p.add_argument("--device", default="cuda:0")
    p.add_argument("--out-dir", default="longrollout_out")
    p.add_argument(
        "--init", default="2021-06-15", help="Init date (ARCO ERA5), default 2021-06-15"
    )
    p.add_argument(
        "--plot-every",
        type=int,
        default=0,
        help="Save a state-panel plot every N steps (0 = final only)",
    )
    p.add_argument(
        "--skip-precip",
        action="store_true",
        help="Skip the precip sanity check (faster startup)",
    )
    p.add_argument(
        "--no-ttr",
        action="store_true",
        help=(
            "Load the model with use_ttr=False (OLR input space). By default "
            "use_ttr=True so the full ERA5 ttr → OLR transform code path runs."
        ),
    )
    args = p.parse_args()

    init_time = np.datetime64(args.init)

    device = args.device
    if device.startswith("cuda") and not torch.cuda.is_available():
        print(f"[warn] {device} not available, falling back to cpu")
        device = "cpu"

    out_dir = Path(args.out_dir)
    out_dir.mkdir(parents=True, exist_ok=True)

    pkg, cfg = _load_pkg()
    bounds = _build_bounds(cfg)
    if not _HAS_CARTOPY:
        print("[warn] cartopy not available — maps drawn without coastlines")

    print(f"[bounds] {SIGMA_WARN}σ warn / {SIGMA_HARD}σ hard  (from config.yaml scaling stats)")
    for v, (wlo, whi, hlo, hhi) in sorted(bounds.items()):
        print(f"  {v:14s}  warn=[{wlo:+.3e}, {whi:+.3e}]  hard=[{hlo:+.3e}, {hhi:+.3e}]")

    # ── pure-E2S model + data source ──────────────────────────────────────────
    from earth2studio.data import ARCO
    from earth2studio.data.utils import fetch_data
    from earth2studio.models.dx import DLESyMv0_ISCCP_ERA5Precip
    from earth2studio.models.px import DLESyMv0_ISCCP_ERA5LatLon

    use_ttr = not args.no_ttr
    model = DLESyMv0_ISCCP_ERA5LatLon.load_model(pkg, use_ttr=use_ttr).to(device)
    data = ARCO()

    # ── build IC from ARCO ERA5 (lat/lon, pure E2S) ──────────────────────────
    in_coords = model.input_coords()
    print(f"\n[IC] init={init_time}  use_ttr={use_ttr}")
    print(f"[IC] input variables: {list(in_coords['variable'])}")
    x, coords = fetch_data(
        source=data,
        time=np.array([init_time]),
        variable=np.array(in_coords["variable"]),
        lead_time=in_coords["lead_time"],
        device=device,
    )
    x = x[None]
    coords = coords.copy()
    coords["batch"] = np.array([0])
    coords.move_to_end("batch", last=False)

    # ── precip sanity (one coupled step, before committing to 100 years) ──────
    if not args.skip_precip:
        precip = DLESyMv0_ISCCP_ERA5Precip.load_model(pkg, use_ttr=False).to(device)
        with torch.no_grad():
            y0, y0c = model(x, coords)
        _run_precip_check(model, precip, y0, y0c, out_dir)
        del y0, precip
        if device.startswith("cuda"):
            torch.cuda.empty_cache()

    # ── long rollout ─────────────────────────────────────────────────────────
    # 1 coupled step = 96 h = 4 days
    total_steps = int(args.years * 365.25 * 24 / 96)
    report_every = max(1, total_steps // 100)

    violations: list[str] = []
    hard_violations: list[str] = []
    last_y: torch.Tensor | None = None
    last_yc: dict | None = None
    first_violation_step: int | None = None

    print(
        f"\n[longrollout] {args.years} years → {total_steps} coupled steps  "
        f"device={device}"
    )
    print("[longrollout] starting ...\n")

    it = model.create_iterator(x, coords)
    next(it)  # skip IC yield

    t0_wall = time.time()
    with torch.no_grad():
        for step in range(1, total_steps + 1):
            y, yc = next(it)
            _check_step(y, yc, bounds, step, violations, hard_violations)
            last_y, last_yc = y, yc

            if hard_violations and first_violation_step is None:
                first_violation_step = step
                # Save a snapshot immediately at the first hard violation
                vt = np.datetime64(yc["time"][0]) + yc["lead_time"][-1]
                _plot_fields(
                    y, yc,
                    out_dir / f"state_first_hard_step{step:06d}.png",
                    f"FIRST HARD VIOLATION — step {step} | {vt}",
                    hard_violations=hard_violations,
                )

            if args.plot_every and step % args.plot_every == 0:
                vt = np.datetime64(yc["time"][0]) + yc["lead_time"][-1]
                _plot_fields(
                    y, yc,
                    out_dir / f"state_step{step:06d}.png",
                    f"Step {step} | {vt}",
                )

            if step % report_every == 0 or step == total_steps:
                elapsed = time.time() - t0_wall
                vt = np.datetime64(yc["time"][0]) + yc["lead_time"][-1]
                pct = 100 * step / total_steps
                eta_h = (elapsed / step) * (total_steps - step) / 3600
                print(
                    f"  step {step:6d}/{total_steps}  ({pct:5.1f}%)  vt={vt}  "
                    f"elapsed={elapsed/3600:.2f}h  eta={eta_h:.2f}h  "
                    f"warns={len(violations)}  hards={len(hard_violations)}"
                )

            if device.startswith("cuda") and step % 500 == 0:
                torch.cuda.empty_cache()

    # ── final plot ───────────────────────────────────────────────────────────
    if last_y is not None and last_yc is not None:
        final_vt = np.datetime64(last_yc["time"][0]) + last_yc["lead_time"][-1]
        _plot_fields(
            last_y,
            last_yc,
            out_dir / "final_step.png",
            f"Final step ({total_steps}) | {final_vt} | {args.years}-year rollout",
            hard_violations=hard_violations,
        )

    # ── summary ──────────────────────────────────────────────────────────────
    elapsed_total = time.time() - t0_wall
    print("\n" + "=" * 70)
    print(f"[summary]  {total_steps} coupled steps ({args.years} years)")
    print(f"[summary]  wall-clock: {elapsed_total/3600:.2f} h")
    print(f"[summary]  {len(violations)} warn violations (>{SIGMA_WARN}σ)")
    print(f"[summary]  {len(hard_violations)} HARD violations (>{SIGMA_HARD}σ or non-finite)")

    if violations:
        print(f"\n[warns] (first 20 of {len(violations)})")
        for msg in violations[:20]:
            print(f"  {msg}")
        if len(violations) > 20:
            print(f"  ... and {len(violations) - 20} more")

    if hard_violations:
        print(f"\n[HARD VIOLATIONS] ({len(hard_violations)} total)")
        for msg in hard_violations:
            print(f"  {msg}")
        print(f"\n[FAIL] hard violations detected — see above and {out_dir}")
    else:
        print(
            f"\n[PASS] no hard violations over {args.years} years "
            f"({total_steps} coupled steps).  "
            f"{'No warnings either.' if not violations else f'{len(violations)} warn-level exceedances (>{SIGMA_WARN}σ) are normal climate variability.'}"
        )

    print(f"[done] outputs → {out_dir.resolve()}")
    return 1 if hard_violations else 0


if __name__ == "__main__":
    import sys
    sys.exit(main())
```

</details>

earth2studio/models/dx/dlesym_v0_isccp_era5_precip.py
```diff
@@ -0,0 +1,473 @@
...
+from collections import OrderedDict
+from pathlib import Path
+
+import numpy as np
+import torch
+
+from earth2studio.models.auto import AutoModelMixin, Package
+from earth2studio.models.batch import batch_coords, batch_func
+from earth2studio.models.dx.base import DiagnosticModel
+from earth2studio.models.px.dlesym_v0_isccp_era5 import (
```
NickGeneva: These aren't too complicated, can these just get repeated in this file to eliminate the import to a px file?

earth2studio/models/px/dlesym_v0_isccp_era5.py
```diff
@@ -0,0 +1,720 @@
...
+
+    The architecture is similar to
+    :class:`~earth2studio.models.px.DLESyM` -- both use
+    ``physicsnemo.models.dlwp_healpix`` on a HEALPix ``nside=64``
+    (≈ 1°) grid with coupled atmosphere/ocean rollout -- but the upstream
+    checkpoints carry a different variable set (9 atmospheric variables
+    including outgoing longwave radiation trained on ISCCP-distributed OLR,
+    and a 3-variable ocean coupling) rather than ERA5 TTR.
+
+    When ``use_ttr=True`` (default), the wrapper accepts ERA5 ``ttr`` and
```
NickGeneva, line 162: This seems complicated, is ttr an uncommon variable? just wondering if the option needs to exist at all

earth2studio/models/px/dlesym_v0_isccp_era5.py
```diff
@@ -0,0 +1,720 @@
...
+    .. code-block:: python
+
+        pkg = DLESyMv0_ISCCP_ERA5.load_default_package()
+        model = DLESyMv0_ISCCP_ERA5.load_model(pkg, use_ttr=True)
+        for step, (x, coords) in enumerate(model.create_iterator(x0, coords0)):
+            ...
+
+    Badges
+    ------
+    region:global class:cm product:wind product:temp product:atmos product:ocean year:2024
```
NickGeneva, line 217: This has some radiation vars as well... maybe `product:solar`

earth2studio/models/px/dlesym_v0_isccp_era5.py
```diff
@@ -0,0 +1,720 @@
...
+
+        When ``use_ttr=True``, ``rlut`` in the model's atmos variable list is
+        replaced by ``ttr`` so that any ERA5-compatible
+        :class:`~earth2studio.data.DataSource` resolves the input cleanly.
+        """
+        coords = super().input_coords()
+        # ``use_ttr`` is set after the parent __init__ chain runs, and the
+        # parent's __init__ calls ``self.input_coords()`` while computing
+        # variable indices. Default to False until the subclass finishes init.
+        if getattr(self, "use_ttr", False):
```
NickGeneva: why is this if statement needed? I'm not following the comment, as long as use_ttr is defined before calling this, this should be defined right?

earth2studio/models/px/dlesym_v0_isccp_era5.py
```diff
@@ -0,0 +1,720 @@
...
+            self.register_buffer(
+                "olr_clim_std", torch.from_numpy(np.asarray(olr_clim_std)).float()
+            )
+
+    def input_coords(self) -> CoordSystem:
+        """Input coordinate system of the prognostic model.
+
+        When ``use_ttr=True``, ``rlut`` in the model's atmos variable list is
+        replaced by ``ttr`` so that any ERA5-compatible
+        :class:`~earth2studio.data.DataSource` resolves the input cleanly.
```
NickGeneva: Nit: this second bit of the doc string seems redundant with the parameters of the class....

maybe just remove this sentence

earth2studio/models/px/dlesym_v0_isccp_era5.py
```diff
@@ -0,0 +1,720 @@
...
+            axis=1,
+        )
+
+        ttr_clim_mean = ttr_clim_std = None
+        olr_clim_mean = olr_clim_std = None
+        if use_ttr:
+            try:
+                ttr_path = Path(package.resolve(_TTR_CLIM_FILE))
+                olr_path = Path(package.resolve(_OLR_CLIM_FILE))
+            except Exception as e:
```
NickGeneva: When would this occur? Wondering if try catch is really needed

earth2studio/models/px/dlesym_v0_isccp_era5.py
```diff
@@ -0,0 +1,720 @@
...
+        Per-day-of-year mean of ISCCP OLR (same shape).
+    olr_clim_std : np.ndarray | None, optional
+        Per-day-of-year std of ISCCP OLR (same shape).
+    olr_floor : float, optional
+        Lower bound applied to the transformed OLR (the upstream pipeline
+        uses a small positive quantile floor). Defaults to 0.0.
+    **kwargs
+        Keyword arguments forwarded to :class:`DLESyM`.
+
+    Note
```
NickGeneva, line 195: Can you add a note block with the link to the papers?

earth2studio/models/px/dlesym_v0_isccp_era5.py
```diff
@@ -0,0 +1,720 @@
...
+        # x, coords come straight from an ERA5 data source on the lat/lon grid
+        x, coords = fetch_data(...)
+        y, y_coords = model(x, coords)
+
+        atmos, atmos_coords = model.retrieve_valid_atmos_outputs(y, y_coords)
+        ocean, ocean_coords = model.retrieve_valid_ocean_outputs(y, y_coords)
+
+    Badges
+    ------
+    region:global class:cm product:wind product:temp product:atmos product:ocean year:2024
```
NickGeneva, line 595: same product update here if solar is one

earth2studio/models/dx/dlesym_v0_isccp_era5_precip.py
```diff
@@ -0,0 +1,473 @@
...
+                "same_names": True,
+            },
+        )
+
+    @classmethod
+    @check_optional_dependencies()
+    def load_model(
+        cls,
+        package: Package,
+        model_idx: int = 0,
```
NickGeneva: What is this model index? Does it matter and how many indexes are there?

examples/06_seasonal/03_dlesym_v0_isccp_era5_example.py
```diff
@@ -0,0 +1,307 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# %%
+"""
+Running Upstream DLESyM (Cresswell-Clay et al. 2024)
```
NickGeneva: Needs a better title for the docs page, concise and without the paper ref. Will look weird

examples/06_seasonal/03_dlesym_v0_isccp_era5_example.py
```diff
@@ -0,0 +1,307 @@
...
+#
+# - Use :py:class:`earth2studio.models.px.DLESyMv0_ISCCP_ERA5LatLon`. This
+#   variant accepts ERA5 inputs on the lat/lon grid, regrids them to HEALPix
+#   internally, and returns lat/lon outputs. This is the recommended entry point
+#   and is what we use throughout this example.
+# - Use :py:class:`earth2studio.models.px.DLESyMv0_ISCCP_ERA5` directly with
+#   HEALPix inputs, handling the regridding and derived-variable preparation
+#   yourself (see the base DLESyM example for that lower-level pattern).
+#
+# The upstream checkpoints are published to HuggingFace at
```
NickGeneva: nit: not really needed to be stated

examples/06_seasonal/03_dlesym_v0_isccp_era5_example.py
```diff
@@ -0,0 +1,307 @@
...
+
+x, coords = fetch_data(
+    source=data,
+    time=np.array([ic_date]),
+    variable=np.array(in_coords["variable"]),
+    lead_time=in_coords["lead_time"],
+    device=device,
+)
+
+# ``fetch_data`` returns data without a batch dimension. Add an explicit leading
```
NickGeneva: why do this? not following, it should be able to handle no batch fine

earth2studio/models/dx/dlesym_v0_isccp_era5_precip.py
pzharrington: It was copying the existing (multi-checkpoint) DLESyM package layout and ingest pattern, but yeah just one model so this arg can be dropped

examples/06_seasonal/03_dlesym_climate.py
```diff
@@ -0,0 +1,307 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES.
```
NickGeneva, line 1: nit: maybe clean up the file name

earth2studio/models/px/dlesym_v0_isccp_era5.py
pzharrington: The more uncommon variable is ISCCP OLR, but that is what the model is trained on and predicts. So probably 90% of users will go with `use_ttr=True` and init from ERA5 TTR, but a select few that happen to have the OLR data can run it directly without the TTR->OLR transform. Mainly was convenient to have for parity testing against the original reference implementation

earth2studio/models/px/dlesym_v0_isccp_era5.py
pzharrington: Outgoing longwave radiation (or TTR) are more relevant to cloud/moisture processes rather than downward irradiance stuff relevant for solar, IIUC.

earth2studio/models/px/dlesym_v0_isccp_era5.py
pzharrington: Heh, they were probably not found when I was mocking this up and didn't have the data files yet. Can remove

examples/06_seasonal/03_dlesym_v0_isccp_era5_example.py
pzharrington: Idk why Claude did this, good catch. Not needed

earth2studio/models/px/dlesym_v0_isccp_era5.py
NickGeneva: Sounds good.

earth2studio/models/px/dlesym_v0_isccp_era5.py
NickGeneva: nvm

examples/06_seasonal/03_dlesym_v0_isccp_era5_example.py
NickGeneva: claude doing claude things :paintbrush:

earth2studio/models/px/dlesym_v0_isccp_era5.py
pzharrington: It's a subclassing thing. The base `DLESyM` class does not define a `use_ttr` attribute since it doesn't need a TTR->OLR transform. When the parent `DLESyM` init runs (triggered via the `super().__init__()` in this subclass's init), it runs `in_coords = self.input_coords()`, but due to Python dispatch this subclass's version of the method is used. That's when it would hit an undefined `use_ttr` and error, hence the `getattr(self, "use_ttr", False)`.

pzharrington: But, having typed that out, I see how the comment does not clarify things and it can probably be simplified by doing:
```
def __init__(self, ..., use_ttr: bool = True, ...):
    self.use_ttr = use_ttr
    super().__init__(*args, **kwargs)
    ...
```
Then keep the current override shape:
```
def input_coords(self) -> CoordSystem:
    coords = super().input_coords()
    if self.use_ttr:
        variables = list(coords["variable"])
        variables[variables.index("rlut")] = "ttr"
        coords["variable"] = np.array(variables)
    return coords
```

NickGeneva (approved):
LGTM

earth2studio/models/dx/dlesym_v0_isccp_era5_precip.py
pzharrington: Done

PR #931, opened 2026-06-20
Fix ZarrBackend chunk reload for coordinate arrays

fallintoplace:
## Summary
- Load Zarr dimension names before checking whether an array is a coordinate array.
- Skip coordinate arrays when rebuilding chunk metadata while reopening stores.
- Add a regression test that forces a coordinate/data/coordinate iteration order.

## Testing
- `PYTHONPATH=. uv run --no-project --python 3.12 --with pytest --with numpy --with torch --with xarray --with zarr --with netCDF4 --with cftime --with fsspec --with loguru --with h5netcdf --with rich pytest test/io/test_zarr.py -k 'cpu or reopen or uninitialized'`\n- `uv run --no-project --python 3.12 --with ruff ruff check earth2studio/io/zarr.py test/io/test_zarr.py`


NickGeneva (approved):
LGTM, thanks for the patch

PR #919, opened 2026-06-10
Stormcast-CONUS

jleinonen:
# Earth2Studio Pull Request

## Description

Adds StormCast-CONUS model: A CONUS-wide StormCast-like model with a DiT backbone.

Main features/additions:

1. `models.nn.stormcastconus.StormCastCONUSBase` model wrapper:
   - Uses `_SplitModelWrapper` to handle per-sigma and per-variable expert split
   - Support for SDA
   - Support for splitting model to subregion of full CONUS
   - Optimizations: toggleable model compilation and AMP inference
2. `models.px.stormcastconus.StormCastCONUS` and `models.da.sda_stormcastconus.StormCastCONUSSDA` specialize the above as forecasting and DA models
3. `utils.obs.ObsGridMapping` for converting between point obs and gridded variables, with CuPy support
3. ~~`utils.cos_zenith` for a torch port of cos zenith calculations (Stormscope uses version from PhysicsNeMo, which is implemented in NumPy and runs on CPU)~~
4. Example, following the structure of the StormCast example
5. SDA example, following the structure of the StormCast SDA example
6. Earth2Studio API example utilizing StormCast-CONUS
7. Unit tests
8. Docs updated

TBA/TBD:

1. ~~Do we want to further generalize `SplitModelWrapper`?~~
2. ~~Discuss what to do with `utils.cos_zenith` which admittedly adds code duplication~~
3. ~~Add example using SDA~~
4. ~~Add workflow for Earth2Studio API for using the model with various conditioning models (e.g. GFS, FCN3, AIFS)~~
5. ~~Add unit tests~~

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/earth2studio/blob/main/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.
- [ ] The [CHANGELOG.md](https://github.com/NVIDIA/earth2studio/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/earth2studio/issues) is linked to this pull request.
- [x] Assess and address Greptile feedback (AI code review bot for guidance; use discretion, addressing all feedback is not required).

PR #939, opened 2026-06-26
feat: add CosmoDownscaling diagnostic model (ERA5 -> COSMO-REA6/REA2)

### LINKED ISSUE #940: 🚀[FEA]: Add CosmoDownscaling diagnostic model (ERA5 → COSMO-REA6/REA2)

New Feature

Add a diagnostic model that downscales global ERA5 to high-resolution COSMO-REA regional reanalysis over Europe for COSMO-REA6 (~6 km) and COSMO-REA2 (~2.2 km).

**Motivation.** Earth2Studio has global prognostic models but no high-resolution regional downscaler for the European COSMO-REA reanalysis. A diagnostic that takes an ERA5 state lets users downscale an ERA5 analysis directly or run behind a global forecast (e.g. `SFNO` → `CosmoDownscaling`).

**Proposed scope.** A single `CosmoDownscaling` diagnostic with two `load_model` selectors: `mode` (`mean` deterministic regression / `diffusion` generative ensemble) and `resolution` (`rea6` / `rea2`). Movable sub-domains via `set_domain`, optional derived hub-height winds, plus a runnable example.

Implemented in #939

gertln:
## Description

Add `CosmoDownscaling`, a diagnostic model that downscales global ERA5 to high-resolution COSMO-REA regional reanalysis over Europe — COSMO-REA6 (~6 km) and COSMO-REA2 (~2.2 km). Its input is an ERA5 state, so it can downscale an ERA5 analysis directly or run behind a global forecast model (e.g. SFNO → CosmoDownscaling). A runnable example is included.

A single class loads one of four checkpoints via two `load_model` selectors:

- `mode="mean"` — a deterministic regression that predicts the conditional mean (the expected field, `E[y | x]`): a single, smooth field in one forward pass. Fast and cheap — a good first high-resolution look.
- `mode="diffusion"` — a generative diffusion model that samples the conditional distribution `p(y | x)`: an ensemble of realizations that also captures the spread the mean cannot represent (~N forward passes; seeded for reproducibility).
- `resolution="rea6"|"rea2"` — selects the per-resolution weights and grid from one package.

Both modes use a PhysicsNeMo DiT (diffusion transformer) at a fixed output resolution; the network is crop-size agnostic, so `set_domain(...)` returns a new instance for lat/lon sub-region. From a small box up to the full grid. Optional hub-height wind adds derived `u{H}m`/`v{H}m` components (wind speed via the stock `DerivedWS`)

Blocked by:
Closes: #940 

### Diagnostic details

| Property | Value |
|---|---|
| **Diagnostic type** | Generative (EDM/Karras diffusion) + deterministic regression, one class via `mode` |
| **Framework** | PyTorch (PhysicsNeMo DiT) |
| **Input variables** | ERA5 surface + pressure-level fields (package metadata; see `input_coords()`) |
| **Output variables** | COSMO-REA surface **and** model-level (3D) fields — surface temperature/wind/precip/cloud/fluxes plus model-level winds, temperature, humidity, and TKE; overlap vars mapped to the Earth2Studio lexicon, COSMO-only kept raw; plus optional derived hub-height `u{H}m`/`v{H}m` |
| **Input spatial resolution** | ERA5 regular lat/lon crop on the native footprint |
| **Output spatial resolution** | 2D rotated-pole curvilinear grid: COSMO-REA6 ~6 km / COSMO-REA2 ~2.2 km |
| **Samples** | `sample` dim always present: 1 for `mean`, `number_of_samples` for `diffusion` (per-sample seed `seed + i`) |
| **Reference** | COSMO-REA6 / COSMO-REA2 regional reanalysis (DWD and the Hans-Ertel-Centre for Weather Research, University of Bonn) |
| **GitHub** | https://github.com/NVIDIA/earth2studio |

### Dependencies added

New `cosmo` model extra (`pip install earth2studio[cosmo]` / `uv add earth2studio --extra cosmo`), also added to the `all` extra.

| Package | Version | License | License URL | Reason |
|---|---|---|---|---|
| `nvidia-physicsnemo` | `>=2.0` | Apache-2.0 | https://github.com/NVIDIA/physicsnemo/blob/main/LICENSE.txt | DiT (RoPE + NATTEN), EDM preconditioner/sampler, zenith-angle util |
| `natten` | `; python_version < '3.14'` | MIT | https://github.com/SHI-Labs/NATTEN/blob/main/LICENSE | Neighborhood attention in the DiT |
| `einops` | `>=0.8.1` | MIT | https://github.com/arogozhnikov/einops/blob/master/LICENSE | Tensor rearranges in the network |
| `nvtx` | `>=0.2.11` | Apache-2.0 | https://github.com/NVIDIA/NVTX/blob/release-v3/LICENSE.txt | Profiling ranges (consistent with other model extras) |

All permissive (Apache-2.0 / MIT); no non-permissive licenses introduced.

### Validation

- **CPU suite** (`test/models/dx/test_cosmo_downscaling.py`): 51 passed, 5 skipped. The 5 skips are intentionally resource-gated tests — the 4 real-weight GPU `--package` cases and 1 real-DiT-builder smoke test — which require a GPU, a built package, and/or physicsnemo's not-yet-released RoPE op, so they cannot run in the default CI lane (the standard `@pytest.mark.package` pattern used by every model).
- **GPU lane** (those gated tests, run locally on real weights): all pass — the 4 `{rea6, rea2} × {mean, diffusion}` combos load 0 missing / 0 unexpected keys and run a full `dx(x, coords)` forward (output shape, finiteness, and every metadata min/max bound asserted in both modes), plus the DiT-builder smoke test.
- `black` + `ruff` clean; SPDX headers present.
- Example `examples/03_downscaling/04_cosmo_rea_downscaling.py` verified end-to-end on real weights (full domain, sub-domain, rollout, diffusion ensemble vs regression mean, hub-height wind, COSMO-REA2).

The CPU tests are construction-driven and GPU-free: they pin the wrapper's contract — coords/handshake + lexicon output names, channel-transform round-trips, physical-space min/max bounds and a mode-identical solar-zenith day/night gate, `set_domain` sub-domain slicing (incl. NATTEN-minimum and patch-snap edges), the diffusion sampler's schedule termination + ensemble seeding, and the hub-wind derivation + `DerivedWS` naming handshake.

## Notes for reviewers

Two known, bounded items, each guarded so CI stays green today:

1. **PhysicsNeMo RoPE not yet released.** The DiT uses `natten2d_rope`, merged on physicsnemo `main` but not in a tagged release. The `cosmo` extra therefore pins `nvidia-physicsnemo>=2.0`, and the real-DiT-builder + `--package` tests are skip-guarded when the op is unavailable (they skip, not fail). A follow-up bumps the pin and drops the guards once a release includes it.
2. **Weights not hosted yet.** `load_default_package` raises `NotImplementedError`; the example is in `expected_failing_examples` in `docs/conf.py` so the gallery build stays green. Flipping `DEFAULT_PACKAGE_URI` on and removing the whitelist entry is the only change needed once hosting lands. (The model was validated against real weights locally.)

## Checklist

- [x] I am familiar with the [Contributing Guidelines][contrib].
- [x] The [CHANGELOG.md][changelog] is up to date with these changes.
- [x] An [issue][issues] is linked to this pull request.
- [x] Assess and address Greptile feedback (AI code review bot).

[contrib]: https://github.com/NVIDIA/earth2studio/blob/main/CONTRIBUTING.md
[changelog]: https://github.com/NVIDIA/earth2studio/blob/main/CHANGELOG.md
[issues]: https://github.com/NVIDIA/earth2studio/issues


gertln:
cosmo_domains

Output domains: native and extended grid footprints for both resolutions. REA6 spans broad Europe at ~6 km; REA2 is central-European natively at ~2.2 km, with an extended grid reaching a comparable broad domain.

| Domain | Resolution | Grid size |
|---|---|---|
| REA6 native | ~6 km | 824 × 848 |
| REA6 extended | ~6 km | 1006 × 1030 |
| REA2 native | ~2.2 km | 780 × 724 |
| REA2 extended | ~2.2 km | 2780 × 2724 |

earth2studio/models/dx/cosmo_downscaling.py
```diff
@@ -0,0 +1,1995 @@
...
+    def _forward(
+        self,
+        era5: torch.Tensor,
+        valid_time: datetime,
+        lat2d: torch.Tensor,
+        lon2d: torch.Tensor,
+    ) -> torch.Tensor:
+        """Forward for one ERA5 frame -> [number_of_samples, C_out, H, W] physical
+        (C_out = the trained outputs, plus the derived hub-height wind components
+        when ``hub_heights`` is set).
```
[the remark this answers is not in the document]
gertln: The network is shared across set_domain sub-domains and each forward sets its grid on it in place. So sub-domains are fine to run sequentially but two running at the same time would lead to conflicts. I documented this in set_domain and _rebind_latent with the advice to run concurrent domains in separate processes. (d498ccf2)

earth2studio/models/dx/cosmo_downscaling.py
```diff
@@ -0,0 +1,1995 @@
...
+            "qk_norm": rp.get("qk_norm", False),
+        }
+        backend = rp.get("natten_backend", "cutlass-fna")
+        if backend is not None:
+            attn["na2d_kwargs"] = {"backend": backend}
+        conv_layers = rp.get("detok_conv_layers", 0)
+        return DiT(
+            input_size=list(rp.get("img_resolution", [256, 256])),
+            in_channels=cond_channels,
+            out_channels=out_channels,
```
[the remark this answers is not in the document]
gertln: Fixed in d498ccf2. set_domain now reads the input spacing from the package's own grid instead of assuming 0.25.

earth2studio/models/dx/cosmo_downscaling.py
```diff
@@ -0,0 +1,1995 @@
...
+            "qk_norm": dp.get("qk_norm", False),
+        }
+        backend = dp.get("natten_backend", "cutlass-fna")
+        if backend is not None:
+            attn["na2d_kwargs"] = {"backend": backend}
+        dit = DiT(
+            input_size=list(dp.get("img_resolution", [256, 256])),
+            in_channels=out_channels + cond_channels,
+            out_channels=out_channels,
+            patch_size=dp["patch_size"],
```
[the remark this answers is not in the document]
gertln: By design, now documented (d498ccf2). A sub-domain only keeps its own cropped grid, so calling set_domain again on it can't go back to the full footprint.

earth2studio/models/dx/cosmo_downscaling.py
```diff
@@ -0,0 +1,1995 @@
...
+    """
+
+    def __init__(
+        self,
+        era5_variables: Sequence[str],
+        output_variables: Sequence[str],
+        regression_model: torch.nn.Module | None,
+        diffusion_model: torch.nn.Module | None,
+        resolution: str,
+        mode: str,
```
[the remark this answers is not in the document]
gertln: renamed the local to interp_val so it no longer shadows the interp import (d498ccf2)

docs/conf.py
```diff
@@ -214,6 +214,12 @@ def linkcode_resolve(domain, info):
     "show_memory": False,
     "exclude_implicit_doc": {r"load_model", r"load_default_package"},
     "log_level": {"backreference_missing": "warning", "gallery_examples": "debug"},
+    # COSMO-REA weights are not yet publicly hosted, so this example raises without
+    # a local $COSMO_REA_PACKAGE. Whitelist it so the full gallery build does not
+    # fail; remove this entry once the package is hosted (load_default_package).
```
pzharrington: Add a TODO so we don't forget this

earth2studio/models/dx/__init__.py
```diff
@@ -24,6 +24,7 @@
     CorrDiffTaiwan,
 )
 from earth2studio.models.dx.corrdiff_cmip6 import CorrDiffCMIP6
+from earth2studio.models.dx.cosmo_downscaling import CosmoDownscaling  # noqa
```
pzharrington: Why `# noqa`?

pyproject.toml
```diff
@@ -238,6 +238,12 @@ corrdiff = [
     "nvtx>=0.2.11",
     "scipy>=1.15.2",
 ]
+cosmo = [
+    "einops>=0.8.1",
+    "natten ; python_version < '3.14'",
+    "nvidia-physicsnemo>=2.0",
```
pzharrington: I believe we need a stricter constraint on this if we need RoPE stuff from what I recently merged

earth2studio/models/dx/cosmo_downscaling.py
```diff
@@ -0,0 +1,2013 @@
...
+
+"""COSMO-REA downscaling diagnostic models (ERA5 -> COSMO-REA6 / REA2).
+
+Earth2Studio :class:`DiagnosticModel` wrapper for the ``cosmo-rea-downscaling``
+package. One class: ``mode`` selects mean (deterministic) or diffusion (sampled
+with the EDM/Karras scheme); ``resolution`` selects the rea6/rea2 checkpoint. Both
+modes are DiT-RoPE (a diffusion transformer with rotary position embedding) and
+resolution-agnostic (single forward at any grid size).
+
+Domain handling (see the integration handoff + design memory):
```
pzharrington: Don't reference implementation notes or agent memory context

earth2studio/models/dx/cosmo_downscaling.py
```diff
@@ -0,0 +1,2013 @@
...
+  an **extended** grid + invariants covering a margin around it. Full-domain
+  ``__call__`` takes input regridded onto ``input_coords()`` (the native footprint);
+  sub-regions -- including into the extended margin -- are reached via :meth:`set_domain`
+  (next bullet).
+* A **sub-region** is obtained with :meth:`set_domain`, which slices the grid +
+  exact invariants to a bounding box (no aggregation). The bbox may lie inside the
+  native trained footprint, or reach into the extended invariant margin -- the
+  latter proceeds with a one-time out-of-distribution warning; beyond the extended
+  extent raises. Both the mean and diffusion models are DiT-RoPE
+  (resolution-agnostic), so a sub-domain runs at any size in a single forward.
```
pzharrington: It's actually fixed resolution, and more domain/crop size agnostic, right?

earth2studio/models/dx/cosmo_downscaling.py
```diff
@@ -0,0 +1,2013 @@
...
+
+try:
+    import natten  # noqa: F401  # the DiT needs NATTEN (neighborhood attention)
+    from physicsnemo.utils.zenith_angle import cos_zenith_angle
+except ImportError:
+    OptionalDependencyFailure("cosmo")
+    cos_zenith_angle = None
+
+
+# Fixed background channel layout (see cosmo_rea2_era5.py::_load_full_frame).
```
pzharrington: Don't reference missing/external files

earth2studio/models/dx/cosmo_downscaling.py
```diff
@@ -0,0 +1,2013 @@
...
+    "10U": "u10m",
+    "V_10M": "v10m",
+    "10V": "v10m",
+    "T_2M": "t2m",
+    "2MT": "t2m",
+    "TD_2M": "d2m",  # 2 m dewpoint, K — canonical, identical units
+    "CLCT": "tcc",  # total cloud cover; % -> 0-1 fraction (COSMO_OUTPUT_UNIT_SCALE)
+    "PS": "sp",
+    "PMSL": "msl",
+}
```
pzharrington: Can we open and merge a COSMO-REA data source before this PR so that the model wrapper can simply reuse the COSMO lexicon? Adding a lexicon to the model wrapper file is not standard

pzharrington: And, I don't see a particular for the model wrapper to be doing any lexicon interaction in the first place. Wrappers operate based on `input_coords` and `output_coords`, which should specify variables that exist in the E2S lexicon. If there are things in the model package using these other names, ideally we rename them to avoid having to deal with lexicon conversions in model code. But, if absolutely necessary, then we should pull from a COSMO lexicon as I suggested above

earth2studio/models/dx/cosmo_downscaling.py
pzharrington: Probably will pop up in several other places -- have your agents make a pass to remove the parts of comments like this that reference stuff which won't end up getting merged

earth2studio/models/dx/cosmo_downscaling.py
```diff
@@ -0,0 +1,2013 @@
...
+# ``derived.py::DerivedSurfacePressure``, a frozen full-domain aux grid). The
+# model also owns ``z0``/``SHFL_S`` and the resolution-specific level->channel
+# map. ``corrdiff.py`` is the precedent for keeping grid-tied invariants on the
+# model.
+#
+# Scope: the shipped ``"linear"`` (in z) and ``"log"`` (in ln z) methods are
+# z0-free (roughness cancels between the bracketing levels).
+
+
+def interp_levels_to_height(
```
pzharrington: I appreciate the design note to clarify intent/choices. I agree given the dependence on elevation it makes sense to attach this to the model. Now that we have it in a commit I think we can drop the lengthy comment explaining rationale.

As for the method itself, is there a reason to make it a standalone, separate public helper rather than a method of the class (possible just an internal helper method too, i.e. `_interp_levels_to_height`)?

earth2studio/models/dx/corrdiff_cosmo_era5.py
```diff
@@ -0,0 +1,2013 @@
...
+        interp_val = (
+            values[..., j, :, :] + (values[..., j + 1, :, :] - values[..., j, :, :]) * w
+        )
+        out = torch.where((t >= lo) & (t < hi), interp_val, out)
+    return torch.where(t >= h[k - 1], values[..., k - 1, :, :], out)
+
+
+@check_optional_dependencies()
+class CosmoDownscaling(torch.nn.Module, AutoModelMixin):
+    """COSMO-REA downscaling model: ERA5 -> high-resolution COSMO-REA.
```
pzharrington, line 222: This should eventually get filled out with details of the model, use-cases, intended usage, etc

earth2studio/models/dx/cosmo_downscaling.py
```diff
@@ -0,0 +1,2013 @@
...
+        Native, already-normalized static invariant channels (non-position),
+        each [out_lat, out_lon] — e.g. ``elevation_norm``, ``land_fraction``,
+        ``z0_lu``, ``slope_east``, ``slope_north``. Position channels (sin/cos
+        lat/lon) and ``cos_zenith`` are computed, not stored here.
+    pre_invariant_variables, post_invariant_variables : Sequence[str]
+        Names of the invariant channels before / after ``cos_zenith`` in the
+        background channel order.
+    channel_transforms : dict | None
+        Per-output-channel nonlinear transform spec (from the training
+        ``channel_transforms`` zarr attr), used to invert after de-normalizing.
```
pzharrington: training zarr attrs shouldn't be referenced

earth2studio/models/dx/cosmo_downscaling.py
```diff
@@ -0,0 +1,2013 @@
...
+    constraints : dict | None
+        Physical-constraint spec (``metadata["constraints"]``): per-channel
+        ``bounds`` (min/max clamps) and a shortwave ``sza_gate`` (the dawn/dusk
+        solar gate), applied in physical space in postprocess for both modes.
+    number_of_samples : int
+        Number of samples (diffusion); the ``sample`` dim is kept (size 1) for
+        ``mode="mean"`` so both modes share an output contract.
```
pzharrington: Why is this a constructor arg? Shouldn't it be dynamic?

earth2studio/models/dx/cosmo_downscaling.py
```diff
@@ -0,0 +1,2013 @@
...
+    physical_clamp : bool
+        Apply physical-bounds + the shortwave dawn/dusk solar gate to outputs.
+    number_of_steps : int
+        Diffusion sampler step count (EDM Heun). Defaults to 18.
+    sigma_min, sigma_max : float
+        Diffusion noise-schedule bounds. Default to 0.002 / 800.0.
+    rho : float
+        Karras noise-schedule exponent. Defaults to 7.0.
+    solver : str
```
pzharrington: Do the `rho` and `solver` need to be configurable via constructor args? Can they just be fixed or do they change?

earth2studio/models/dx/cosmo_downscaling.py
```diff
@@ -0,0 +1,2013 @@
...
+        roughness-free between levels. Defaults to ``"linear"``.
+    wind_levels : dict, optional
+        Package metadata describing the 3D wind levels for hub-height derivation:
+        each level's ``{u, v}`` output channels + its ``a, b`` height coefficients
+        (``height = a + b*elevation_invariant``) and the elevation invariant name.
+        Supplied by the package; required when ``hub_heights`` is set.
+
+    Badges
+    ------
+    region:eu class:ds product:wind product:precip product:temp product:atmos year:2024 gpu:80gb
```
pzharrington: Release year 2026

earth2studio/models/dx/cosmo_downscaling.py
```diff
@@ -0,0 +1,2013 @@
...
+    """
+
+    def __init__(
+        self,
+        era5_variables: Sequence[str],
+        output_variables: Sequence[str],
+        regression_model: torch.nn.Module | None,
+        diffusion_model: torch.nn.Module | None,
+        resolution: str,
+        mode: str,
```
pzharrington: Use `Literal[...]` for string args with a fixed set of options

earth2studio/models/dx/cosmo_downscaling.py
```diff
@@ -0,0 +1,2013 @@
...
+                d_cur = (x - net(x.float(), sig, condition=cond).double()) / t_cur
+                x_next = x + (t_next - t_cur) * d_cur
+                if self.solver == "heun" and i < n - 1:
+                    sig_n = t_next.to(torch.float32).repeat(x.shape[0])
+                    d_prime = (
+                        x_next - net(x_next.float(), sig_n, condition=cond).double()
+                    ) / t_next
+                    x_next = x + (t_next - t_cur) * 0.5 * (d_cur + d_prime)
+                x = x_next
+        return x.float()
```
pzharrington: Any reason to use a custom sampler over the standard `physicsnemo.diffusion` offerings?

earth2studio/models/dx/cosmo_downscaling.py
```diff
@@ -0,0 +1,2013 @@
...
+        extended one, so call ``set_domain`` once on the full loaded model: a
+        second ``set_domain`` on a returned sub-domain cannot reach back to the
+        full or extended footprint (a bbox outside the sub-domain's grid raises).
+        For several sub-regions, call ``set_domain`` multiple times on the full
+        model instead.
+
+        ``halo`` (px, default 0 = off): run on a block expanded by ``halo`` real
+        cells per side and trim it off the output, keeping the returned bbox
+        interior clear of the DiT's boundary artifact (~32 px); clamps + warns at
+        the grid edge. Both models are DiT-RoPE (resolution-agnostic), so any size
```
pzharrington: Same resolution-agnostic comment applies here

earth2studio/models/dx/cosmo_downscaling.py
```diff
@@ -0,0 +1,2013 @@
...
+            raise ValueError(
+                f"package metadata has no checkpoint entry {ckpt_key!r} "
+                f"(available: {list(metadata.get('checkpoints', {}))})."
+            ) from e
+        regression_model = None
+        diffusion_model = None
+        # Build the DiT with upstream physicsnemo's axial-2D-RoPE NATTEN backend
+        # (attention_backend="natten2d_rope" in _build_*_dit). RoPE adds no
+        # parameters, so a non-RoPE DiT would load the same weights 0/0 yet run a
+        # wrong (un-RoPE'd) forward -- building with the backend avoids that.
```
pzharrington: This approach of loading checkpoints with manual helpers extracting out of a zip file seems brittle and error-prone. If anything in upstream PNM changes about the internal zip format, it could break here, for example. I strongly prefer the standard `DiT.from_checkpoint(...)` approach.

earth2studio/models/dx/cosmo_downscaling.py
```diff
@@ -0,0 +1,2013 @@
...
+    def _build_diffusion_dit(
+        out_channels: int, cond_channels: int, dp: dict
+    ) -> torch.nn.Module:
+        """Build the EDM-preconditioned diffusion DiT from upstream physicsnemo:
+        ``EDMPreconditioner(ConcatConditionWrapper(DiT))``. The DiT sees
+        ``out_channels + cond_channels`` (noised target concat with the condition);
+        checkpoint keys are ``model.model.*`` plus ``sigma_data``."""
+        from physicsnemo.diffusion.preconditioners import EDMPreconditioner
+        from physicsnemo.diffusion.utils import ConcatConditionWrapper
+        from physicsnemo.models.dit import DiT
```
pzharrington: Move imports to the top in a standard protected import like other models (see e.g. StormScope)

earth2studio/models/dx/cosmo_downscaling.py
```diff
@@ -0,0 +1,2013 @@
...
+
+        attn = {
+            "attn_kernel": dp["attn_kernel_size"],
+            "rope_theta": dp.get("rope_theta", 10000.0),
+            "qk_norm": dp.get("qk_norm", False),
+        }
+        backend = dp.get("natten_backend", "cutlass-fna")
```
pzharrington: Why hardcode the backend to `cutlass-fna` (or manually impose it)? This could block e.g. the Hopper bf16 kernels getting used on H100. Natten should be able to select the appropriate one.

earth2studio/models/dx/corrdiff_cosmo_era5.py
```diff
@@ -0,0 +1,2013 @@
...
+    @check_optional_dependencies()
+    def load_model(
+        cls,
+        package: Package,
+        device: str | None = None,
+        mode: str = "mean",
+        resolution: str = "rea6",
+        hub_heights: Sequence[float] | None = None,
+        hub_interp: str = "linear",
+    ) -> DiagnosticModel:
```
pzharrington, line 1625: Use `Literal[...]` for all string args with a fixed set of options

earth2studio/models/dx/corrdiff_cosmo_era5.py
```diff
@@ -0,0 +1,2013 @@
...
+            conditioning_embedder="dit",
+            layernorm_backend=dp.get("layernorm_backend", "torch"),
+        )
+        return EDMPreconditioner(
+            model=ConcatConditionWrapper(dit), sigma_data=dp.get("sigma_data", 1.0)
+        )
+
+    @classmethod
+    @check_optional_dependencies()
+    def load_model(
```
pzharrington, line 1617: General comment. Add a minimal `config.json` in the top level of the model package directory (can just be  super simple like [this](https://huggingface.co/nvidia/stormscope-goes-mrms/blob/main/config.json)) if there aren't any files already named that (or `config.yaml`) . Then, add a line in the `load_model` to resolve/load that file from the package. That will enable automatic download tracking in HuggingFace (I know it's a weird system but it comes from HF rules, not ours)

earth2studio/models/dx/cosmo_downscaling.py
gertln: The constructor arg just provides the default but it is dynamic and can be set between calls. This matches CorrDiff behavior and the diagnostic `__call__(x, coords)` signature has no per-call kwargs, so sample count is carried as instance state. I added a note to the docstring making this behavior more explicit.

earth2studio/models/dx/cosmo_downscaling.py
gertln: They're kept configurable on purpose. We expose them because these settings (sigma_max, rho, solver, steps) can shape the sampled output distribution e.g. ensemble spread and how extremes are represented. So leaving them tunable is useful for inference experimentation and optimization. I've added a docstring note on this.

docs/conf.py
gertln: done (bab2e904)

earth2studio/models/dx/__init__.py
gertln: Remove it - not needed.

pyproject.toml
gertln: Pinned nvidia-physicsnemo>=2.2.0. natten2d_rope is only on main (2.2.0a0) and the latest release (2.1.1) doesn't have it. So it is intentionally uninstallable until 2.2.0 ships. (Also removed cosmo from the all extra because >=2.2.0 conflicts with da-healda <=2.1.1 )

earth2studio/models/dx/cosmo_downscaling.py
gertln: done

earth2studio/models/dx/cosmo_downscaling.py
gertln: yes, it is fixed grid resolution and flexible domain size. Reworded "resolution-agnostic" to "crop-size agnostic" here and everywhere it appeared.

earth2studio/models/dx/cosmo_downscaling.py
gertln: done

earth2studio/models/dx/cosmo_downscaling.py
gertln: Added a CosmoLexicon in earth2studio/lexicon/ and the wrapper uses it

earth2studio/models/dx/cosmo_downscaling.py
gertln: Trimmed the comment and made it a private module helper function `_interp_levels_to_height`

earth2studio/models/dx/corrdiff_cosmo_era5.py
gertln: added details about the two modes, use cases and intended usage

earth2studio/models/dx/cosmo_downscaling.py
gertln: removed this reference

earth2studio/models/dx/cosmo_downscaling.py
gertln: fixed -> 2026

earth2studio/models/dx/cosmo_downscaling.py
gertln: done

earth2studio/models/dx/cosmo_downscaling.py
gertln: Switched `_denoise` to the standard `EDMNoiseScheduler `+ `get_denoiser `+` sample()`

earth2studio/models/dx/cosmo_downscaling.py
gertln: fixed, renamed

earth2studio/models/dx/cosmo_downscaling.py
gertln: Switched to the standard `from_checkpoint`

earth2studio/models/dx/cosmo_downscaling.py
gertln: done

earth2studio/models/dx/cosmo_downscaling.py
gertln: I am no longer forcing a specifc backend

earth2studio/models/dx/corrdiff_cosmo_era5.py
gertln: done

earth2studio/models/dx/corrdiff_cosmo_era5.py
gertln: Added a minimal `config.json` at the package root

pyproject.toml
pzharrington: > So it is intentionally uninstallable until 2.2.0 ships

We can just have the `cosmo` dependency pin physicsnemo to a specific git commit from `main` and then it will be usable in the meantime. Just add a `TODO` comment to update to 2.2.0 when ready, and drop `cosmo` from the `all` group if there are any conflicts that arise

earth2studio/lexicon/cosmo.py
```diff
@@ -0,0 +1,117 @@
...
+    (e.g. ``t2m``, ``sp``); COSMO-specific fields with no canonical equivalent
+    (model-level winds/temperatures/humidity, turbulent fluxes, TKE, PBL height,
+    ...) keep a descriptive lowercase name.
+
+    Each entry's value is ``(cosmo_name(s), scale)``: ``cosmo_name`` is the
+    interior COSMO output name (a tuple when the two resolutions spell the same
+    field differently, e.g. ``U_10M`` in REA6 vs ``10U`` in REA2), and ``scale``
+    converts the COSMO value to the Earth2Studio name's units (e.g. ``CLCT`` is a
+    percentage and ``tcc`` is a 0-1 fraction, so ``scale = 0.01``).
+    """
```
pzharrington, line 45: Add some links/references for the COSMO-REA dataset

earth2studio/models/dx/corrdiff_cosmo_era5.py
pzharrington: Nice, looks good. Can we add an additional Note with references/links to the dataset?

examples/03_downscaling/04_cosmo_rea_downscaling.py
pzharrington, line 1: This example is pretty complex and long, I get that there are several different features to showcase though. How long does it take to run (on e.g. A100 or H100), and how much GPU memory is required?

gertln: with cached model weights it takes approx. 3 minutes to run and uses ~18GB peak GPU memory (tested on RTX A6000)

earth2studio/lexicon/cosmo.py
gertln: done

pyproject.toml
```diff
@@ -276,6 +282,9 @@ da-stormcast = [
     "cudf-cu13>=26.2.0",
 ]
 # All, must not have conflicts
+# NOTE: `cosmo` is intentionally excluded here. It needs nvidia-physicsnemo
+# >=2.2.0 (RoPE), which cannot co-resolve with `da-healda`'s <=2.1.1 cap in
+# `all`. TODO(cosmo): add `cosmo` back once `da-healda` moves off the 2.1.1 cap.
```
pzharrington: As-is, this will not really work for the July E2S release because physicsnemo 2.2.0 is not officially out yet. To make it work in the interim, add a pin to a specific physicsnemo commit (e.g. current `main` branch) as a dedicated source for the `cosmo` extra deps group, and specify the conflict with `healda`. Check out the setup for #919 , can we try to mirror that here?

test/models/dx/test_corrdiff_cosmo_era5.py
pzharrington, line 1: I appreciate the deep test coverage here. Does it take a long time for all of these to complete? If so, we could consider trimming them down, but if not we can keep them.

However, can we ensure the required tests are implemented directly (see [model creation skill](https://github.com/NVIDIA/earth2studio/blob/main/skills/earth2studio-create-diagnostic/SKILL.md#step-7---write-tests) for details)?

> | Function | Purpose |
> |----------|---------|
> | `test_<model>_call` | Forward pass with mock or simple model |
> | `test_<model>_exceptions` | Invalid coordinate order, values, or variables raise errors |
> | `test_<model>_package` | Real weights with `@pytest.mark.package` for AutoModel/generative diagnostics |
> 
> Generative diagnostics also require sample-count and deterministic-seed tests.

Some of the existing tests likely cover these and could be renamed if this is exactly what they target, otherwise create new tests specifically addressing these to make sure we have them in.

gertln: Thanks. Adressed in (9539c95). On timing all tests run in ~7s.

pzharrington (approved):
Approved. FYI @gertln I made a [small cleanup to the lexicon](https://github.com/NVIDIA/earth2studio/pull/939/changes/6681ab4df8bc7ec5786f00af60cfebb0923635d5) for better alignment with existing E2S variable conventions

NickGeneva:
Coverage: `earth2studio/models/dx/corrdiff_cosmo_era5.py             596    133    78%   208, 392, 399, 401, 403, 405, 576, 578, 581-582, 585, 689, 708, 714, 1179, 1194, 1218, 1284, 1316, 1364, 1421-1424, 1435, 1447, 1460, 1509-1517, 1609-1613, 1647-1890`

Need to do better. Merging. Can do some follow ups.

Thanks for the effort guys!

PR #943, opened 2026-06-30
Add Solar Model to StormScope

albertocarpentieri:
# Earth2Studio Pull Request
## Description
Adds `StormScopeNSRDB`, a StormScope-family diagnostic that estimates surface
Global Horizontal Irradiance (GHI) from StormScope GOES satellite imagery on the
3-km HRRR grid. It is a same-time estimator (given GOES at time *T* it estimates
GHI at *T*) and is designed to chain after `StormScopeGOES` to produce solar
irradiance nowcasts over the Continental US.
The model supports three inference modes:
- **Pure diffusion** — sample GHI entirely from noise (default when no regression
  model is present).
- **Regression only** — deterministic prediction from the regression model.
- **Regression + SDEdit** — warm-start diffusion from the regression prediction by
  injecting noise at `sdedit_sigma` and denoising from there.
It also supports clearness-index normalization (`state_norm_mode="clearness_index"`),
where GHI is normalized by insolation (`max(cos_zenith, 0) * solar_constant`) rather
than z-score, using the physicsnemo 1995 orbital insolation model to match training.
### What's included
- **`earth2studio/models/px/stormscope.py`** — new `StormScopeNSRDB` class
  (sharing `StormScopeBase`), including `estimate_from_goes(...)` as the intended
  entry point when chaining `StormScopeGOES -> StormScopeNSRDB`, plus regression +
  SDEdit support and clearness-index normalization.
- **`earth2studio/models/nn/stormscope_util.py`** — `MaskedModel` denoiser wrapper
  that zeros outputs at invalid (off-grid / ocean) pixels after every sampling step
  while preserving the `round_sigma` interface used by the EDM sampler.
- **`earth2studio/models/px/__init__.py`** — export `StormScopeNSRDB`.
- **`examples/04_nowcasting/04_stormscope_solar_example.py`** — minimal example
  chaining `StormScopeGOES -> StormScopeNSRDB` to nowcast GHI from GOES-16 imagery
  and plot the result.
- **`test/models/px/test_stormscope.py`** — unit tests for `StormScopeNSRDB`
  (`estimate_from_goes` output shape/coords/masking and `output_coords` handshakes),
  skipped when `physicsnemo.utils.insolation` is unavailable.
- **`docs/modules/models_px.rst`** + **`CHANGELOG.md`** — documentation and changelog
  entries.
- **`earth2studio/statistics/fss.py`** — minor unrelated fix: cast
  `np.prod(...)` to `int` before `Tensor.view(...)` (happy to drop if preferred).
### Notes / follow-ups
- `StormScopeNSRDB` currently has no `load_default_package()`; weights are loaded
  via an explicit `Package(...)`. A default public Hugging Face package can be wired
  up once the official weights repo is available.
## Checklist
## Dependencies
No new dependencies. `StormScopeNSRDB` relies on packages already required by the
`stormscope` extra (e.g. `physicsnemo` for the insolation model, `earth2grid`,
`scipy`).


earth2studio/models/nn/stormscope_util.py
pzharrington, line 1: Note this file (and the whole `DropInDiT` approach has been dropped with the integration of the 3km / 10 min  StormScope models -- you should update this branch and migrate `MaskedModel` to live next to the model wrapper

earth2studio/models/px/stormscope.py
pzharrington, line 1: I think a separate file / wrapper for the NSRDB model might be better than adding it into this one, but maybe update the branch first and see how it looks with the newer stormscope base class/models. If in a separate file, can still import and subclass the base stormscope wrapper class if it makes sense and if a lot of the code would get reused.

examples/04_nowcasting/04_stormscope_solar_example.py
CharlelieLrt, line 1: I don't think we want an example for this new wrapper. @NickGeneva to conform, but I think we've had this discussion before for another wrapper and the conclusion was "no example". However the Pr discussion section should have a code snippet + images/figures to help validate the model wrapper.

earth2studio/models/px/stormscope.py
```diff
@@ -50,11 +51,32 @@
 
 try:
     from physicsnemo import Module
+    from physicsnemo.utils.insolation import insolation as pnm_insolation
     from physicsnemo.utils.zenith_angle import cos_zenith_angle
```
CharlelieLrt, line 53: This is using none of the physicsnemo diffusion module, which I am not approving at all

earth2studio/models/px/stormscope.py
```diff
@@ -2405,3 +2427,873 @@ def load_model(
...
+class StormScopeNSRDB(StormScopeBase):
+    """Solar radiation nowcasting model with StormScope-style interface.
+
+    Supports three inference modes:
+
+    1. **Pure diffusion** -- sample entirely from noise (default when no regression
+       model is present).
+    2. **Regression only** -- deterministic prediction from the regression model.
+    3. **Regression + SDEdit** -- warm-start diffusion from the regression prediction
+       by injecting noise at ``sdedit_sigma`` and running the denoiser from there.
```
CharlelieLrt: I really don't like these 3 modes. It's immense complexity, but is it really justified? What's the point do have a regression-only mode? And what is this SDEdit thing?

earth2studio/models/px/stormscope.py
```diff
@@ -2405,3 +2427,873 @@ def load_model(
...
+
+    When ``state_norm_mode="clearness_index"``, the GHI target is normalized by
+    dividing by ``(insolation + clearness_index_eps)`` rather than by z-score.
+    Insolation is ``max(cos_zenith, 0) * solar_constant``.
+
+    Parameters
+    ----------
+    diffusion_model : torch.nn.Module
```
CharlelieLrt: So it's a torch Module. not even a physicsnemo Module? Is it even using physicsnemo at any point?

earth2studio/models/px/stormscope.py
```diff
@@ -2405,3 +2427,873 @@ def load_model(
...
+        x_coords: np.ndarray | None = None,
+        lead_time_steps: int = 0,
+        bootstrap_lead_time_label: int = 0,
+        autoregressive_lead_time_label: int = 1,
+        sdedit_sigma: float | None = None,
+        mask_regression: bool = False,
+        state_norm_mode: str = "standard",
+        clearness_index_eps: float = 10.0,
+        solar_constant: float = 1361.0,
+        disable_norm: bool = False,
```
CharlelieLrt: The immense over-complexity is very clear in this endless list of arguments

earth2studio/models/px/stormscope.py
```diff
@@ -2405,3 +2427,873 @@ def load_model(
...
+            sigma_max_sde = self._sigma_max
+
+        # Use a single-stage spec covering the active sampling range, then wrap
+        # the model with _mask_denoiser_outputs so invalid pixels (ocean, out-of-domain)
+        # are zeroed after EVERY denoising step.  Order matters: _mask_denoiser_outputs
+        # must nest INSIDE _use_forecast_checkpoint or the wrapper is silently dropped.
+        with self._use_forecast_checkpoint(
+            self.diffusion_model, sigma_min_sde, sigma_max_sde,
+        ):
+            with self._mask_denoiser_outputs():
```
CharlelieLrt: It would be better if you could get rid of these two context managers. It just seems unnecessarily over complicated IMO

CharlelieLrt (changes requested):
Left a few comments. Main things:
- Relies entirely on legacy APIs from physicnemo, that's a clear no-go for me
- Over-engineered
- No need for an example

My suggestion it's to create a wrapper class from scratch (don't inherit the StormScope one), because I am in any case not sure how much code you are really saving by composing the classes here. Also, keep the parameters from the model package that are absolutely necessary. You should have an idea of what works best, so we should stick to that in this wrapper. This will help simplify a lot of things.

I'll review again after these changes are implemented.

earth2studio/models/px/stormscope.py
albertocarpentieri: Fixed

earth2studio/models/nn/stormscope_util.py
albertocarpentieri: fixed

earth2studio/models/px/stormscope.py
albertocarpentieri: I updated the PR, I believe that since this model works only on stormscope goes it is appropriate here, but also open to move it to its own py script

examples/04_nowcasting/04_stormscope_solar_example.py
albertocarpentieri: I added the nsrdb model in the present example and removed the new one

earth2studio/models/px/stormscope.py
albertocarpentieri: Simplified to one strategy only

earth2studio/models/px/stormscope.py
albertocarpentieri: Fixed, I kept only the necessary arguments

earth2studio/models/px/stormscope.py
albertocarpentieri: This is more a global dependency for the full stormscope model, maybe for a future update of the model? I do not think it is part of this PR

examples/04_nowcasting/04_stormscope_solar_example.py
CharlelieLrt: I am just unsure if this is okay to do that. Please check with @NickGeneva

earth2studio/models/px/stormscope.py
```diff
@@ -2405,3 +2422,399 @@ def load_model(
...
+
+        # Regression first guess (masked), then warm-started diffusion refinement.
+        reg = self._sanitize(self.regression_model(condition))
+        reg = reg * self.valid_mask.reshape(1, 1, h, w).to(dtype=reg.dtype)
+        noise = torch.randn(
+            (b * t, len(self.variables), h, w),
+            device=device,
+            dtype=self._SAMPLER_DTYPE,
+        )
+        x_init = reg / self.sigma_max + noise
```
CharlelieLrt: Are you initializing the latent `x_N` with a noised prediction from the regression model? Does it work better than starting from plain noise? Just trying to understand the rationale for having a separate regresion model

albertocarpentieri: Yes correct, i tested using the diffusion model only but it gave worse results. The rationale is that the regression model can make you skip the high sigma diffusion steps, running the diffusion model only for the final part. It's an effort to make the model faster and temporally coherent.

CharlelieLrt:
@albertocarpentieri add to the PR description a code snippets to generate some predictions using the wrapper, as well as the images it generates. This is going to be used for V&V later on.

earth2studio/models/px/stormscope.py
CharlelieLrt: My comment was that thsi PR uses a bunch of legacy sampling utilities from physicnemo instead of the redesigned diffusion module. But it seems this is just inherited from the parent stormscope class, so it's probably out of scope for this PR.
@pzharrington is there any plan to update the stormscope model wrapper to make it use the updated `physicnemo.diffusion`?

earth2studio/models/px/stormscope.py
```diff
@@ -2405,3 +2422,415 @@ def load_model(
...
+
+    Given a GOES image at time ``T`` this estimates surface Global Horizontal
+    Irradiance (GHI) at ``T`` on the model's 3-km grid. It runs a single, fixed
+    pipeline:
+
+    1. A deterministic regression network produces a first-guess GHI from the
+       GOES conditioning.
+    2. That guess is refined by a diffusion model: noise is injected at the
+       checkpoint's registry ``sigma_max`` and the EDM sampler denoises down to
+       the model's minimum sigma (a warm-started reverse diffusion).
```
CharlelieLrt: It's basically SDEdit, right? Maybe worth briefly mentioning there. Also mention benefits of doing something like that?

earth2studio/models/px/stormscope.py
```diff
@@ -2405,3 +2422,415 @@ def load_model(
...
+
+    GHI is modelled in *clearness-index* space (target divided by
+    ``insolation + eps``); outputs are converted back to physical W/m^2 using the
+    insolation at the observation time. Both networks are complete
+    ``physicsnemo`` checkpoints loaded via :py:meth:`StormScopeBase._load_checkpoints`
+    / :py:meth:`physicsnemo.Module.from_checkpoint`, and the shared base EDM
```
CharlelieLrt: AI slop. I don't think anyone cares about the fact these checkpoints are loaded with some _load_checkpoints. Those who care can read the code.

earth2studio/models/px/stormscope.py
```diff
@@ -2405,3 +2422,415 @@ def load_model(
...
+        conditioning_variables: np.ndarray | None = None,
+        invariants: torch.Tensor | None = None,
+        valid_mask: torch.Tensor | None = None,
+        conditioning_data_source: DataSource | ForecastSource | None = None,
+        input_times: np.ndarray = np.array([np.timedelta64(0, "m")]),
+        output_times: np.ndarray = np.array([np.timedelta64(10, "m")]),
+        y_coords: np.ndarray | None = None,
+        x_coords: np.ndarray | None = None,
+        input_interp_max_dist_km: float = 12.0,
+        amp: bool = True,
```
CharlelieLrt: Is this argument really necessary? Remove if possible. Hardcoding `amp = True` should be fine

earth2studio/models/px/stormscope.py
```diff
@@ -50,11 +50,28 @@
...
+    module as ``net(x, sigma, condition=...)``.
+    """
+
+    def __init__(self, model: nn.Module, mask: torch.Tensor):
+        super().__init__()
+        self.model = model
+        self.register_buffer("mask", mask)
+
+    def forward(self, *args: Any, **kwargs: Any) -> torch.Tensor:
+        return self.model(*args, **kwargs) * self.mask
```
CharlelieLrt: I am a little confused by this. The checkpoint was trained without a mask, but the mask is added at inference? I don't understand how this can work...
Even though it is indeed transparent to the EDM sampler, it is not transparent for the model. In training it is trained to predict clean data `x_0` from full image and at inference we are making it predict clean data x0 from partially masked images. How does this work?

CharlelieLrt (changes requested):

earth2studio/models/px/stormscope.py
pzharrington: @CharlelieLrt  the base class was updated when I updated the base stormscope checkpoints a few weeks ago. I think the original version of this PR had that legacy pattern in but it should be updated/consistent now. The update that I did a few weeks ago was to eliminate the auxiliary stormscope utils file that defined shims and preconditioner, now the checkpoints are genuine `EDMPreconditioner`s from `physicsnemo.diffusion`.

pzharrington: While it's true the sampler in the base class remains custom, there are no imports of legacy items from PNM. The custom sampler is there mainly for exact parity with the multi-expert sampling approach used by NVR. In theory one could rewrite the same functionality using more of the `physicsnemo.diffusion` tools without a ton of extra effort, but I considered that P1 when doing the work awhile back. Not in scope for this PR to address though, it's sort of cosmetic in that very few users will see (or likely care about) that bit of code in this model wrapper

CharlelieLrt: Ok, thanks for the clarifications.

earth2studio/models/px/stormscope.py
albertocarpentieri: The mask is used because the groupnorm is trained to get 'masked' input but after one denoising iteration, the next input does not have such 'masked' pixels. This creates a bias in the model's activations, so there's the need to mask them at every iteration.

NickGeneva:
Heres my minimal sanity check script hooking it up with stormscope GOES:

<details>

```python
"""Quick sanity-check script for StormScope NSRDB diagnostic model."""

import numpy as np
import torch

from earth2studio.data import GOES, fetch_data
from earth2studio.models.dx import StormScopeDxNSRDB
from earth2studio.models.px.stormscope import StormScopeBase, StormScopeGOES

device = torch.device("cuda" if torch.cuda.is_available() else "cpu")

# Load models
package = StormScopeBase.load_default_package()
goes_model = StormScopeGOES.load_model(package, conditioning_data_source=None, amp=True)
goes_model = goes_model.to(device).eval()

nsrdb_model = StormScopeDxNSRDB.load_model(StormScopeDxNSRDB.load_default_package())
nsrdb_model = nsrdb_model.to(device)

# Build interpolators
goes_lat, goes_lon = GOES.grid(satellite="goes16", scan_mode="C")
goes_model.build_input_interpolator(goes_lat, goes_lon)
nsrdb_model.build_input_interpolator(goes_lat, goes_lon)

# Fetch initial GOES data
in_coords = goes_model.input_coords()
x, x_coords = fetch_data(
    GOES(satellite="goes16", scan_mode="C"),
    time=[np.datetime64("2024-01-08T18:00")],
    variable=np.array(in_coords["variable"]),
    lead_time=in_coords["lead_time"],
    device=device,
)

# Add batch dimension
x = x.unsqueeze(0).to(dtype=torch.float32)
x_coords["batch"] = np.arange(1)
x_coords.move_to_end("batch", last=False)

# Run one GOES forecast step
y_goes, y_coords = goes_model(x, x_coords)

# Prepare input for diagnostic (remove lead_time)
ghi_coords = y_coords.copy()
del ghi_coords["lead_time"]
y_goes_input = y_goes.squeeze(2)

# Run NSRDB diagnostic
ghi, ghi_out_coords = nsrdb_model(y_goes_input, ghi_coords)
print(f"GHI shape: {ghi.shape}")
print(f"GHI coords: {list(ghi_out_coords.keys())}")
print(f"GHI min={ghi[~ghi.isnan()].min().item():.1f}, max={ghi[~ghi.isnan()].max().item():.1f} W/m^2")

# Plot
import cartopy.crs as ccrs

lat = nsrdb_model.latitudes.cpu().numpy()
lon = nsrdb_model.longitudes.cpu().numpy()
ghi_field = ghi[0, 0, 0, 0].detach().cpu().numpy()

# GOES cloud channel (abi13c — Clean IR 10.35um)
goes_lat_out = goes_model.latitudes.cpu().numpy()
goes_lon_out = goes_model.longitudes.cpu().numpy()
goes_ch = "abi13c"
goes_ch_idx = list(goes_model.variables).index(goes_ch)
goes_field = y_goes[0, 0, 0, goes_ch_idx].detach().cpu().numpy()

proj = ccrs.LambertConformal(
    central_longitude=262.5,
    central_latitude=38.5,
    standard_parallels=(38.5, 38.5),
    globe=ccrs.Globe(semimajor_axis=6371229, semiminor_axis=6371229),
)

fig, (ax1, ax2) = plt.subplots(
    1, 2, subplot_kw={"projection": proj}, figsize=(16, 6)
)

# Left: GOES IR cloud imagery
im1 = ax1.pcolormesh(
    goes_lon_out, goes_lat_out, goes_field,
    transform=ccrs.PlateCarree(), cmap="gray_r", shading="auto",
)
ax1.coastlines(color="black", linewidth=0.8)
ax1.add_feature(cfeature.STATES, edgecolor="black", linewidth=0.4)
plt.colorbar(im1, ax=ax1, label=f"GOES {goes_ch} [K]", orientation="horizontal", pad=0.05)
ax1.set_title(f"GOES {goes_ch} (Clean IR)")

# Right: GHI estimate
im2 = ax2.pcolormesh(
    lon, lat, ghi_field,
    transform=ccrs.PlateCarree(), cmap="inferno", shading="auto", vmin=0, vmax=800,
)
ax2.coastlines(color="white", linewidth=0.8)
ax2.add_feature(cfeature.STATES, edgecolor="white", linewidth=0.4)
plt.colorbar(im2, ax=ax2, label="GHI [W/m$^2$]", orientation="horizontal", pad=0.05)
ax2.set_title(f"StormScope NSRDB GHI")

fig.suptitle(f"StormScope GOES + NSRDB — {ghi_out_coords['time'][0]} UTC", fontsize=13)
plt.tight_layout()
plt.savefig("test_nsrdb_ghi.png", dpi=200)
print("Saved test_nsrdb_ghi.png")

```

</details>

test_nsrdb_ghi

NickGeneva (approved):

NickGeneva:
coverage is 99%

PR #948, opened 2026-07-01
Fix MPI variable ordering in ACE2 by making variable lists deterministic

amandasd:
In earth2studio/models/px/ace2.py (ACE2.__init__), the code builds the forcing and prognostic variable lists as:

self._forcing_vars_fme = list(self.stepper._input_only_names)
self._prog_vars_fme = list(self.stepper.prognostic_names)

These are created from Python sets, so their iteration order differs across MPI ranks due to hash randomization. As a result, prognostic.input_coords()["variable"] is different on different ranks, causing

ValueError: Coordinate systems for required dim variable are not the same

when running with multiple MPI ranks.

Replacing list(...) with sorted(...) makes the ordering deterministic:

self._forcing_vars_fme = sorted(self.stepper._input_only_names)
self._prog_vars_fme = sorted(self.stepper.prognostic_names)

After this change, the variable ordering is identical on all MPI ranks and the model runs correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed ACE2 distributed inference failures caused by nondeterministic variable ordering
+  across MPI ranks.
+
 ### Security
 
 ### Dependencies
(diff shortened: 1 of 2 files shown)


NickGeneva (approved):
LGTM! Thanks Amanda!

PR #955, opened 2026-07-07
Migrate Zarr data sources (ARCO, WB2, rx) to Obstore

negin513:
# Earth2Studio Pull Request

## Description

Migrates the Zarr-backed data sources (`ARCO`, WeatherBench 2 sources, and the `rx` prescriptive sources), plus `UCast` static-field loading, from fsspec/gcsfs to `obstore`-backed Zarr stores.

Resolves https://github.com/NVIDIA/physicsnemo-roadmap/issues/2675

### What changed

- Added `obstore_zarr_store(url, cache_storage=None, **kwargs)` in `earth2studio/data/utils.py`.
  - This is the single integration point for obstore-backed Zarr reads.
  - It builds a read-only `zarr.storage.ObjectStore` around `obstore.store.from_url()`.
  - URL-scheme dispatch supports `gs://`, `s3://`, and `file://`, keeping future migrations on the same path.

- Added `LocalCachingStore`, a Zarr `WrapperStore` that provides whole-object on-disk caching through Zarr's `LocalStore`.
  - This replaces the `AsyncCachingFileSystem` role for the migrated Zarr sources.
  - Cache entries are stored under URL-hashed subdirectories, so multiple stores can safely share the same cache root.

- Migrated the following sources to use `obstore_zarr_store()`:
  - `ARCO`, including pressure/surface stores, model-level stores, and `available()`
  - `_WB2Base`, covering all WeatherBench 2 sources
  - `ARCORxBase`, including `LandSeaMask` and related `rx` sources
  - `UCast.load_model` static-condition loading

- Removed source-specific fsspec/gcsfs wiring from the migrated paths:
  - gcsfs config handling
  - MSC preference logic
  - event-loop workaround
  - `AsyncCachingFileSystem` wrapping

### Behavior notes for reviewers

- Outputs are unchanged. Byte-identical checksums were verified against `main` for ARCO, WB2, `rx`, and the `UCast` static-condition tensor.

- Multi-Storage Client support is removed entirely because it has no remaining consumers after this migration:
  - `get_msc_filesystem`
  - `AsyncCachingFileSystem`
  - `msc_config.yaml`
  - `EARTH2STUDIO_DISABLE_MSC`

- The new obstore cache layout is not compatible with old fsspec cache directories. Existing stale fsspec caches are ignored; they are not reused or modified.

### Performance

Measured against pre-PR `main` (`f193bc8`) with runs interleaved in time. Tests used public GCS buckets. Cache was disabled for store-open tests and enabled for cached-fetch tests.

| Workload | main, fsspec | this PR, obstore |
| --- | ---: | ---: |
| ARCO store open, median of 3 | 0.78 s | 0.69 s |
| WB2 store open, median of 3 | 0.25 s | 0.22 s |
| ARCO cached fetch, cold, 2 rounds | 7.3 / 5.7 s | 5.8 / 4.5 s, ~20% faster |
| ARCO cached fetch, warm rerun | 3.3 / 4.9 s | 2.8 / 4.6 s, parity, disk-bound |
| 512 concurrent 64 KB range GETs, warm | 450–470 req/s | 990–1190 req/s, ~2.4x |

Single large-chunk fetches remain bandwidth-bound and are effectively at parity. The gains are concentrated in request-heavy access patterns, especially high-concurrency small reads and cold cached fetches. This is the pattern expected to matter most for the upcoming GRIB byte-range migration, tracked in roadmap #2747.

### Testing

- Added network-free unit tests for `obstore_zarr_store()` and `LocalCachingStore` using local `file://` stores.
- Full ARCO/WB2/`rx` network suites pass: 66 tests.
- UCast unit tests pass.
- `black`, `ruff`, `mypy`, and `pre-commit` are clean.

## Checklist

- [x] Assess and address Greptile feedback. Greptile is an AI code review bot for guidance; addressing every suggestion is not required.

## Dependencies

No new dependencies.

`obstore>=0.8` and `zarr>=3.1.0` are already required. `obstore` was added during the UFS observation migration in #913.

`gcsfs` remains a dependency for the HRRR `source="google"` path until the GRIB byte-range migration is complete. That follow-on work is tracked in roadmap #2747.


earth2studio/data/arco.py
```diff
@@ -105,49 +100,32 @@ def __init__(
...
-            "cache_timeout": -1,
-            "token": "anon",  # noqa: S106 # nosec B106
-            "access": "read_only",
-            "block_size": 8**20,
-            "asynchronous": True,
-            "skip_instance_cache": True,
-        }
-
-        # Try to use Multi-Storage Client if available, otherwise fallback to gcsfs
-        MSCFileSystem = get_msc_filesystem()
```
NickGeneva, line 126: This can get removed from the toml now right?

earth2studio/data/utils.py
```diff
@@ -778,456 +767,131 @@ async def cancellable_to_thread(
...
 
+    def __init__(self, store: zarr.abc.store.Store, cache_storage: str) -> None:
+        super().__init__(store)
+        self._cache = zarr.storage.LocalStore(cache_storage)
+        # Per-key locks serialise concurrent fills of the same key. Held only
+        # across a cache miss, so distinct chunks still fetch concurrently.
+        self._fetch_locks: dict[str, asyncio.Lock] = {}
 
-@typing.no_type_check
-class AsyncCachingFileSystem(AsyncFileSystem):
```
NickGeneva, line 811: nice : )

earth2studio/data/utils.py
```diff
@@ -804,6 +807,85 @@ def get_msc_filesystem() -> filesystem | None:
...
+        byte_range: Any | None = None,
+    ) -> Any | None:
+        if byte_range is not None:
+            return await self._store.get(key, prototype, byte_range)
+        cached = await self._cache.get(key, prototype)
+        if cached is not None:
+            return cached
+        value = await self._store.get(key, prototype)
+        if value is not None:
+            await self._cache.set(key, value)
```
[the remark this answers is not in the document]
negin513: fixed in f6ec68b

earth2studio/data/arco.py
negin513: Good catch! I dropped multi-storage-client from the data optional extra (00743cc). 
 
 Still being used for serve server? Not sure what is that?

negin513: is that like a FASTAPI server?

earth2studio/data/arco.py
```diff
@@ -105,49 +100,32 @@ def __init__(
...
-        # at the end of the script
-        fs._loop = asyncio.get_event_loop()
+        cache_storage = self.cache if self._cache else None
+        zstore = obstore_zarr_store(
+            f"gs:/{self.ARCO_PATH}",
+            cache_storage=cache_storage,
+            skip_signature=True,
+        )
+        ml_zstore = obstore_zarr_store(
+            f"gs:/{self.ARCO_ML_PATH}",
```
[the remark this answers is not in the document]
negin513: Good bot! Applied your suggestion. ;-)

negin513:
/ok to test [91fe49f](https://github.com/NVIDIA/earth2studio/pull/955/commits/91fe49fbf15c60b9ace818e246e75ea89d5aab1e)

negin513:
For future @negin513

Here is the relevant docs page: https://zarr.readthedocs.io/en/stable/user-guide/storage/

``` 
AI-created:
 ┌─────────────────────────────────────────────────────────────┐
 │  DATA SOURCE            ARCO / WB2 / rx / UCast               │
 │                         zarr.open(zstore, mode="r")          │
 └──────────────────────────────┬──────────────────────────────┘
                                │ get(key)
 ┌──────────────────────────────▼──────────────────────────────┐
 │  LocalCachingStore           (only when cache=True)          │
 │  zarr WrapperStore                                           │
 │  • chunk hit  → serve from disk          ┌────────────────┐  │
 │  • chunk miss → fetch, then fill  ◄──────►  LOCAL DISK     │  │
 │  • metadata   → always refetch           │  $E2S_CACHE/    │  │
 │  • ranged     → pass through             │  <url-hash>/    │  │
 │  • per-key lock, best-effort write       └────────────────┘  │
 └──────────────────────────────┬──────────────────────────────┘
                                │ miss / metadata / ranged
 ┌──────────────────────────────▼──────────────────────────────┐
 │  zarr.storage.ObjectStore    (read_only=True)                │
 │  adapter: Zarr async store API  ↔  obstore API               │
 └──────────────────────────────┬──────────────────────────────┘
 ┌──────────────────────────────▼──────────────────────────────┐
 │  obstore.store.from_url(url, **kwargs)      ← factory         │
 │      gs://  → GCSStore   ◄── this PR (ARCO/WB2/rx, anon)      │
 │      s3://  → S3Store        file:// → LocalStore            │
 │  PyO3 binding → Rust object_store crate (async HTTP, no GIL) │
 └──────────────────────────────┬──────────────────────────────┘
                                │ HTTPS GET
                     ╔══════════▼══════════╗
                     ║   GCS  (public)     ║
                     ║  arco-era5 / wb2    ║
                     ╚═════════════════════╝
```

## Why did I chose `LocalStore` not `MemoryStore` for `cache=True`? 
`cache=True` meant a cache that survives the process, the typical pattern is re-running a script over the same data. Within one run each chunk is only read once, so a RAM cache would have nothing to hit and would just fight torch for memory. Warm disk reads come out of the OS page cache anyways...

⚠️ Future @negin513 / @NickGeneva : if we ever have a long-lived process re-reading the same chunks over and over (a serving mode, or inference loops that keep going back to the same timesteps?), think about `MemoryStore` too... I cannot find it useful for now!

NickGeneva:
@greptile-apps

CHANGELOG.md
```diff
@@ -69,6 +74,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Automatic test skipping for missing optional dependencies via
   `pytest_ignore_collect` hook in `test/conftest.py`
 
+### Deprecated
+
+### Removed
+
+- Removed `AsyncCachingFileSystem` and `get_msc_filesystem` from
```
NickGeneva: Nit: I think we can remove this bullet since not user facing. Try to keep the change log more user targeted in e2s and less verbose . Blame / PRs are always present for developers.


E.g. for this PR good change bullets would be:

Changed:

- Zarr-reading data sources (`ARCO`, `WB2ERA5` and other WeatherBench 2, etc) now read via obstore-backed zarr stores instead of fsspec

Dependencies:

- Removed multi-storage-client from data group, succeeded with obstore

earth2studio/data/wb2.py
```diff
@@ -69,32 +67,12 @@ def __init__(
...
-
-        zstore = zarr.storage.FsspecStore(
-            fs,
-            path=f"/weatherbench2/datasets/{self._product}/{self._zarr_store_name}",
+        """Async initialization of zarr group"""
+        store_path = f"/weatherbench2/datasets/{self._product}/{self._zarr_store_name}"
+        zstore = obstore_zarr_store(
+            f"gs://{store_path.lstrip('/')}",
+            cache_storage=self.cache if self._cache else None,
+            skip_signature=True,
```
NickGeneva: Whats `skip_signature=True`?

earth2studio/data/utils.py
```diff
@@ -778,456 +767,162 @@ async def cancellable_to_thread(
...
+        self._cache = zarr.storage.LocalStore(cache_storage)
+        # Per-key locks serialise concurrent fills of the same key. Held only
+        # across a cache miss, so distinct chunks still fetch concurrently.
+        self._fetch_locks: dict[str, asyncio.Lock] = {}
 
+    async def get(
+        self,
+        key: str,
+        prototype: Any,
+        byte_range: Any | None = None,
```
NickGeneva: Can this type be any more specific? What is a byte_range object? Tuple[int, int]? Not familiar

earth2studio/data/utils.py
```diff
@@ -778,456 +767,162 @@ async def cancellable_to_thread(
...
+        Local cache directory. If provided, whole-object reads are cached to
+        a URL-specific sub-directory via :class:`LocalCachingStore`, by
+        default None (no caching)
+    credential_provider : Any | None, optional
+        An obstore credential provider for authenticated access. Takes
+        precedence over ``auth_token``, by default None
+    auth_token : str | None, optional
+        Bearer token sent as an ``Authorization`` header for authenticated
+        access, by default None
+    **store_kwargs : Any
```
NickGeneva: Could the be converted into dict paramter rather than an open kwargs similar to
https://github.com/NVIDIA/earth2studio/blob/main/earth2studio/io/netcdf4.py#L60

I typically push us aware from having things like *args, **kwargs so API parameters stay explicit

earth2studio/data/utils.py
```diff
@@ -778,456 +767,162 @@ async def cancellable_to_thread(
...
+    # .zmetadata) — ARCO/WB2 are v2 despite the "-v3" store name.
+    _METADATA_SUFFIXES = ("zarr.json", ".zmetadata", ".zarray", ".zgroup", ".zattrs")
 
-T = TypeVar("T")
+    def __init__(self, store: zarr.abc.store.Store, cache_storage: str) -> None:
+        super().__init__(store)
+        self._cache = zarr.storage.LocalStore(cache_storage)
+        # Per-key locks serialise concurrent fills of the same key. Held only
+        # across a cache miss, so distinct chunks still fetch concurrently.
+        self._fetch_locks: dict[str, asyncio.Lock] = {}
```
NickGeneva: Should this be a singlton patter here / class variable so the log is shared between instances of the cache? Or maybe keys based on the cahce location as well?

If i create two instances of data sources, they would potentially clash still with each other I think since they would have different cache file system isntances?

earth2studio/models/px/ucast.py
```diff
@@ -794,10 +794,10 @@ def load_model(
                 logger.info(
                     "Loading U-CAST static fields from WeatherBench2 public ERA5 zarr"
                 )
-                import gcsfs
+                from earth2studio.data.utils import obstore_zarr_store
 
-                fs = gcsfs.GCSFileSystem(token="anon")  # noqa: S106
-                ds = xr.open_zarr(fs.get_mapper(UCAST_WB2_DATASET), zarr_format=2)
+                zstore = obstore_zarr_store(UCAST_WB2_DATASET, skip_signature=True)
```
NickGeneva: does this need to pass a `cache_storage`  maybe using the models cache location? ` Package.default_cache("ucast")`?

earth2studio/data/utils.py
```diff
@@ -778,456 +767,162 @@ async def cancellable_to_thread(
...
+        lock = self._fetch_locks.get(key)
+        if lock is None:
+            lock = self._fetch_locks[key] = asyncio.Lock()
 
-@typing.no_type_check
-class AsyncCachingFileSystem(AsyncFileSystem):
-    """Async locally caching filesystem, layer over any other FS for use with zarr 3.0.
-    Presently extremely limited, much is just copied from the WholeFileCache
+        try:
+            async with lock:
```
NickGeneva: If theres two workers that hit the lock at the same time and get locked, one will wait, other will proceed... but when the first one finishes it will got to the finally and delete the cache entry.

Then the second will basically do a repeat of the first workers job right since the first just deleted its cache entry? I'm not sure this works correctly when theres two immediately concurrent requests where the lock matters.

NickGeneva: Maybe im not following this right.

NickGeneva (commented):
Overall looks great, mainly some questions / comments around the cache / locking 👍

CHANGELOG.md
negin513: Fair point, thanks for pointing it out.  Fixed it in: [18847f5](https://github.com/NVIDIA/earth2studio/pull/955/commits/18847f560a36f71ac6f00764f93d00ae0576c611)

earth2studio/data/utils.py
negin513: Let me think more about this. I went a bit back and forth on how to address this.

earth2studio/data/wb2.py
negin513: It's obstore's flag for unsigned/anonymous requests, needed to read public buckets without GCP credentials. 
 
Equivalent of gcsfs's token="anon" in the old path. I think it will fail without it.

negin513: confirmed, it will fail without using the `skip_signature=True`...

earth2studio/data/utils.py
negin513: Typed it to match `zarr.abc.store.Store.get: byte_range: ByteRequest | None` 

https://zarr.readthedocs.io/en/stable/api/zarr/abc/store/index.html#zarr.abc.store.ByteRequest
https://zarr.readthedocs.io/en/stable/api/zarr/abc/store/index.html#zarr.abc.store.RangeByteRequest

But I am using it here for partial reads.

earth2studio/models/px/ucast.py
negin513: I think it is already cached at a level above.

negin513: this store is only read once on a cold cache. I can potentially cache it at this level instead and remove the above. Not sure why we cache it one level above?

earth2studio/data/utils.py
negin513: Yup! It's super confusing. TBH, [I asked AI to draw](https://github.com/NVIDIA/earth2studio/pull/955#issuecomment-4931325525) it because of it. 

the finally only drops the asyncio.Lock from _fetch_locks & nothing is removed from self._cache itself. and worker B still holds a local reference to lock. 

That said, if it is confusing both of us; it is bad. Let me see what I can do to simplify this.

earth2studio/models/px/ucast.py
negin513: Okay, added cache_storage=package.cache. Still the static_condition.pt just is where the computed static field tensors are bing read from and this would only get hit when the file is missing.

earth2studio/data/utils.py
negin513: addressed in [74e9839](https://github.com/NVIDIA/earth2studio/pull/955/commits/74e9839b7523beb7000e2433b222d29bca585a2c).

earth2studio/data/utils.py
negin513: Done; moved `_fetch_locks` to a class variable keyed by `{cache_dir}::{key}`, so instances sharing a cache dir now dedupe each other. Added a cross-instance test.

earth2studio/data/utils.py
negin513: Simplified in the latest commit: the entry is now popped while the lock is still held, so the "is it still ours" guard is gone. The invariant is one line — nobody can delete a lock they don't hold; late waiters holding a reference just acquire, re-check the cache, and hit it.

earth2studio/data/utils.py
negin513: Good idea; done.

negin513: Kept the values as Any since valid keys vary per store backend ;; I wish they adopted the same args....
(obstore types them the same way).

negin513:
/ok to test [af6a3cc](https://github.com/NVIDIA/earth2studio/pull/955/commits/af6a3cce78f359091ae81ff671ba102dbe037560)

NickGeneva (approved):
caching looks good to me now, approving!

NickGeneva:
Force merging for some reason ecmwf tests are failing, need to look into that

FAILED test/data/test_ecmwf.py::test_ifs_time_available[msl-time1-IFS_ENS] - Failed: Timeout (>30.0s) from pytest-timeout.
FAILED test/data/test_ecmwf.py::test_ifs_time_available[msl-time1-IFS] - Failed: Timeout (>30.0s) from pytest-timeout.

PR #959, opened 2026-07-08
MTG FCI fixes / improvements

jleinonen:
# Earth2Studio Pull Request

## Description

Refactoring of `MeteosatFCI` for fixes/improvements:

1. Added 4 channels that were missing from the original version.
1. Implement special HDR decoding of `ir_38` channel.
1. Add support for high-resolution (`HRFI`) products; downloaded automatically if available for the requested resolution.
1. Make north-south flipping optional; internal processing is done in FCI native coordinates and flipping done at the last minute if needed.
1. Update lexicon: channel numbers were incomplete and in any case misleading since EUMETSAT does not use channel numbers but rather descriptors such as `"vis_06"`. Lexicon now names variables as `"fci_vis_06"` etc.
1. Added pixel_bbox parameter to provide the pixel bounding box directly rather than computed from the lat-lon box.
1. EUMETSAT credentials can now be loaded from a config file (`~/.eumdac/credentials` or `$EUMDAC_CONFIG_DIR/credentials`) when environment variables are not set.
1. Fix coordinate handling:
  * Old version always returned the `"x"` and `"y"` coordinates as integer arrays from 0 to `nrows/ncols` regardless of bounding box.
  * Source of truth for coordinates is now in `MeteosatFCI.FCI_X` and ``MeteosatFCI.FCI_Y``.
  * Coordinate conversion from integer arrays to radians now more closely follows FCI data files.
1. Update tests for refactored data source.

## Checklist


earth2studio/lexicon/meteosat.py
```diff
@@ -37,22 +37,25 @@ class MeteosatFCILexicon(metaclass=LexiconType):
...
-        "fci11": ("ir_123", lambda x: x),  # IR 12.30 µm
-        "fci12": ("ir_133", lambda x: x),  # IR 13.30 µm
+        "fci_nir_13": ("nir_13", lambda x: x),  # NIR 1.380 µm (1 km)
+        "fci_nir_16": ("nir_16", lambda x: x),  # NIR 1.610 µm (1 km)
+        "fci_nir_22": ("nir_22", lambda x: x),  # NIR 2.250 µm (0.5 km / 1 km)
+        # IR bands
+        "fci_ir_38": ("ir_38", lambda x: x),  # IR 3.800 µm (1 km / 2 km)
+        "fci_wv_63": ("wv_63", lambda x: x),  # WV 6.300 µm (2 km)
+        "fci_wv_73": ("wv_73", lambda x: x),  # WV 7.350 µm (2 km)
+        "fci_ir_87": ("ir_87", lambda x: x),  # IR 8.700 µm (2 km)
```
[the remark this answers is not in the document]
jleinonen: WV bands are IR bands (they are still in the infrared wavelengths) - just specially named.

earth2studio/data/meteosat_fci.py
```diff
@@ -205,30 +209,33 @@ def _compute_pixel_roi(
 
     Parameters
     ----------
-    lat_lon_bbox : tuple[float, float, float, float]
-        ``(lat_min, lon_min, lat_max, lon_max)`` in degrees.
+    lat_lon_bbox : tuple[tuple[float, float], tuple[float, float]]
+        ``((lat_min, lat_max), (lon_min, lon_max))`` in degrees.
```
NickGeneva, line 212: Not sure I agree with this change, the reason for the previous lat_min, lon_min, lat_max, lon_max is this creates two points of the bbox rectangle and is similar to ordering seen in segmentation:

https://docs.pytorch.org/vision/main/generated/torchvision.utils.draw_bounding_boxes.html#draw-bounding-boxes

Additionally this was done for consistency with cbottle which has a similar bbox
https://github.com/NVIDIA/earth2studio/blob/main/earth2studio/models/dx/cbottle_sr.py#L120

Is there a reason for this this change?

earth2studio/data/meteosat_fci.py
```diff
@@ -313,20 +337,47 @@ class MeteosatFCI:
...
+    FCI_X: dict[FCIResolution, np.ndarray] = {
+        res: np.arange(1, _GRID_SIZE[res][1] + 1) * _X_SCALE[res] + _X_OFFSET[res]
+        for res in _GRID_SCALE
+    }
 
     def __init__(
         self,
-        resolution: str = "2km",
-        lat_lon_bbox: tuple[float, float, float, float] | None = None,
+        resolution: FCIResolution = "2km",
```
NickGeneva: Consider just making this Literal["aaa", "bbb", etc], the reason is explicitly spelling it out makes it self documenting in the rendered API docs which can be useful if there user needs to know what the options are.

earth2studio/data/meteosat_fci.py
```diff
@@ -335,17 +386,29 @@ def __init__(
...
-        self._pixel_roi: tuple[int, int, int, int] | None = None
 
         # Credentials from environment variables
         self._consumer_key = os.environ.get("EUMETSAT_CONSUMER_KEY", "")
         self._consumer_secret = os.environ.get("EUMETSAT_CONSUMER_SECRET", "")
         if not self._consumer_key or not self._consumer_secret:
-            logger.warning(
-                "EUMETSAT_CONSUMER_KEY and/or EUMETSAT_CONSUMER_SECRET not set. "
-                "Data fetching will fail."
+            config_env = os.environ.get("EUMDAC_CONFIG_DIR")
```
NickGeneva: Can you link me the docs on this? Is this standard for EUMETSAT? I understand convience but unless their package automatically loads a credential file like say CDS does with `.cdsapirc`

we should avoid being in the business of telling people how to store their API keys out side of env variables. 

Is this a standard approach from EUMETSAT? If so then this is fine, although odd your reading it.

earth2studio/data/meteosat_fci.py
```diff
@@ -772,23 +812,31 @@ def _check_resolution_consistency(self, variable: list[str]) -> None:
...
+            ]
+            available_res = [res for res in available_res if res is not None]
+            if self._resolution not in available_res:
                 raise ValueError(
-                    f"Variable '{v}' (channel '{channel_name}') has native "
-                    f"resolution {native_res}, but data source is configured "
+                    f"Variable '{v}' (channel '{channel_name}') has available "
+                    f"resolutions {available_res}, but data source is configured "
                     f"for {self._resolution}. All requested variables must "
                     f"share the same resolution."
```
NickGeneva: nit: no need for f string here

earth2studio/lexicon/meteosat.py
```diff
@@ -37,22 +37,25 @@ class MeteosatFCILexicon(metaclass=LexiconType):
     """
 
     VOCAB: dict[str, tuple[str, Callable[[Any], Any]]] = {
-        # VIS bands (1 km resolution)
-        "fci01": ("vis_04", lambda x: x),  # VIS 0.44 µm
-        "fci02": ("vis_05", lambda x: x),  # VIS 0.51 µm
-        "fci03": ("vis_08", lambda x: x),  # VIS 0.86 µm
-        "fci04": ("vis_09", lambda x: x),  # VIS 0.91 µm
+        # VIS bands
+        "fci_vis_04": ("vis_04", lambda x: x),  # VIS 0.444 µm (1 km)
```
NickGeneva: make a note in the change log about this since this is breaking technically

earth2studio/data/meteosat_fci.py
```diff
@@ -694,10 +774,18 @@ def _read_channel(
                 offset = getattr(var, "add_offset", 0.0)
 
                 data = raw.astype(np.float64) * scale + offset
+                if channel_name in ("ir_38", "ir_38_hr"):
```
NickGeneva, line 780: > Implement special HDR decoding of ir_38 channel.

Can you provide some info on why this is to the comment or in your description. Maybe a link to a PDF or something for future reference.

Can be useful we have to come back to this to debug something

earth2studio/data/meteosat_fci.py
```diff
@@ -869,27 +917,29 @@ def available(
 
     @staticmethod
     def grid(
-        resolution: str = "2km",
+        resolution: FCIResolution = "2km",
```
NickGeneva: Same deal here, likely worth writing out Literal["a", "b", "c"] here so its just right where for the user and you dont need to repeat the options in the description of the paramter.

NickGeneva (commented):
Thanks Jussi for digging into the FCI data and correcting everything.

Just a few items, can you also create a small plotting script that can validate the data source is still producing correct results and post it in a comment with the rendered images? Very useful for sanity check + documenting

Thanks!

earth2studio/data/meteosat_fci.py
jleinonen: The reason for changing it in the first place was that the latlon bounding box and the pixel bounding box were inconsistently ordered - one had `(dim0_min, dim0_max, dim1_min, dim1_max)` and the other `(dim0_min, dim1_min, dim0_max, dim1_max)` which I found confusing (and becomes extra confusing now that the pixel bbox can be specified by the user).

I think the two boxes should have consistent ordering. I made them both use the nested-tuples convention because I find the "a point is a tuple; a box is a tuple of ranges" convention the most intuitive and less suspectible to this kind of ambiguity, but if you prefer a flat tuple I can revert to that as long as the ordering is consistent for both boxes.

earth2studio/data/meteosat_fci.py
jleinonen: EUMETSAT's own Python client CLI supports this environment variable and file: see lines 35 and 62 in https://gitlab.eumetsat.int/eumetlab/data-services/eumdac/-/blob/958bb1ea0753fd3f3af50317fbda4a7073a95027/eumdac/cli.py.

Although I don't find this actually documented anywhere except example code...

earth2studio/data/meteosat_fci.py
jleinonen: Okay, I thought a single point of definition would be useful but it's a good point about the docs. I removed the `FCIResolution` and `FCICollection` type aliases and replaced them with literals.

earth2studio/data/meteosat_fci.py
jleinonen: Replaced

earth2studio/data/meteosat_fci.py
jleinonen: Done

earth2studio/data/meteosat_fci.py
jleinonen: The `ir_38` channel special handling is described here: https://user.eumetsat.int/resources/user-guides/mtg-fci-level-1c-data-guide#ID-Radiance-encoding

> The IR 3.8 channel is a special case. Instead of 12 bit as for the other spectral channels, 13 bit are used to store the data. For IR3.8 the additional attributes valid_cold_range, warm_scale_factor and warm_add_offset are used to encode and compress the counts above 2^12-1 (4095) to cover the extended radiometric range. For counts below and equal to 4095, the same conversion with scale_factor and add_offset as for the other channels is used (Figure 22).

jleinonen:
> Thanks Jussi for digging into the FCI data and correcting everything.
> 
> Just a few items, can you also create a small plotting script that can validate the data source is still producing correct results and post it in a comment with the rendered images? Very useful for sanity check + documenting
> 
> Thanks!

Sure I can create this script. Where do you want it? I could be in examples but we don't have a directory for data source examples at the moment.

earth2studio/lexicon/meteosat.py
jleinonen: Updated the changelog

earth2studio/data/meteosat_fci.py
NickGeneva: Sweet thanks, just added some inline comments for this

NickGeneva:
@greptile-apps

earth2studio/data/meteosat_fci.py
NickGeneva: Nope, both work for me, if there was a inconsistency happy to change it. Thanks for finding this.

NickGeneva:
meteosat_fci_sanity_check


<details>

<summary>Plotting script</summary>

```python
import argparse
from datetime import datetime

import matplotlib

matplotlib.use("Agg")

from earth2studio.data import MeteosatFCI


def main() -> None:
    p = argparse.ArgumentParser()
    p.add_argument("--time", default="2025-03-01T12:00:00")
    p.add_argument("--output", default="meteosat_fci_sanity_check.png")
    p.add_argument(
        "--variables",
        nargs="+",
        default=["fci87ir", "fci63wv", "fci105ir"],
    )
    args = p.parse_args()

    bbox = ((35.0, 60.0), (-10.0, 30.0))
    ds = MeteosatFCI(
        resolution="2km",
        lat_lon_bbox=bbox,
        cache=True,
        verbose=True,
    )
    data = ds(datetime.fromisoformat(args.time), args.variables)
    fig, axes = plt.subplots(
        1, len(args.variables), figsize=(5 * len(args.variables), 4)
    )
    axes = [axes] if len(args.variables) == 1 else axes

    for ax, variable in zip(axes, args.variables):
        field = data.sel(variable=variable).isel(time=0)
        image = ax.imshow(field, origin="lower", cmap="viridis")
        ax.set_title(variable)
        ax.set_ylabel("y")
        fig.colorbar(image, ax=ax, shrink=0.8)

    fig.tight_layout()
    fig.savefig(args.output, dpi=160)
    print(f"Wrote {args.output}; data shape={data.shape}")


if __name__ == "__main__":
    main()

```

</details>

NickGeneva (approved):

PR #960, opened 2026-07-09
Add EarthMover ERA5 data source

### LINKED ISSUE #928: 🚀[FEA]: Earthmover Arraylake / Marketplace data sources

New Feature

Medium

A growing catalog of analysis-ready, cloud-optimized weather/climate datasets are available via the [Earthmover Data Marketplace](https://app.earthmover.io/marketplace). We expect the volume and diversity of data to grow substantially over time. Connecting the Earthmover Data Marketplace to Earth2Studio would immediately unlock all of these datasets for E2S users.

NickGeneva:
## Summary

Add `EarthMoverERA5` (`DataSource`) for accessing the [Earthmover ERA5 0.25 degree reanalysis](https://app.earthmover.io/marketplace/6a19bcfe9aa6e97720a2fad2) dataset hosted on the Earthmover Arraylake Marketplace.

This extends the EarthMover data-source family added for the BrightBand IFS sources with a reanalysis source that opens the ERA5 `single/spatial` and `pressure/spatial` groups and resolves Earth2Studio variable IDs through a dedicated ERA5 lexicon.

Closes: https://github.com/NVIDIA/earth2studio/issues/928

### New data source
- **`EarthMoverERA5`** — Gridded reanalysis source (implements `DataSource`) for hourly ERA5 data.
- Uses `EARTHMOVER_API_KEY` for Arraylake authentication, or accepts a pre-authenticated `arraylake.AsyncClient`.
- Derives the default subscription repo as `<EARTHMOVER_ORGANIZATION>/era5-subscription`, while still allowing an explicit `repo="org/repo"` override.
- Opens the spatially chunked ERA5 groups: `single/spatial` and `pressure/spatial`.

### Shared EarthMover base class
- Reuses `_EarthMoverBase` for Arraylake connection handling, metadata-driven variable resolution, coordinate normalization, time validation, and subscription/access error messages.
- Keeps variable resolution lexicon-specific so IFS and ERA5 can share the same source implementation without sharing the same variable inventory.
- Avoids thread offloading; data selection and xarray loading remain direct.

### Lexicon
- Adds `EarthMoverERA5Lexicon` with ERA5-specific variable metadata and modifiers.
- Covers 38 single-level Earth2Studio variables and 8 pressure-level variable families across 13 isobaric levels (`50, 100, 150, 200, 250, 300, 400, 500, 600, 700, 850, 925, 1000 hPa`).
- Adds missing base `E2STUDIO_VOCAB` entries needed for ERA5 single-level and pressure-level variables.

### Documentation and registration
- Registers `EarthMoverERA5` in `earth2studio.data` and `EarthMoverERA5Lexicon` in `earth2studio.lexicon`.
- Adds the source to the analysis data-source API docs.
- Updates `CHANGELOG.md`.

### Tests
- Extends EarthMover data-source tests for ERA5 constructor behavior, protocol compliance, repo derivation, explicit repo precedence, group selection, and fetch behavior.
- Adds dedicated ERA5 lexicon tests for Marketplace variable coverage, spec parsing, key resolution, and base vocab coverage.

## Data Source Details

| Property | Value |
|---|---|
| **Source type** | `DataSource` |
| **Remote store** | [Earthmover Arraylake Marketplace ERA5](https://app.earthmover.io/marketplace/6a19bcfe9aa6e97720a2fad2) |
| **Format** | Icechunk / Zarr v3 via Arraylake |
| **Spatial resolution** | 0.25 degree global latitude-longitude grid |
| **Temporal resolution** | Hourly |
| **Date range** | 1940-01-01 to 2025 in the public registry entry; subscription availability follows the Earthmover Marketplace repo |
| **Authentication** | Earthmover / Arraylake API key via `EARTHMOVER_API_KEY` |
| **Repository default** | `<EARTHMOVER_ORGANIZATION>/era5-subscription` |
| **Groups used** | `single/spatial`, `pressure/spatial` |

## Data Licensing

> **License**: Creative Commons Attribution 4.0 International (CC-BY-4.0)  
> **URL**: https://creativecommons.org/licenses/by/4.0/

Dataset details are also listed in the [AWS Registry of Open Data entry for Icechunk ERA5](https://registry.opendata.aws/earthmover-era5/), which points to the same Earthmover Marketplace dataset browser.

## Dependencies Added

| Package | Version | License | License URL | Reason |
|---|---|---|---|---|
| `arraylake` | `>=1.0.0` | Apache-2.0 | https://github.com/earth-mover/arraylake | Access Earthmover Arraylake Marketplace repositories |
| `pcodec` | `>=1.0.2` | Apache-2.0 | https://github.com/pcodec/pcodec | Decode Pcodec-compressed Zarr/Icechunk ERA5 arrays |

### Single-level variables

`blh`, `cape`, `cp`, `d2m`, `fdir`, `fg10m`, `fsr`, `hcc`, `ie`, `lcc`, `lsp`, `mcc`, `msl`, `sd`, `sf`, `skt`, `slhf`, `sp`, `ssr`, `ssrd`, `sst`, `stl1`, `stl2`, `stl3`, `stl4`, `swvl1`, `t2m`, `tcc`, `tcw`, `tcwv`, `tisr`, `tp`, `tsr`, `u10m`, `u100m`, `v10m`, `v100m`, `zust`

### Pressure-level variable families

`pv`, `q`, `r`, `t`, `u`, `v`, `w`, `z` at `50`, `100`, `150`, `200`, `250`, `300`, `400`, `500`, `600`, `700`, `850`, `925`, and `1000 hPa`.

## Validation

- Focused EarthMover data-source tests pass for sync and async fetch paths, repo configuration, variable resolution, and error handling.
- Dedicated EarthMover lexicon tests cover ERA5 vocab/spec/key behavior.
- `make lint` passes, including ruff and mypy.
- `make license` passes.

## Checklist

- [x] The documentation is up to date.
- [x] The CHANGELOG.md is up to date.
- [ ] Assess and address Greptile feedback.


NickGeneva:
### EarthMover ERA5 Sanity-Check Validation

This sanity check compares `EarthMoverERA5` against the existing `ARCO` ERA5 data source for the same timestamp and variables. It is intentionally kept out of the committed files.

#### Variable Coverage Table

| Variable | Type | Validation performed |
|---|---|---|
| `t2m` | single-level | Source finite %, min/max, and EarthMover-minus-ARCO bias/MAE/RMSE/p95/maxabs |
| `msl` | single-level | Source finite %, min/max, and EarthMover-minus-ARCO bias/MAE/RMSE/p95/maxabs |
| `u10m` | single-level | Source finite %, min/max, and EarthMover-minus-ARCO bias/MAE/RMSE/p95/maxabs |
| `z500` | pressure-level | Source finite %, min/max, and EarthMover-minus-ARCO bias/MAE/RMSE/p95/maxabs |
| `q850` | pressure-level | Source finite %, min/max, and EarthMover-minus-ARCO bias/MAE/RMSE/p95/maxabs |

#### Data Validation Summary

- Default time: `2022-01-01T00:00:00`
- Region: global 0.25 degree latitude-longitude grid
- Reference source: `ARCO`
- New source: `EarthMoverERA5`
- Grid handling: verifies both sources use the same grid and refuses to interpolate if latitude/longitude differ
- Output plot: `earthmover_era5_arco_sanity_check.png`

#### Key Findings Checked By The Script

- Both sources return finite data for the requested variables.
- EarthMover and ARCO latitude/longitude grids match before comparison.
- Difference statistics are printed per variable.
- Map panels are written for EarthMover, ARCO, and EarthMover-minus-ARCO.

#### Sanity-Check Plot



<details>

<summary>Full validation script</summary>

```python

import argparse
import os
from collections.abc import Sequence
from datetime import datetime, timezone
from pathlib import Path

import numpy as np
import xarray as xr

from earth2studio.data import ARCO, EarthMoverERA5

DEFAULT_TIMES = ("2022-01-01T00:00:00",)
DEFAULT_VARIABLES = ("t2m", "msl", "u10m", "z500", "q850")


def parse_csv(value: str) -> list[str]:
    """Parse a comma-separated CLI value."""
    return [item.strip() for item in value.split(",") if item.strip()]


def parse_time(value: str) -> datetime:
    """Parse an ISO timestamp and return a naive UTC datetime."""
    parsed = datetime.fromisoformat(value.replace("Z", "+00:00"))
    if parsed.tzinfo is not None:
        parsed = parsed.astimezone(timezone.utc).replace(tzinfo=None)
    return parsed


def parse_times(values: Sequence[str]) -> list[datetime]:
    """Parse a sequence of ISO timestamps."""
    return [parse_time(value) for value in values]


def normalize_grid(data: xr.DataArray) -> xr.DataArray:
    """Normalize latitude and longitude coordinates for exact grid comparison."""
    rename = {}
    if "latitude" in data.coords and "lat" not in data.coords:
        rename["latitude"] = "lat"
    if "longitude" in data.coords and "lon" not in data.coords:
        rename["longitude"] = "lon"
    if rename:
        data = data.rename(rename)

    if "lat" not in data.coords or "lon" not in data.coords:
        raise ValueError(f"Expected lat/lon coordinates, got {list(data.coords)}")

    data = data.sortby("lat")
    lon = np.mod(np.asarray(data.lon.values, dtype=float), 360.0)
    data = data.assign_coords(lon=lon)

    _, unique_indexes = np.unique(lon, return_index=True)
    if unique_indexes.size != lon.size:
        data = data.isel(lon=np.sort(unique_indexes))

    return data.sortby("lon")


def align_same_grid(
    earthmover: xr.DataArray,
    arco: xr.DataArray,
    atol: float = 1.0e-8,
) -> tuple[xr.DataArray, xr.DataArray, float, float]:
    """Align two data arrays after verifying they use the same grid."""
    earthmover = normalize_grid(earthmover)
    arco = normalize_grid(arco)

    if earthmover.sizes["lat"] != arco.sizes["lat"]:
        raise ValueError(
            f"Latitude size differs: EarthMover={earthmover.sizes['lat']} "
            f"ARCO={arco.sizes['lat']}"
        )
    if earthmover.sizes["lon"] != arco.sizes["lon"]:
        raise ValueError(
            f"Longitude size differs: EarthMover={earthmover.sizes['lon']} "
            f"ARCO={arco.sizes['lon']}"
        )

    lat_delta = float(
        np.max(np.abs(np.asarray(earthmover.lat.values) - np.asarray(arco.lat.values)))
    )
    lon_delta = float(
        np.max(np.abs(np.asarray(earthmover.lon.values) - np.asarray(arco.lon.values)))
    )
    if lat_delta > atol or lon_delta > atol:
        raise ValueError(
            "EarthMover and ARCO grids differ. Refusing to interpolate in this "
            f"sanity check. max(|lat diff|)={lat_delta:g}, "
            f"max(|lon diff|)={lon_delta:g}"
        )

    # Reuse the same coordinate objects so xarray can align exactly even if the
    # original floating-point representations differ at harmless precision.
    earthmover = earthmover.assign_coords(lat=arco.lat.values, lon=arco.lon.values)
    earthmover, arco = xr.align(earthmover, arco, join="exact")
    return earthmover, arco, lat_delta, lon_delta


def finite_percentile(values: np.ndarray, percentiles: int | Sequence[int]):
    """Return percentiles over finite values only."""
    finite = values[np.isfinite(values)]
    if not finite.size:
        return None
    return np.nanpercentile(finite, percentiles)


def summarize_source(data: xr.DataArray, label: str) -> None:
    """Print compact source statistics."""
    print(f"\n{label}")
    print(f"dims={dict(data.sizes)}")
    for variable in data.coords["variable"].values:
        values = np.asarray(data.sel(variable=variable).values)
        finite = np.isfinite(values)
        pct = 100.0 * finite.sum() / finite.size if finite.size else 0.0
        vmin = float(np.nanmin(values)) if finite.any() else float("nan")
        vmax = float(np.nanmax(values)) if finite.any() else float("nan")
        print(f"{variable:<8} finite={pct:6.2f}% min={vmin:12.4g} max={vmax:12.4g}")


def summarize_difference(
    earthmover: xr.DataArray,
    arco: xr.DataArray,
) -> tuple[xr.DataArray, xr.DataArray, xr.DataArray]:
    """Print compact EarthMover-minus-ARCO comparison statistics."""
    earthmover, arco, lat_delta, lon_delta = align_same_grid(earthmover, arco)
    diff = earthmover - arco

    print("\nEarthMoverERA5 - ARCO")
    print(f"dims={dict(diff.sizes)}")
    print(f"grid max deltas: lat={lat_delta:g}, lon={lon_delta:g}")
    for variable in diff.coords["variable"].values:
        values = np.asarray(diff.sel(variable=variable).values)
        finite = np.isfinite(values)
        if not finite.any():
            print(f"{variable:<8} no finite comparison values")
            continue

        finite_values = values[finite]
        bias = float(np.nanmean(finite_values))
        mae = float(np.nanmean(np.abs(finite_values)))
        rmse = float(np.sqrt(np.nanmean(finite_values**2)))
        maxabs = float(np.nanmax(np.abs(finite_values)))
        p95abs = float(np.nanpercentile(np.abs(finite_values), 95))
        print(
            f"{variable:<8} bias={bias:12.4g} mae={mae:12.4g} "
            f"rmse={rmse:12.4g} p95abs={p95abs:12.4g} maxabs={maxabs:12.4g}"
        )
    return earthmover, arco, diff


def time_label(time: np.datetime64) -> str:
    """Return a compact timestamp label."""
    return np.datetime_as_string(np.datetime64(time), unit="h")


def plot_comparison(
    earthmover: xr.DataArray,
    arco: xr.DataArray,
    diff: xr.DataArray,
    output: Path,
) -> None:
    """Write EarthMover/ARCO/difference map panels."""
    import matplotlib

    matplotlib.use("Agg")

    try:
        import cartopy.crs as ccrs

        projection = ccrs.PlateCarree()
    except ImportError:
        ccrs = None
        cfeature = None
        projection = None

    variables = list(earthmover.coords["variable"].values)
    times = list(earthmover.coords["time"].values)
    panels = [(time, variable) for time in times for variable in variables]
    subplot_kw = {"projection": projection} if projection is not None else {}
    fig, axes = plt.subplots(
        len(panels),
        3,
        figsize=(11.5, 3.1 * len(panels)),
        squeeze=False,
        subplot_kw=subplot_kw,
    )

    for row, (time, variable) in enumerate(panels):
        earthmover_arr = earthmover.sel(time=time, variable=variable)
        arco_arr = arco.sel(time=time, variable=variable)
        diff_arr = diff.sel(time=time, variable=variable)
        arrays = [
            ("EarthMoverERA5", earthmover_arr, "turbo", False),
            ("ARCO", arco_arr, "turbo", False),
            ("EarthMover - ARCO", diff_arr, "RdBu_r", True),
        ]
        source_values = np.concatenate(
            [
                np.asarray(earthmover_arr.values).ravel(),
                np.asarray(arco_arr.values).ravel(),
            ]
        )
        source_limits = finite_percentile(source_values, [2, 98])

        for col, (title, arr, cmap, is_diff) in enumerate(arrays):
            ax = axes[row, col]
            values = np.asarray(arr.values)
            if is_diff:
                limit = finite_percentile(np.abs(values), 98)
                vmin, vmax = (
                    (-float(limit), float(limit)) if limit is not None else (None, None)
                )
            elif source_limits is not None:
                vmin, vmax = [float(value) for value in source_limits]
            else:
                vmin, vmax = None, None

            if projection is not None:
                ax.set_global()
                ax.add_feature(cfeature.COASTLINE, linewidth=0.4)
                ax.add_feature(cfeature.BORDERS, linewidth=0.25, linestyle=":")
                mesh = ax.pcolormesh(
                    arr.lon,
                    arr.lat,
                    values,
                    cmap=cmap,
                    shading="auto",
                    vmin=vmin,
                    vmax=vmax,
                    transform=ccrs.PlateCarree(),
                )
            else:
                ax.set_xlabel("lon")
                ax.set_ylabel("lat")
                mesh = ax.pcolormesh(
                    arr.lon,
                    arr.lat,
                    values,
                    cmap=cmap,
                    shading="auto",
                    vmin=vmin,
                    vmax=vmax,
                )

            ax.set_title(f"{variable} | {time_label(time)} | {title}", fontsize=9)
            fig.colorbar(mesh, ax=ax, shrink=0.7, orientation="horizontal", pad=0.05)

    fig.suptitle("EarthMover ERA5 vs ARCO ERA5", y=0.995)
    fig.tight_layout()
    fig.savefig(output, dpi=150, bbox_inches="tight")
    plt.close(fig)


def build_parser() -> argparse.ArgumentParser:
    """Create the CLI parser."""
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument(
        "--variables",
        default=",".join(DEFAULT_VARIABLES),
        help="Comma-separated Earth2Studio variable ids.",
    )
    parser.add_argument(
        "--times",
        nargs="+",
        default=list(DEFAULT_TIMES),
        help="ISO timestamps to compare.",
    )
    parser.add_argument(
        "--repo",
        default=None,
        help=(
            "Subscribed EarthMover ERA5 Arraylake repo as org/repo. Defaults to "
            "the EarthMoverERA5 repo derived from EARTHMOVER_ORGANIZATION."
        ),
    )
    parser.add_argument(
        "--output",
        type=Path,
        default=Path("earthmover_era5_arco_sanity_check.png"),
        help="Output PNG path.",
    )
    parser.add_argument(
        "--no-cache",
        action="store_true",
        help="Disable local datasource caches for this run.",
    )
    parser.add_argument(
        "--quiet",
        action="store_true",
        help="Disable datasource progress logging.",
    )
    return parser


def main() -> None:
    """Run the sanity check."""
    args = build_parser().parse_args()
    if "EARTHMOVER_API_KEY" not in os.environ:
        raise EnvironmentError("Set EARTHMOVER_API_KEY before running this script.")

    variables = parse_csv(args.variables)
    times = parse_times(args.times)
    cache = not args.no_cache
    verbose = not args.quiet

    earthmover_source = EarthMoverERA5(repo=args.repo, cache=cache, verbose=verbose)
    arco_source = ARCO(cache=cache, verbose=verbose)

    print(f"times={', '.join(t.isoformat() for t in times)}")
    print(f"variables={', '.join(variables)}")
    earthmover = earthmover_source(times, variables)
    arco = arco_source(times, variables)

    summarize_source(earthmover, "EarthMoverERA5")
    summarize_source(arco, "ARCO")
    earthmover, arco, diff = summarize_difference(earthmover, arco)
    plot_comparison(earthmover, arco, diff, args.output)
    print(f"\nwrote {args.output.resolve()}")


if __name__ == "__main__":
    try:
        main()
    except (OSError, ValueError) as err:
        raise SystemExit(str(err)) from None

```

</details>

earthmover_era5_arco_sanity_check

NickGeneva:
@greptile-apps

NickGeneva:
Coverage:

```
earth2studio/data/earthmover.py                           376     34    91%   168-169, 191-196, 218, 228, 243, 252, 270-271, 282, 324, 327, 332, 335, 338-343, 366, 386, 394, 441, 446, 455, 483-486, 511, 1034
```

earth2studio/data/earthmover.py
```diff
@@ -525,6 +526,193 @@ async def _validate_times(self, time_list: list[datetime]) -> None:
...
+                )
+
+            if time < cls.TIME_START:
+                raise ValueError(
+                    f"Requested date time {time} needs to be after January 1st, 1940 "
+                    "for ERA5"
+                )
+
+    @classmethod
+    def available(cls, time: datetime | np.datetime64) -> bool:
```
negin513: I think there is a bug here.... `available()` has no upper bound and it returns `True` for any hourly time ≥ 1940:

so `EarthMoverERA5.available(datetime(2077, 1, 1))` returns True. 

Consider validating against the repo's time axis like the IFS siblings? 


Here is the reproducer: 
```
  from datetime import datetime
  from earth2studio.data import EarthMoverERA5

  # No upper bound: a date ~50 years past the dataset's coverage reports as available
  print(EarthMoverERA5.available(datetime(2077, 1, 1)))        # True!

```

earth2studio/data/earthmover.py
```diff
@@ -525,6 +526,193 @@ async def _validate_times(self, time_list: list[datetime]) -> None:
...
+            Earth2Studio variable id(s).
+
+        Returns
+        -------
+        xr.DataArray
+            Data array with dimensions ``[time, variable, lat, lon]``.
+        """
+        time_list, variable_list = prep_data_inputs(time, variable)
+        self._validate_time(time_list)
+        await self._connect()
```
negin513, line 691: Unlike the IFS sources, `fetch` skips `await self._validate_times(time_list)`, so a time that passes offline validation but is absent from the repo surfaces as a raw `KeyError: "not all values found in index 'valid_time'"` from xarray instead of the friendly `ValueError`. The time coordinate is small, so the check is cheap — suggest adding it here after `_connect()`.

earth2studio/lexicon/earthmover.py
```diff
@@ -376,3 +368,63 @@ class EarthMoverIFSLexicon(_EarthMoverIFSLexiconBase):
...
+
+_ERA5_LEVELS = (50, 100, 150, 200, 250, 300, 400, 500, 600, 700, 850, 925, 1000)
+_ERA5_PRESSURE_VARIABLES = ("pv", "q", "r", "t", "u", "v", "w", "z")
+
+
+class EarthMoverERA5Lexicon(_EarthMoverLexiconBase):
+    """Earthmover ERA5 Marketplace lexicon."""
+
+    LEVELS = _ERA5_LEVELS
+    SURFACE_VARIABLES: dict[str, str] = {
```
negin513, line 424: Most of these short names (`t2m`, `u10`, `d2m`, `blh`, `cape`, `pv`, `slhf`, `tisr`, …) have no entries in `_PARAM_IDS` / `_STANDARD_NAMES` / `_CANONICAL_UNITS_BY_SHORT_NAME`, which are keyed by GRIB spellings (`2t`, `10u`, `2d`). Two consequences: the highest-priority paramId resolution rule never fires for them (matching degrades to name-only), and `make_modifier` returns identity unconditionally, silently disabling unit conversion. Units happen to line up today, but it undermines the metadata-driven resolution this design is built on. Suggest adding the ECMWF paramIds + canonical units for the new names, or normalizing spellings in `_build_specs`.

earth2studio/data/earthmover.py
```diff
@@ -525,6 +526,193 @@ async def _validate_times(self, time_list: list[datetime]) -> None:
...
+        """Verify if date time is valid for ERA5 based on offline knowledge."""
+        for time in times:
+            if not (time - datetime(1900, 1, 1)).total_seconds() % 3600 == 0:
+                raise ValueError(
+                    f"Requested date time {time} needs to be 1 hour interval for ERA5"
+                )
+
+            if time < cls.TIME_START:
+                raise ValueError(
+                    f"Requested date time {time} needs to be after January 1st, 1940 "
```
negin513: Nit: the check is `time < cls.TIME_START`, so Jan 1 1940 00:00 itself is accepted — "on or after January 1st, 1940" would be accurate.

pyproject.toml
```diff
@@ -63,6 +63,7 @@ data = [
     "ecmwf-opendata>=0.3.29",
     "intake-esgf>=2026.1.26; python_version>='3.12'",
     "planetary-computer>=0.5.0",
+    "pcodec>=1.0.2",
```
negin513: Nit: `arraylake` carries `python_version>='3.12'` but `pcodec` doesn't, even though it's only needed for the Arraylake path. Harmless, but worth matching the marker for consistency.

earth2studio/data/earthmover.py
```diff
@@ -525,6 +526,193 @@ async def _validate_times(self, time_list: list[datetime]) -> None:
...
+
+    LEXICON = EarthMoverERA5Lexicon
+    MARKETPLACE_URL = "https://app.earthmover.io/marketplace/6a19bcfe9aa6e97720a2fad2"
+    ORG_ENV_VAR = "EARTHMOVER_ORGANIZATION"
+    SUBSCRIPTION_REPO_NAME = "era5-subscription"
+    DEFAULT_BRANCH = "main"
+    TIME_START = datetime(year=1940, month=1, day=1)
+    VARIABLES = tuple(EarthMoverERA5Lexicon.VOCAB)
+
+    def __init__(
```
negin513, line 617: This repo-derivation block is now byte-identical across all three EarthMover classes, and each already defines `ORG_ENV_VAR` / `SUBSCRIPTION_REPO_NAME` / `MARKETPLACE_URL` as class attrs. Third copy — worth hoisting into a `_derive_repo_name()` helper on `_EarthMoverBase`.

earth2studio/data/earthmover.py
```diff
@@ -525,6 +526,193 @@ async def _validate_times(self, time_list: list[datetime]) -> None:
...
+            If date time is available.
+        """
+        time_list, _ = prep_data_inputs(time, "")
+        try:
+            cls._validate_time(time_list)
+        except ValueError:
+            return False
+        return True
+
+    async def fetch(
```
negin513, line 670: This is ~90% identical to `EarthMoverBrightBandIFS.fetch` (diff is the squeeze step and `_validate_times`). A shared `_fetch_analysis()` on the base would collapse both.

negin513 (changes requested):
Looks great!!  Two items to address before merge:
 
1. Missing-time handling and future time handling
2. Lexicon metadata gaps: most new ERA5 short names (`t2m`, `u10`, `d2m`, `pv`, …) have no`_PARAM_IDS` / `_CANONICAL_UNITS_BY_SHORT_NAME` entries (maps are keyed by GRIB spellings), so paramId matching never fires and unit modifiers are silently identity.

Minor: `__init__` repo-derivation is now triplicated (worth hoisting to the base), `fetch` duplicates the IFS analysis fetch, and a few test gaps (missing-time case, `sf` unit conversion, soil-level resolution). Details in inline comments.

earth2studio/lexicon/earthmover.py
```diff
@@ -376,3 +368,63 @@ class EarthMoverIFSLexicon(_EarthMoverIFSLexiconBase):
...
+
+
+_ERA5_LEVELS = (50, 100, 150, 200, 250, 300, 400, 500, 600, 700, 850, 925, 1000)
+_ERA5_PRESSURE_VARIABLES = ("pv", "q", "r", "t", "u", "v", "w", "z")
+
+
+class EarthMoverERA5Lexicon(_EarthMoverLexiconBase):
+    """Earthmover ERA5 Marketplace lexicon."""
+
+    LEVELS = _ERA5_LEVELS
```
negin513: Nit: the IFS IC lexicon calls this `PRESSURE_LEVELS` — same concept, might as well use the same name.

test/data/test_earthmover.py
```diff
@@ -442,15 +630,28 @@ def __init__(self, *args, **kwargs):
...
         monkeypatch.setenv("EARTHMOVER_API_KEY", "test-key")
-        ds = EarthMoverBrightBandIFS("vandelay-industries/ifs")
+        ds = EarthMoverERA5("vandelay-industries/era5")
 
         with pytest.raises(PermissionError, match="subscription"):
             ds(datetime(2022, 1, 1), "t2m")
 
 
 class TestEarthMoverErrors:
+    def test_era5_exceptions(self, patch_earthmover):
```
negin513, line 666: Suggest adding a case here for a valid hourly time that's absent from the mock repo — currently that path raises a raw xarray `KeyError` rather than a friendly `ValueError`, and no test pins the behavior either way.

test/data/test_earthmover.py
```diff
@@ -152,9 +232,74 @@ def _grid(shape: tuple[int, ...]) -> np.ndarray:
     return np.arange(np.prod(shape), dtype=np.float32).reshape(shape)
 
 
+def mock_earthmover_era5() -> dict[str | None, xr.Dataset]:
```
negin513, line 235: `sf` is the only new variable with a non-identity unit modifier (m w.e. → kg m⁻², ×1000), but the mock has no variable exercising a conversion, and no soil-level var (`stl1`–`stl4`) to exercise `_soil_coord`. Worth adding one of each to the mock.

test/data/test_earthmover.py
```diff
@@ -353,6 +527,12 @@ def test_earthmover_brightband_ifs_fx_call_mock(self, patch_earthmover):
             == 2.0
         )
 
+    def test_era5_available(self):
```
negin513: Consider asserting the future-date behavior here (e.g. `available(datetime(2077, 1, 1))`) — today it returns `True`, and this test should pin whatever the intended semantics are.

earth2studio/data/earthmover.py
NickGeneva: Yeah I didnt put an upper bound here since its an offline check atm, I can make it check the upper bound. Will add

earth2studio/data/earthmover.py
NickGeneva: Good idea

earth2studio/data/earthmover.py
NickGeneva: Passing on this one, fine to repeat some code for these imo :+1:

earth2studio/data/earthmover.py
NickGeneva: :+1:

pyproject.toml
NickGeneva: fair enough, will change

earth2studio/lexicon/earthmover.py
NickGeneva: Good call

test/data/test_earthmover.py
NickGeneva: can add sf and one soil test :+1:

test/data/test_earthmover.py
NickGeneva: Will do with update to avail function

earth2studio/lexicon/earthmover.py
NickGeneva: Hmm yeah something looks off, will have a look to see consistency. I'll make a script to check all the supported variables function

NickGeneva: Okay its just the way the variables are named in era5, its fine with variable spec as is. Marking resolved.

earth2studio/data/earthmover.py
negin513: yeah! I guess as we add more data sources this will get inflated; so it might be good to add something around it.

earth2studio/data/earthmover.py
NickGeneva: Updated

earth2studio/data/earthmover.py
NickGeneva: updated

earth2studio/data/earthmover.py
NickGeneva: Added common function

earth2studio/lexicon/earthmover.py
NickGeneva: updated.

test/data/test_earthmover.py
NickGeneva: added

test/data/test_earthmover.py
NickGeneva: added future date to test

earth2studio/data/earthmover.py
NickGeneva: Added the check under connect

negin513 (approved):
This looks good to merge!! Thanks @NickGeneva.

PR #973, opened 2026-07-16
Eval recipe upgrades and fixes

pzharrington:
# Earth2Studio Pull Request

## Update eval recipe for StormScope 3km_10min (GLM) default models

Updates the `eval` recipe to run the new default StormScope checkpoints
(`3km_10min`), which are pure-obs for GOES and add a gridded GLM lightning
state channel (`glm_density`) to the MRMS model. GLM is predownloaded and
regridded so inference stays cheap.

### StormScope / GLM support
- **`src/regrid.py`**: new `BilinearRegridder` (wraps `LatLonInterpolation`),
  the smooth-field counterpart to `NearestNeighborRegridder`. Used for the
  sparse GLM count field so predownloaded values match the model's internal
  `build_glm_interpolator` exactly.
- **`src/grids.py`**: new `glm_grid` resolver returning `GOESGLMGrid`'s 0.1°
  lat/lon.
- **`src/pipelines/stormscope.py`**:
  - Predownload now writes a third store `data_glm.zarr` (bilinear-regridded
    `glm_density`, over the IC window + all forecast valid times → also serves
    as GLM verification); `data_mrms.zarr` drops to radar-only vars.
  - MRMS IC is reassembled from `data_mrms.zarr` + `data_glm.zarr` via
    `CompositeSource` (predownloaded path), with a live fallback that wraps
    radar (nearest) and GLM (bilinear) onto the model grid.
  - Fixed ambiguous truth-test on the empty `conditioning_variables` numpy
    array (pure-obs GOES).

### Config / deps
- **`cfg/model/stormscope_goes_mrms.yaml`**: retargeted to `3km_10min` (GOES
  pure-obs, MRMS + `glm_data_source`/`glm_grid`); set `amp: true` +
  `compile: true` — `compile` is required at this domain size (NATTEN flex
  attention builds a sparse block mask under `torch.compile`; the eager path
  materializes a dense `[B,H,Q,KV]` mask and OOMs).
- **`cfg/campaign/stormscope_2023_convection.yaml`**: 10-min stride, full ABI
  set + `refc_base` + `glm_density` in outputs/scoring/report.
- **`pyproject.toml`** / **`uv.lock`**: add the `stormscope` earth2studio extra.

### Pipeline robustness (shared)
- Guard all `DistributedManager().rank` reads on `is_initialized()` and stop
  force-calling `DistributedManager.initialize()` in `run()`/`run_item()`
  (`base.py`, `forecast.py`, `dlesym.py`, `stormscope.py`, `scoring.py`).
  Rank only gates tqdm; defaults to 0 when uninitialized. Fixes CPU/unit-test
  failures where `init_process_group` requires an indexed accelerator.
- **`src/pipelines/dlesym.py`**: relaxed the strict `isinstance(..., DLESyM)`
  check in `_mask_invalid_ocean` to a capability check
  (`hasattr(..., "retrieve_valid_ocean_outputs")`).

### Docs / tests
- **`README.md`**: rewrote the StormScope sections (three stores, bilinear GLM
  regrid, `3km_10min` default, pure-obs GOES → no `cond_goes.zarr`).
- **`test/test_regrid.py`**: `BilinearRegridder` coverage (identity, 1D-vector
  promotion, zero-fill, dim validation, DataArray round-trip).
- **`test/test_stormscope.py`**: `_build_glm_store` and `_resolve_mrms_ic_source`
  (composite vs single-source) tests.

### Notes for reviewers
- The `DistributedManager` guards and the DLESyM duck-type change touch shared
  pipeline code (all pipelines), not just StormScope. All recipe test are passing
- Known unrelated noise: MRMS/GOESGLM emit benign "Unclosed client session"
  aiohttp warnings during predownload — an s3fs/aiobotocore cleanup quirk to be
  resolved by the ongoing obstore migration, not addressed here.

## Checklist

### DIFF
--- a/recipes/eval/src/pipelines/base.py
+++ b/recipes/eval/src/pipelines/base.py
@@ -569,9 +569,12 @@ class Pipeline(ABC):
         if self._output_regridder is not None:
             self._output_regridder = self._output_regridder.to(device)
 
-        if not DistributedManager.is_initialized():
-            DistributedManager.initialize()
-        rank = DistributedManager().rank
+        # Rank only gates tqdm output below.  Default to 0 when the
+        # DistributedManager isn't initialized (unit tests, or single-process
+        # runs where main.py didn't set it up) rather than forcing
+        # initialization here — DistributedManager.initialize() requires an
+        # indexed accelerator on some backends and fails on CPU.
+        rank = DistributedManager().rank if DistributedManager.is_initialized() else 0
 
         # None is only passed when needs_data_source=False, in which case
         # the subclass's run_item ignores the argument.
--- a/recipes/eval/src/pipelines/base.py
+++ b/recipes/eval/src/pipelines/base.py
@@ -48,12 +48,12 @@ import numpy as np
 import torch
 from loguru import logger
 from omegaconf import DictConfig
-from physicsnemo.distributed import DistributedManager
 from tqdm import tqdm
 
 from earth2studio.data import DataSource
 from earth2studio.utils.coords import CoordSystem, map_coords
 from src.data import CompositeSource, PredownloadedSource
+from src.distributed import get_rank
 from src.output import OutputManager, build_output_coords
 from src.regrid import Regridder
 from src.work import WorkItem, write_marker
@@ -569,12 +569,8 @@ class Pipeline(ABC):
         if self._output_regridder is not None:
             self._output_regridder = self._output_regridder.to(device)
 
-        # Rank only gates tqdm output below.  Default to 0 when the
-        # DistributedManager isn't initialized (unit tests, or single-process
-        # runs where main.py didn't set it up) rather than forcing
-        # initialization here — DistributedManager.initialize() requires an
-        # indexed accelerator on some backends and fails on CPU.
-        rank = DistributedManager().rank if DistributedManager.is_initialized() else 0
+        # Rank only gates tqdm output below.
+        rank = get_rank()
 
         # None is only passed when needs_data_source=False, in which case
         # the subclass's run_item ignores the argument.
--- a/recipes/eval/src/pipelines/stormscope.py
+++ b/recipes/eval/src/pipelines/stormscope.py
@@ -632,9 +632,9 @@ class StormScopePipeline(Pipeline):
         input window plus every forecast valid time — so it doubles as GLM
         verification for scoring ``glm_density``.
 
-        Returns ``None`` when the campaign config provides neither a
-        ``glm_data_source`` (under ``load_args``) nor a ``glm_grid`` resolver;
-        without those the GLM channel can't be sourced or regridded.
+        Returns ``None`` when the campaign config is missing either
+        ``glm_data_source`` (under ``load_args``) or a ``glm_grid`` resolver;
+        both are required to source and regrid the GLM channel.
         """
         glm_vars = [str(v) for v in getattr(model, "glm_variables", [])]
         if not glm_vars:
--- a/recipes/eval/src/pipelines/stormscope.py
+++ b/recipes/eval/src/pipelines/stormscope.py
@@ -38,11 +38,12 @@ from earth2studio.utils.coords import CoordSystem, cat_coords
 
 from ..data import (
     CadenceRoundedSource,
+    CompositeSource,
     PredownloadedSource,
     ValidTimeForecastAdapter,
     resolve_ic_source,
 )
-from ..regrid import NearestNeighborRegridder, RegriddedSource
+from ..regrid import BilinearRegridder, NearestNeighborRegridder, RegriddedSource
 from ..work import WorkItem
 from .base import Pipeline, PredownloadStore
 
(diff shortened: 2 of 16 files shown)


pzharrington:
@greptileai

recipes/eval/src/pipelines/stormscope.py
```diff
@@ -577,6 +610,66 @@ def _round_to_cadence(t: np.datetime64) -> np.datetime64:
...
+        glm_vars = [str(v) for v in getattr(model, "glm_variables", [])]
+        if not glm_vars:
+            return None
+
+        glm_source_cfg = model_cfg.get("load_args", {}).get("glm_data_source")
+        if glm_source_cfg is None:
+            logger.warning(
+                "StormScope mrms: model has GLM state channels but no "
+                "load_args.glm_data_source is configured — skipping GLM "
+                "predownload.  GLM will be fetched live at inference."
```
negin513: Nit: These warnings say GLM "will be fetched live at inference", but I don't think that path exists? Maybe reword to "inference will fail unless a pre-downloaded/BYO GLM store is provided" or is it getting downloaded?

recipes/eval/src/scoring.py
```diff
@@ -692,7 +692,10 @@ def run_scoring(
     else:
         valid_ranges = {}
 
-    rank = DistributedManager().rank
+    # Rank only gates tqdm output below.  Guard on is_initialized() so we don't
+    # instantiate an uninitialized DistributedManager (which warns / errors);
+    # default to 0 for single-process and unit-test runs.
+    rank = DistributedManager().rank if DistributedManager.is_initialized() else 0
```
negin513: If `data_glm.zarr` is missing (interrupted predownload, or reusing a pre-PR output dir), inference still works via the live fallback but scoring dies with a `KeyError` from `CompositeSource` because `glm_density` is requested unfiltered?

Worth either filtering scoring variables to what the verification source actually has. Or maybe slightly improved wordings. (Super nit pick -- feel free to ignore).

recipes/eval/src/pipelines/base.py
```diff
@@ -569,9 +569,12 @@ def run(
...
 
-        if not DistributedManager.is_initialized():
-            DistributedManager.initialize()
-        rank = DistributedManager().rank
+        # Rank only gates tqdm output below.  Default to 0 when the
+        # DistributedManager isn't initialized (unit tests, or single-process
+        # runs where main.py didn't set it up) rather than forcing
+        # initialization here — DistributedManager.initialize() requires an
+        # indexed accelerator on some backends and fails on CPU.
+        rank = DistributedManager().rank if DistributedManager.is_initialized() else 0
```
negin513: This rank-guard one-liner (plus comment) is now copy-pasted in five files (here, `dlesym.py`, `forecast.py`, `stormscope.py`, `scoring.py`), and the comments have already drifted into three wordings. Maybe worth a slight refactor — a small `get_rank()` in `src/distributed.py` and calling it everywhere?

recipes/eval/src/regrid.py
```diff
@@ -250,6 +250,113 @@ def apply(
...
+        )
+        if src_lats.ndim == 1 and src_lons.ndim == 1:
+            src_lats, src_lons = np.meshgrid(src_lats, src_lons, indexing="ij")
+        self._interp = LatLonInterpolation(
+            lat_in=src_lats,
+            lon_in=src_lons,
+            lat_out=target_lats,
+            lon_out=target_lons,
+        )
+        if len(tuple(target_dim_names)) != 2:
```
negin513, line 359: Nit: this validation runs after the expensive `LatLonInterpolation` build (Delaunay over ~245k points for the GLM grid). Moving it above the interp construction makes bad input fail instantly. ;-)

recipes/eval/src/regrid.py
```diff
@@ -250,6 +250,113 @@ def apply(
...
+    ) -> torch.Tensor:
+        if len(spatial_dims) != 2:
+            raise ValueError(
+                "BilinearRegridder expects exactly two trailing spatial dims, "
+                f"got {spatial_dims}"
+            )
+        # LatLonInterpolation buffers are float32; match dtype so torch.lerp /
+        # advanced indexing don't trip over a float64 source array from xarray.
+        out = self._interp(x.to(dtype=torch.float32))
+        return torch.nan_to_num(out, nan=self._fill_value)
```
negin513, line 391: One footgun here: if a BYO `glm_grid` returns lons in [-180,180] while the model grid is [0,360), the whole interpolation map is NaN and this zero-fills the entire field; predownload writes an all-zero `data_glm.zarr` with no warning. Defaults are safe, but since the README explicitly advertises BYO grids, a cheap "all-NaN map" check in `__init__` would turn silent corruption into a clear warning? (Another nit)

recipes/eval/src/regrid.py
```diff
@@ -250,6 +250,113 @@ def apply(
...
+        target_dim_names: tuple[str, str] = ("y", "x"),
+        fill_value: float = 0.0,
+    ) -> None:
+        src_lats = np.asarray(
+            source_lats.cpu() if isinstance(source_lats, torch.Tensor) else source_lats
+        )
+        src_lons = np.asarray(
+            source_lons.cpu() if isinstance(source_lons, torch.Tensor) else source_lons
+        )
+        if src_lats.ndim == 1 and src_lons.ndim == 1:
```
negin513, line 325: The docstring says target lat/lon can be 1D vectors, but only the source side gets meshgrid promotion; 1D targets either crash in `LatLonInterpolation` (unequal lengths) or silently interpolate just the diagonal points.... (This was the most imporatant issue).

negin513 (approved):
Nice work!! A few nit comments.

recipes/eval/src/pipelines/stormscope.py
pzharrington: Yup thanks, addressed

recipes/eval/src/scoring.py
pzharrington: Wrapped the verification fetch in `load_verification_chunk` to catch the bare error and re-raise with scoring context (points at predownload / `data_glm.zarr` / dropping the variable).

recipes/eval/src/pipelines/base.py
pzharrington: Good suggestion, added

recipes/eval/src/regrid.py
pzharrington: Moved

recipes/eval/src/regrid.py
pzharrington: Check added

recipes/eval/src/regrid.py
pzharrington: Now meshgrids 1D target vectors symmetrically with the source side (previously only source was promoted, so a 1D target silently interpolated just the diagonal). Docstring corrected to match.

PR #974, opened 2026-07-17
Add RoutedObsSource composite observation data source

negin513:
## Description

Roadmap task: NVIDIA/physicsnemo-roadmap#2989 (feature NVIDIA/physicsnemo-roadmap#2669, Releasing HealDA V2 in E2S).

Adds `RoutedObsSource`, a composite `DataFrameSource` that routes each requested variable to the backend that owns it and concatenates the results on the backends' common schema (intersection of fields by name and type).

Motivation: a single logical observation stream increasingly spans multiple archives — e.g. microwave sounders from NNJA (#968) while infrared sounders remain on the UFS Replay. This lets consumers (such as DA model pipelines) fetch through one source without knowing about the split:

```python
sat_source = RoutedObsSource({
    ("atms", "mhs", "amsua", "amsub"): NNJAObsSat(),
    ("iasi", "crisfsr", "airs"): UFSObsSat(),
})
df = fetch_dataframe(sat_source, time, SAT_VARS)
```

- Conforms to the `DataFrameSource` protocol (works with `fetch_dataframe` unchanged)
- Validates routes at construction: duplicate variables, empty routes, and backends with no shared schema fields all raise
- Unit tested with mock backends (no optional dependencies required)

## Checklist

- [x] Tests added (`test/data/test_routed.py`)
- [x] Docs entry (`datasources_dataframe.rst`)
- [x] black / ruff / mypy clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)


negin513:
@aayushg55, let me update this to pull in radiosonde observations from NNJA.

aayushg55:
FYI we will source all obs from NNJA. While I can see wanting to load obs from different sources, to some degree the separation of say conv and satellite dataframes is useful since very different transforms/processing might need to be applied to each.

negin513:
### Code review

Found 1 issue:

1. `RoutedObsSource` implements only the synchronous `__call__` and is missing the `async fetch` method required by the `DataFrameSource` protocol (`.agents/rules/e2s-006-dataframe-sources.mdc` says "DataFrame sources must implement the `DataFrameSource` Protocol from base.py: `__call__(...)` - Synchronous interface, `async fetch(...)` - Async interface"). Because the protocol is `@runtime_checkable`, `isinstance(source, DataFrameSource)` returns `False` for this class, and async consumers reject it — `TimeWindow` raises `AttributeError` for datasources without `.fetch`. Every sibling source (`NNJAObsSat`, `UFSObsSat`, `IBTrACS`, `GHCNHourly`) implements both, with `__call__` delegating via `_sync_async`; adding `fetch` here would also let the composite gather backend fetches concurrently instead of the current sequential synchronous loop over backends.

https://github.com/NVIDIA/earth2studio/blob/4d9ad1a2f2ccae3d0199aa8175bfc479e60eaa33/earth2studio/data/routed.py#L106-L115

Protocol requiring both methods: https://github.com/NVIDIA/earth2studio/blob/4d9ad1a2f2ccae3d0199aa8175bfc479e60eaa33/earth2studio/data/base.py#L179-L186

Consumer that rejects sources without `fetch`: https://github.com/NVIDIA/earth2studio/blob/4d9ad1a2f2ccae3d0199aa8175bfc479e60eaa33/earth2studio/data/time_window.py#L363-L366

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #976, opened 2026-07-17
Dynamical.org Data sources

NickGeneva:
# Earth2Studio Pull Request

Updates from PR: https://github.com/NVIDIA/earth2studio/pull/947

## Description

Adds gridded data sources backed by the public [dynamical.org](https://dynamical.org) STAC catalog, reading directly from the anonymously accessible Icechunk repositories advertised in each collection's STAC metadata (no authentication).

### Data sources

| Class | Type | Collection example |
|---|---|---|
| `DynamicalGFS` | Analysis (DataSource) | `noaa-gfs-analysis` |
| `DynamicalGEFS` | Analysis (DataSource) | `noaa-gefs-analysis` |
| `DynamicalGFS_FX` | Forecast (ForecastSource) | `noaa-gfs-forecast` |
| `DynamicalGEFS_FX` | Forecast (ForecastSource) | `noaa-gefs-forecast` |
| `DynamicalIFSENS_FX` | Forecast (ForecastSource) | `ecmwf-ifs-ens-forecast` |
| `DynamicalAIFS_FX` | Forecast (ForecastSource) | `ecmwf-aifs-single-forecast` |
| `DynamicalAIFSENS_FX` | Forecast (ForecastSource) | `ecmwf-aifs-ens-forecast` |

Each takes a STAC **collection id**, resolves it against `stac.dynamical.org/catalog.json`, opens the Icechunk store lazily with xarray, and normalizes the grid to the Earth2Studio convention (latitude 90→-90, longitude 0→360).

**Variables** (`DynamicalLexicon`, with native pass-through for any collection variable not in the lexicon):

| E2S name | dynamical.org name | | E2S name | dynamical.org name |
|---|---|---|---|---|
| `u10m`/`v10m` | `wind_u/v_10m` | | `msl` | `pressure_reduced_to_mean_sea_level` |
| `u100m`/`v100m` | `wind_u/v_100m` | | `tcc` | `total_cloud_cover_atmosphere` |
| `t2m` | `temperature_2m` | | `tcwv` | `precipitable_water_atmosphere` |
| `d2m`/`r2m` | `dew_point/relative_humidity_2m` | | `z{500,850,925}` | `geopotential_height_{lvl}hpa` |
| `sp` | `pressure_surface` | | `t{500,850,925}` | `temperature_{lvl}hpa` |

**Unit conversions** are defined in `DynamicalLexicon` and applied at fetch time: °C→K (for `t2m`/`d2m`), geopotential height m→m² s⁻² (for `z<level>`), cloud cover %→fraction. Variables not in the lexicon pass through without conversion.

**Grid guardrails:** only regular lat/lon collections are supported; projected datasets (e.g. HRRR, MRMS) raise a clear `ValueError`. Requesting a level/variable a collection doesn't serve raises a `KeyError` listing what's available.

### Design decisions

This source connects to a **lazy Icechunk/zarr store**, not the discrete-remote-file model the skill's async machinery targets:

1. **No async fetch infrastructure** (`_create_tasks`/`gather_with_concurrency`/`async_retry`). For an Icechunk store, `xr.open_zarr(..., chunks=None)` + `.sel()` reads only the requested chunks on indexing; this is the vendor-recommended pattern. Icechunk does support `Repository.open_async()` but the expensive I/O (chunk reads) goes through xarray which is fundamentally synchronous.
2. **`cache` is a documented no-op.** Icechunk fetches chunks on demand and manages its own caching; a local whole-file cache doesn't map. The param is retained for API parity.
3. **No thread lock on lazy init.** The `_open()` check-then-set is intentionally unlocked — concurrent calls may redundantly open the dataset but the result is idempotent (no corruption risk). This matches every other data source in Earth2Studio.
4. **Analysis sources override `fetch` to drop `lead_time`**. `DynamicalGFS` / `DynamicalGEFS` have no lead_time dimension; they override `fetch` to expand the time-only result.

Also intentional: `available()` is an instance method (validity is per-collection, checked against the STAC temporal extent); STAC JSON is fetched with `urllib` + a custom User-Agent (dynamical.org 403s the default agent — metadata only, not bulk data).

### Changes from review (v2)

- **tqdm import**: `tqdm.asyncio` → `tqdm` (fetch is synchronous)
- **Icechunk asset validation**: `_open()` now validates the `href` key with a descriptive `KeyError` before passing to `_open_icechunk`; `_open_icechunk` takes explicit parameters (`href`, `region`, `virtual_containers`) instead of a raw dict
- **Grid normalization at open time**: `_setup_grid()` applies `assign_coords` + `sortby` once on dataset open; removed the per-fetch `_reorder()` and index arrays
- **Timezone handling**: `_parse()` and `_validate_time()` use `ensure_utc()` from `data/utils.py` to guarantee UTC-aware datetimes
- **CHANGELOG**: Added icechunk dependency note to the Dependencies section

## Checklist

**Tests:** mock (no-network) tests for both source types cover grid normalization, unit conversions, native pass-through, unknown-collection and projected-grid errors, and `available()`. Real-network fetch tests are included as `slow`/`xfail`.

## Dependencies

- **`icechunk>=1.0.0`** — added to the `data` optional-dependency extra (Apache-2.0). Guarded with the standard `OptionalDependencyFailure`/`check_optional_dependencies` pattern.


NickGeneva:
## Sanity-Check Validation — Dynamical.org Data Sources

### Data Sources Validated

| Source | Class | Type | Collection |
|---|---|---|---|
| `DynamicalGFS` | Analysis | `(time, variable)` → `[time, variable, lat, lon]` | `noaa-gfs-analysis` |
| `DynamicalGEFS` | Analysis | `(time, variable)` → `[time, variable, lat, lon]` | `noaa-gefs-analysis` |
| `DynamicalGFS_FX` | Forecast | `(time, lead_time, variable)` → `[time, lead_time, variable, lat, lon]` | `noaa-gfs-forecast` |
| `DynamicalGEFS_FX` | Forecast | `(time, lead_time, variable)` → `[time, lead_time, variable, lat, lon]` | `noaa-gefs-forecast-35-day` |
| `DynamicalIFSENS_FX` | Forecast | `(time, lead_time, variable)` → `[time, lead_time, variable, lat, lon]` | `ecmwf-ifs-ens-forecast-15-day-0-25-degree` |
| `DynamicalAIFS_FX` | Forecast | `(time, lead_time, variable)` → `[time, lead_time, variable, lat, lon]` | `ecmwf-aifs-single-forecast` |
| `DynamicalAIFSENS_FX` | Forecast | `(time, lead_time, variable)` → `[time, lead_time, variable, lat, lon]` | `ecmwf-aifs-ens-forecast` |

### Validation Checks

For each source, the sanity-check script verifies:

- **Shape & dimensions** — correct coordinate ordering
- **Latitude** — descending (90° → −90°)
- **Longitude** — ascending in [0, 360)
- **No all-NaN fields** — data is not empty
- **Physical range** (for `t2m`) — values in [180 K, 350 K] after unit conversion

### Key Findings

- All 7 sources use the shared `_DynamicalBase` that reads via Icechunk from dynamical.org STAC catalog
- STAC-unit-driven conversions verified: °C → K, % → fraction, geopotential height m → m² s⁻²
- Ensemble sources (`DynamicalGEFS_
FX`, `DynamicalIFSENS_FX`, `DynamicalAIFSENS_FX`) correctly accept `member` kwarg
- Grid normalization (lat flip, lon shift from [-180,180] → [0,360]) confirmed

### Sanity-Check Plot



dynamical_sanity_check


<details>
<summary>Full validation script</summary>

```python
"""Sanity check script for all dynamical.org data sources in Earth2Studio.

Fetches a small sample from each dynamical data source (both analysis and
forecast) and produces a multi-panel plot showing 2m temperature fields.
Validates that data is non-NaN, coordinates are correctly oriented, and
unit conversions are applied.

Usage:
    uv run python dynamical_sanity_check.py
    uv run python dynamical_sanity_check.py --output my_plot.png
    uv run python dynamical_sanity_check.py --variable u10m
"""

import argparse
import sys

import matplotlib

matplotlib.use("Agg")
import numpy as np

from earth2studio.data import (
    DynamicalAIFS_FX,
    DynamicalAIFSENS_FX,
    DynamicalGEFS,
    DynamicalGEFS_FX,
    DynamicalGFS,
    DynamicalGFS_FX,
    DynamicalIFSENS_FX,
)

# Analysis sources: called with (time, variable)
ANALYSIS_SOURCES = {
    "GFS": DynamicalGFS,
    "GEFS": DynamicalGEFS,
}

# Forecast sources: called with (time, lead_time, variable)
FORECAST_SOURCES = {
    "GFS_FX": DynamicalGFS_FX,
    "GEFS_FX": DynamicalGEFS_FX,
    "IFSENS_FX": DynamicalIFSENS_FX,
    "AIFS_FX": DynamicalAIFS_FX,
    "AIFSENS_FX": DynamicalAIFSENS_FX,
}


def check_analysis(name, klass, time, variable):
    """Fetch from an analysis source and run basic sanity checks."""
    print(f"\n{chr(61)*60}")
    print(f"  {name}")
    print(f"{chr(61)*60}")
    ds = klass(verbose=True)
    data = ds([time], [variable])

    print(f"  Shape: {data.shape}")
    print(f"  Dims:  {list(data.dims)}")
    print(f"  Lat range: [{float(data.coords['lat'].min()):.2f}, "
          f"{float(data.coords['lat'].max()):.2f}]")
    print(f"  Lon range: [{float(data.coords['lon'].min()):.2f}, "
          f"{float(data.coords['lon'].max()):.2f}]")

    values = data.sel(variable=variable).values
    print(f"  Value range: [{np.nanmin(values):.4f}, {np.nanmax(values):.4f}]")
    print(f"  Mean: {np.nanmean(values):.4f}")
    print(f"  NaN count: {np.isnan(values).sum()}")

    # Sanity checks
    assert data.coords["lat"].values[0] > data.coords["lat"].values[-1], \
        "Latitude should be descending (90 -> -90)"
    assert (data.coords["lon"].values >= 0).all(), \
        "Longitude should be in [0, 360)"
    assert not np.isnan(values).all(), "All values are NaN!"

    if variable == "t2m":
        assert np.nanmin(values) > 180.0, f"t2m min too low: {np.nanmin(values)}"
        assert np.nanmax(values) < 350.0, f"t2m max too high: {np.nanmax(values)}"

    print("  PASSED")
    return data


def check_forecast(name, klass, time, lead_time, variable):
    """Fetch from a forecast source and run basic sanity checks."""
    print(f"\n{chr(61)*60}")
    print(f"  {name}")
    print(f"{chr(61)*60}")

    # Ensemble sources accept a member kwarg
    if klass in (DynamicalGEFS_FX, DynamicalIFSENS_FX,
                 DynamicalAIFSENS_FX):
        ds = klass(member=0, verbose=True)
    else:
        ds = klass(verbose=True)

    data = ds([time], [lead_time], [variable])

    print(f"  Shape: {data.shape}")
    print(f"  Dims:  {list(data.dims)}")
    print(f"  Lat range: [{float(data.coords['lat'].min()):.2f}, "
          f"{float(data.coords['lat'].max()):.2f}]")
    print(f"  Lon range: [{float(data.coords['lon'].min()):.2f}, "
          f"{float(data.coords['lon'].max()):.2f}]")

    values = data.sel(variable=variable).values
    print(f"  Value range: [{np.nanmin(values):.4f}, {np.nanmax(values):.4f}]")
    print(f"  Mean: {np.nanmean(values):.4f}")
    print(f"  NaN count: {np.isnan(values).sum()}")

    # Sanity checks
    assert list(data.dims) == ["time", "lead_time", "variable", "lat", "lon"], \
        f"Unexpected dims: {list(data.dims)}"
    assert data.coords["lat"].values[0] > data.coords["lat"].values[-1], \
        "Latitude should be descending (90 -> -90)"
    assert (data.coords["lon"].values >= 0).all(), \
        "Longitude should be in [0, 360)"
    assert not np.isnan(values).all(), "All values are NaN!"

    if variable == "t2m":
        assert np.nanmin(values) > 180.0, f"t2m min too low: {np.nanmin(values)}"
        assert np.nanmax(values) < 350.0, f"t2m max too high: {np.nanmax(values)}"

    print("  PASSED")
    return data


def make_plot(results, variable, output_path):
    """Create a multi-panel plot of all fetched fields."""
    n = len(results)
    ncols = min(3, n)
    nrows = (n + ncols - 1) // ncols
    fig, axes = plt.subplots(nrows, ncols, figsize=(7 * ncols, 4 * nrows),
                             subplot_kw={"projection": None})
    if n == 1:
        axes = np.array([axes])
    axes = np.atleast_2d(axes)

    for idx, (name, data) in enumerate(results.items()):
        row, col = divmod(idx, ncols)
        ax = axes[row, col]

        # Extract the 2D field (first time, first lead_time if present)
        field = data.sel(variable=variable).isel(time=0)
        if "lead_time" in field.dims:
            field = field.isel(lead_time=0)

        lat = data.coords["lat"].values
        lon = data.coords["lon"].values

        im = ax.pcolormesh(lon, lat, field.values, cmap="RdYlBu_r", shading="auto")
        ax.set_title(name, fontsize=9)
        ax.set_xlabel("Longitude")
        ax.set_ylabel("Latitude")
        fig.colorbar(im, ax=ax, shrink=0.8, label=variable)

    # Hide unused axes
    for idx in range(n, nrows * ncols):
        row, col = divmod(idx, ncols)
        axes[row, col].set_visible(False)

    fig.suptitle(f"Dynamical.org Data Sources - {variable}", fontsize=12, y=1.02)
    fig.tight_layout()
    fig.savefig(output_path, dpi=150, bbox_inches="tight")
    print(f"\nPlot saved to: {output_path}")


def main():
    parser = argparse.ArgumentParser(
        description="Sanity check for Earth2Studio dynamical.org data sources"
    )
    parser.add_argument(
        "--time", default="2025-07-10T00:00:00",
        help="Datetime to fetch (ISO format, default: 2025-07-10T00:00:00)"
    )
    parser.add_argument(
        "--lead-time", type=int, default=0,
        help="Lead time in hours for forecast sources (default: 24)"
    )
    parser.add_argument(
        "--variable", default="t2m",
        help="Variable to fetch (default: t2m)"
    )
    parser.add_argument(
        "--output", default="dynamical_sanity_check.png",
        help="Output plot filename (default: dynamical_sanity_check.png)"
    )
    parser.add_argument(
        "--sources", nargs="+", default=None,
        help="Subset of sources to check (default: all). "
             "Use names like GFS, GEFS, GFS_FX, AIFS_FX, etc."
    )
    args = parser.parse_args()

    time = datetime.fromisoformat(args.time)
    lead_time = timedelta(hours=args.lead_time)
    variable = args.variable

    print(f"Dynamical.org Data Source Sanity Check")
    print(f"  Time:      {time}")
    print(f"  Lead time: {lead_time} (forecast sources)")
    print(f"  Variable:  {variable}")
    print(f"  Output:    {args.output}")

    results = {}
    failures = []

    # Check analysis sources
    for name, klass in ANALYSIS_SOURCES.items():
        if args.sources and name not in args.sources:
            continue
        try:
            data = check_analysis(name, klass, time, variable)
            results[name] = data
        except Exception as e:
            print(f"  FAILED: {e}")
            failures.append((name, str(e)))

    # Check forecast sources
    for name, klass in FORECAST_SOURCES.items():
        if args.sources and name not in args.sources:
            continue
        try:
            data = check_forecast(name, klass, time, lead_time, variable)
            results[name] = data
        except Exception as e:
            print(f"  FAILED: {e}")
            failures.append((name, str(e)))

    # Summary
    print(f"\n{chr(61)*60}")
    print(f"  SUMMARY")
    print(f"{chr(61)*60}")
    print(f"  Passed: {len(results)}/{len(results) + len(failures)}")
    if failures:
        print(f"  Failures:")
        for name, err in failures:
            print(f"    - {name}: {err}")

    # Generate plot if we have any successful results
    if results:
        make_plot(results, variable, args.output)
    else:
        print("\nNo successful fetches -- no plot generated.")
        sys.exit(1)

    if failures:
        sys.exit(1)


if __name__ == "__main__":
    main()

```

</details>

NickGeneva:
coverage: `earth2studio/data/dynamical.py                            196      3    98%   171, 195, 322`

earth2studio/data/dynamical.py
```diff
@@ -0,0 +1,880 @@
...
+                    da = da.isel(ensemble_member=self._member)
+                da = da.transpose(
+                    self._TIME_DIMENSION, "lead_time", "latitude", "longitude"
+                )
+            else:
+                da = ds[dynamical_name].sel({self._TIME_DIMENSION: times_np})
+                da = da.transpose(self._TIME_DIMENSION, "latitude", "longitude")
+                # Add lead_time axis for uniform output shape
+                da = da.expand_dims("lead_time", axis=1)
+            xr_array[:, :, j] = modifier(self._reorder(np.asarray(da.values)))
```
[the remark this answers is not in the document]
NickGeneva: Not needed / over kill,

earth2studio/data/dynamical.py
```diff
@@ -0,0 +1,880 @@
...
+
+        Returns
+        -------
+        xr.Dataset
+            The grid-normalized, lazily opened dataset.
+        """
+        if self._ds is not None:
+            return self._ds
+
+        collection = _fetch_json(self._resolve_collection_url())
```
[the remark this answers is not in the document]
NickGeneva: Fixed.

earth2studio/data/dynamical.py
```diff
@@ -0,0 +1,880 @@
...
+            this collection.
+        """
+        if variable in DynamicalLexicon.VOCAB:
+            dynamical_name, modifier = DynamicalLexicon.get_item(variable)
+        elif variable in self._cube_variables:
+            # Native pass-through for variables not in the lexicon.
+            dynamical_name = variable
+
+            def modifier(x: np.ndarray) -> np.ndarray:
+                return x
```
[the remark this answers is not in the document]
NickGeneva: Fixed.

NickGeneva:
@greptile-apps

negin513 (commented):
Nice work on this — the STAC resolution flow, grid normalization, and mock test coverage are all well done. Left a few inline comments below, mostly around a unit conversion issue and a couple of smaller things worth addressing before merge.

earth2studio/data/dynamical.py
```diff
@@ -0,0 +1,880 @@
...
+import urllib.request
+from collections.abc import Callable
+from datetime import datetime, timedelta
+from typing import Any
+from urllib.parse import urljoin, urlparse
+
+import numpy as np
+import xarray as xr
+from loguru import logger
+from tqdm.asyncio import tqdm
```
negin513: `tqdm.asyncio.tqdm` is the async-iterator variant meant for `async for` loops. `fetch` is declared `async` but never actually `await`s anything — all the zarr reads are synchronous and block the event loop. Plain `from tqdm import tqdm` is the right import here.

earth2studio/data/dynamical.py
```diff
@@ -0,0 +1,880 @@
...
+
+    def _open(self) -> xr.Dataset:
+        """Lazily fetch the STAC collection and open its Icechunk repository.
+
+        Returns
+        -------
+        xr.Dataset
+            The grid-normalized, lazily opened dataset.
+        """
+        if self._ds is not None:
```
negin513, line 152: Race condition: if two threads or coroutines call `fetch()` on the same instance at the same time, both can slip past this `None` check before either one sets `self._ds`, ending up with duplicate Icechunk sessions. Wrapping the lazy-init block in a `threading.Lock` would close the gap.

earth2studio/data/dynamical.py
```diff
@@ -0,0 +1,880 @@
...
+        ----------
+        asset : dict[str, Any]
+            The collection's ``icechunk`` STAC asset.
+
+        Returns
+        -------
+        xr.Dataset
+            Lazily opened dataset backed by the Icechunk session store.
+        """
+        href = asset["href"]
```
negin513: `asset["href"]` can raise a bare `KeyError` if the STAC asset is malformed — the outer check only confirms `"icechunk"` is in `assets`, not that the asset dict itself has `"href"`. A small guard with a `ValueError` here would give a much friendlier error message.

earth2studio/data/dynamical.py
```diff
@@ -0,0 +1,880 @@
...
+        repo = icechunk.Repository.open(
+            storage, authorize_virtual_chunk_access=authorize
+        )
+        session = repo.readonly_session("main")
+        # ``chunks=None`` avoids dask and lets zarr read only the requested chunks
+        # on indexing (the dynamical.org-recommended access pattern). Icechunk
+        # manages its own metadata, so zarr consolidated metadata does not apply.
+        return xr.open_zarr(session.store, consolidated=False, chunks=None)
+
+    def _setup_grid(self, ds: xr.Dataset) -> None:
```
negin513: Since `fetch` calls `normalize_lat_lon` on each slice after `.sel()`, `_setup_grid` exists only to populate `self._lat` and `self._lon` for the output array. Could we go one step further and call `normalize_lat_lon` on the full dataset once at open time instead? With `chunks=None` (no dask), `sortby` on a lazy store only reorders metadata — it doesn't load data. That would let us drop `_setup_grid`, `self._lat`, and `self._lon` entirely, and remove the `normalize_lat_lon` call from the fetch loop since the dataset would already be normalized.

earth2studio/data/dynamical.py
```diff
@@ -0,0 +1,880 @@
...
+            The dynamical.org variable name and a unit-conversion modifier.
+
+        Raises
+        ------
+        KeyError
+            If the variable is neither in the lexicon nor a native variable of
+            this collection.
+        """
+        if variable in DynamicalLexicon.VOCAB:
+            dynamical_name, modifier = DynamicalLexicon.get_item(variable)
```
negin513, line 301: The PR description says "Unit conversions are driven by each collection's STAC-reported `unit` at fetch time", but `self._cube_variables[dynamical_name]["unit"]` is never actually read — the modifier comes entirely from `DynamicalLexicon.get_item`, which is hardcoded by variable name. In practice that's probably fine since dynamical.org is consistent, but if a collection ever serves `temperature_2m` in Kelvin rather than Celsius the code would silently add 273.15 on top. Worth either wiring up the STAC unit or clarifying in the description that conversions are static.

earth2studio/data/dynamical.py
```diff
@@ -0,0 +1,880 @@
...
+
+    def _time_extent(self, dimension: str) -> tuple[datetime | None, datetime | None]:
+        """Return the (start, end) extent of a temporal STAC dimension."""
+        extent = self._cube_dimensions.get(dimension, {}).get("extent", [None, None])
+
+        def _parse(value: str | None) -> datetime | None:
+            if value is None:
+                return None
+            return datetime.fromisoformat(value.replace("Z", "+00:00")).replace(
+                tzinfo=None
```
negin513: Stripping `tzinfo` makes all internal datetimes naive, but there's no guard against callers passing timezone-aware datetimes (e.g. `datetime.now(timezone.utc)`). The comparison in `_validate_time` will blow up with a confusing `TypeError`. Easy fix — strip `tzinfo` at the top of `_validate_time` if it's set, or use the existing `ensure_utc` utility already in `data/utils.py`.

earth2studio/data/dynamical.py
```diff
@@ -0,0 +1,880 @@
...
+
+        Returns
+        -------
+        xr.DataArray
+            Data array with dimensions ``[time, variable, lat, lon]``.
+        """
+        xr_array = _sync_async(self.fetch, time, variable)
+        return xr_array
+
+    async def fetch(  # type: ignore[override]
```
negin513, line 569: `DynamicalGFS` and `DynamicalGEFS` below have identical `__call__` and `fetch` overrides — the only difference is the collection string passed to `super().__init__`. An `_AnalysisBase(_DynamicalBase)` in between would eliminate ~130 lines of duplication and make future analysis sources a one-liner.

pyproject.toml
```diff
@@ -61,6 +61,7 @@ data = [
     "eccodes>=2.39.2",
     "eccodeslib>=2.39.2",
     "ecmwf-opendata>=0.3.29",
+    "icechunk>=2.0.0; python_version>='3.12'",
```
negin513, line 64: PR description says `icechunk>=1.0.0` but this pins `>=2.0.0` — worth aligning so users reading the description know what they're actually getting.

negin513: And I think since it brings in icechunk we should update the changelog?  @NickGeneva

earth2studio/data/dynamical.py
```diff
@@ -0,0 +1,880 @@
...
+        # regional domain (e.g. DWD ICON-EU, lon -23.5..62.5) into two spatially
+        # discontiguous halves. Only global collections are supported today; when
+        # adding regional ones, make this wrap conditional on global coverage and
+        # keep the native contiguous order otherwise.
+        lon_mod = lon % 360
+        self._lon_idx = np.argsort(lon_mod, kind="stable")
+        self._lat = lat[self._lat_idx]
+        self._lon = lon_mod[self._lon_idx]
+
+    def _reorder(self, arr: np.ndarray) -> np.ndarray:
```
negin513: The `_setup_grid` / `_reorder` pair is doing something other data sources will likely need too. I'd suggest pulling this out into reusable utilities in `data/utils.py` rather than keeping it private to `_DynamicalBase`. Something like:

```python
def normalize_latitude(da, ascending=False, lat_dim=None):
    lat_dim = lat_dim or _find_dim(da, ("latitude", "lat"))
    return da.sortby(lat_dim, ascending=ascending)

def normalize_longitude(da, lon_range=(0.0, 360.0), lon_dim=None):
    lon_dim = lon_dim or _find_dim(da, ("longitude", "lon"))
    start = lon_range[0]
    wrapped = (da[lon_dim].values - start) % 360.0 + start
    return da.assign_coords({lon_dim: wrapped}).sortby(lon_dim)

def normalize_lat_lon(da, lat_ascending=False, lon_range=(0.0, 360.0), lat_dim=None, lon_dim=None):
    da = normalize_latitude(da, ascending=lat_ascending, lat_dim=lat_dim)
    da = normalize_longitude(da, lon_range=lon_range, lon_dim=lon_dim)
    return da
```

A few benefits of this shape:
- Works on both `xr.DataArray` and `xr.Dataset`
- Auto-detects `latitude`/`lat` and `longitude`/`lon` so it isn't tied to dynamical.org's naming
- Parameterized — callers that need `(-180, 180)` or ascending lat can pass that in
- `normalize_latitude` and `normalize_longitude` are independently useful

Then `_setup_grid` becomes a one-liner (`ref = normalize_lat_lon(ref)`) and the fetch loop just calls `normalize_lat_lon(da)` after `.sel()` instead of `_reorder`.

negin513: And @NickGeneva I want to suggest us overall adding some data utils that would be reusuable? similar for da, obs utils, also for regrid, pressure interpol, etc?

pyproject.toml
NickGeneva: Ah yeah, greptile said 2.0 had some APi changes so I needed to bump to 2.0. Let me update this and the changelog. good spot

mrshll:
Thanks @NickGeneva !!

earth2studio/data/dynamical.py
NickGeneva: I dont think this too much an issue, rather I'm not sure its worth adding in thread lock / supporting multi-threading through the same instance... Feels a bit too much of an edge case?

Idk, I'm just not too much of a fan for putting thread locks and multi-processing everywhere, thoughts?

earth2studio/data/dynamical.py
NickGeneva: Hmmm , let me see if I can refactor this a little, I'm not too much of fan of passing a dictionary in regardless.

earth2studio/data/dynamical.py
NickGeneva: Good point, let me see about this

earth2studio/data/dynamical.py
NickGeneva: I propose we push this to another dedicated PR. Good idea, but rather have it isolated.

NickGeneva: Also maybe after the cupy integration? Ideally, cupy is soon^tm i promise

earth2studio/data/dynamical.py
NickGeneva: Right , yeah this is how its done for the other data sources where everything is hard-coded / deterministic.

Unless this happens to change and we get a bug, I vote we just keep it hard coded and then update the description. Similar to just hardcode :)

earth2studio/data/dynamical.py
NickGeneva: Oh, seems to me this should actually be converting to UTC time before stripping the tzinfo rather than just a string replacement which would just drop any offset.

Will fix, good spot

earth2studio/data/dynamical.py
NickGeneva: Repeat code here is okay, its for the doc strings / API docs. Skipping.

earth2studio/data/dynamical.py
NickGeneva: Fixed!

earth2studio/data/dynamical.py
NickGeneva: added a more descriptive error

pyproject.toml
NickGeneva: Updated

earth2studio/data/dynamical.py
NickGeneva: Updated!

earth2studio/data/dynamical.py
NickGeneva: Fixed.

earth2studio/data/dynamical.py
NickGeneva: Updated!

NickGeneva:
@greptile-apps

test/data/test_dynamical.py
```diff
@@ -0,0 +1,463 @@
...
+def _patch(monkeypatch, klass, collection_id, dims, dataset):
+    """Patch STAC fetch + Icechunk open to serve synthetic data (no network)."""
+
+    def fake_fetch_json(url: str) -> dict:
+        if url.endswith("catalog.json"):
+            return _fake_catalog(collection_id)
+        return _fake_collection(dims)
+
+    monkeypatch.setattr("earth2studio.data.dynamical._fetch_json", fake_fetch_json)
+    monkeypatch.setattr(klass, "_open_icechunk", lambda self, asset: dataset)
```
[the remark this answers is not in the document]
NickGeneva: Fixed.

pzharrington (approved):
LGTM!

PR #982, opened 2026-07-21
Support DA models in eval recipe

pzharrington:
# Earth2Studio Pull Request

## Description

Adds evaluation support for `AssimilationModel`-class models (e.g. HealDA) to the eval recipe, plus a small robustness fix to the UFS observation sources in core.

## New capabilities

- **Two DA evaluation modes**, both reusing the existing predownload → inference → score → report flow:
  - **Analysis mode** — score a DA analysis directly against reanalysis.
  - **Forecast mode** — use a DA analysis to initialize a prognostic rollout (e.g. HealDA + DLWP/FCN3), directly comparable to the same prognostic run from reanalysis ICs.
- **Observation predownload** — DA obs (DataFrames, not gridded fields) are cached to disk and transparently reused at inference, so `main.py` needs no live access to obs archives.

## New pipelines (`src/pipelines/assimilation.py`)

- `AssimilationPipeline` — analysis-as-product; writes a singleton `lead_time=0` slice so existing scoring/report apply unchanged.
- `AssimilationForecastPipeline` — subclasses `ForecastPipeline`, overriding only IC acquisition (one DA analysis per prognostic input-lead offset; history models get one per offset). Prognostic inputs the DA model doesn't produce are filled from the standard IC source.

Both set `needs_data_source = False` and resolve their own inputs, so `main.py`/`predownload.py` needed no changes.

## New interfaces

- **`ObsSourceSet`** (`src/assimilation.py`) — resolves N observation `DataFrameSource`s positionally against a model's `input_coords()` tuple; supports `enabled: false` slots and `time_tolerance_hours`.
- **`AssimilationRunner`** seam — `StatelessAssimilationRunner` (default) for models like HealDA; stateful/cycling models plug in via an optional `model.da.runner` block (the stateless runner rejects stateful models with a clear error). This is the generality hook for future StormCastSDA-style models.
- **`PredownloadFrameStore`** + `Pipeline.predownload_frame_stores()` hook — the DataFrame twin of `PredownloadStore`; declares obs stores for `predownload.py` to populate.
- **`PredownloadedFrameSource`** (`src/data.py`) — serves a predownloaded parquet obs store through the `DataFrameSource` protocol; pipelines auto-prefer it over the live source when present.

## Config surface

- New model configs: `healda`, `healda_ucast`; campaigns: `healda_2024_analysis`, `healda_ucast_2024_monthly`
- New `predownload.observations.enabled` (default `true`) to toggle obs caching.
- No changes to existing model/campaign configs or run commands.

## Core change (`earth2studio/data/ufs.py`)

- `UFSObsConv`/`UFSObsSat` now tolerate missing diag files by warning-and-skipping instead of erroring (archive gaps — e.g. absent GNSS-RO cycles or decommissioned platforms — no longer abort a fetch). Behavior lives on the shared base class; a fully-empty request returns a schema-shaped empty frame instead of raising. The recipe surfaces an all-empty predownload as a loud error (likely out-of-range dates).

## Tests

- `recipes/eval/test/test_assimilation.py` — obs source set, runners, conversion helpers, both pipelines end-to-end, frame-store declaration/round-trip, predownload writer + empty-frame handling (no network/GPU).
- `test/data/test_ufs.py` — missing-file tolerance and empty-frame return for both obs sources.

## Checklist


pzharrington:
@greptileai

pzharrington:
Example scores for HealDA+UCast end-to-end DA+forecast workflow:
image

earth2studio/data/ufs.py
```diff
@@ -277,6 +291,18 @@ def _compile_dataframe(
             df = df.loc[mask]
             frames.append(task.gsi_modifier(df))
 
+        if not frames:
+            # Every diag file for this request was missing or skipped (see
```
NickGeneva: nit: little verbose inline comments

earth2studio/data/ufs.py
```diff
@@ -186,10 +186,24 @@ async def _fetch_remote_file(
             self._handle_missing_file(key)
 
     def _handle_missing_file(self, key: str) -> None:
-        """Handle missing file during fetch. Can be overridden by subclasses."""
+        """Handle a missing diag file during fetch.
+
+        GSI diag archives have gaps: a given observation platform or
```
NickGeneva: nit: very verbose, can this be made more concise?

recipes/eval/src/pipelines/assimilation.py
```diff
@@ -0,0 +1,416 @@
...
+    analysis per entry in the prognostic's input ``lead_time`` axis
+    (single-time models need one; history models get one per offset).
+
+    Prognostic input variables the DA model does not produce are filled
+    from the standard IC path (``cfg.ic_source`` BYO → predownloaded
+    ``data.zarr`` → ``cfg.data_source``), unless
+    ``cfg.model.fill_missing_variables`` is set to ``false`` — in which
+    case any gap is an error.
+
+    Model config layout (see ``cfg/model/healda_fcn3.yaml``)::
```
NickGeneva: should this be ucast?

recipes/eval/src/pipelines/assimilation.py
```diff
@@ -0,0 +1,416 @@
...
+            )
+            coords_fill = OrderedDict(coords_fill)
+            coords_fill["time"] = coords["time"]
+            x, coords = cat_coords((x, x_fill), (coords, coords_fill), "variable")
+
+        x, coords = map_coords(x, coords, self._prognostic_ic)
+        return x, coords
+
+    _predownload_prognostic = None
+    """Prognostic model memoized across the predownload hooks —
```
NickGeneva: nit: odd docstring placement?

recipes/eval/src/assimilation.py
```diff
@@ -0,0 +1,479 @@
...
+        self._obs_device = obs_device
+
+    def analysis(self, time: np.datetime64) -> xr.DataArray:
+        obs = self._obs_set.fetch(time, device=self._obs_device)
+        if all(o is None for o in obs):
+            raise ValueError(
+                "All observation sources are disabled — at least one must "
+                "be enabled to produce an analysis."
+            )
+        return self._model(*obs)
```
NickGeneva: claude picked this up, if the model is a DA model it might return a data array or data frame, but looks like this method is data array only?

recipes/eval/src/predownload_utils.py
```diff
@@ -42,6 +45,78 @@
...
+        target_lon: np.ndarray,
+    ) -> None:
+        self._source = source
+        self._lat = xr.DataArray(np.asarray(target_lat), dims="lat")
+        self._lon = xr.DataArray(np.asarray(target_lon), dims="lon")
+
+    def __call__(self, time: object, variable: object) -> xr.DataArray:
+        da = self._source(time, variable)  # type: ignore[operator]
+        if "lat" in da.dims and "lon" in da.dims:
+            da = da.interp(lat=self._lat, lon=self._lon, method="linear")
```
NickGeneva, line 69: we have some interp utils in the package, i guess its not useful here since this data is on the CPU?

CHANGELOG.md
```diff
@@ -32,29 +27,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-  FTP); no fsspec fallbacks remain in the GRIB byte-range sources
 - Refactored UFS observation sources (`UFSObsConv`, `UFSObsSat`) onto the shared
   obstore byte-range helpers
 - Zarr-reading data sources (`ARCO`, `WB2ERA5` and other WeatherBench 2 sources, and
   the `rx` prescriptive sources) now read via `obstore`-backed zarr stores instead of
   fsspec
-- Updated the OPERA data source to represent undetect values as `-99.0`, while
-  retaining `NaN` for no-data values.
-- NNJA Obs data source now accepts any time / tolerance rather than 6-hour strides
+- UFS observation sources (`UFSObsConv`, `UFSObsSat`) now tolerate missing diag files
```
NickGeneva, line 51: nit: this can be made just the first line. verbose

NickGeneva (approved):
Generally minor comments, approving to unblock

recipes/eval/src/assimilation.py
pzharrington: True, good catch. Actually our `AssimilationModel` protocol specifies the return type `tuple[pd.DataFrame | xr.DataArray, ...]` but **all current DA models return one raw `xr.DataArray`** which is probably why my Claude went in this direction assuming that return type 🙂

I'm still struggling to recall why we thought some DA models would map `DataFrame` to `(DataFrame,)`, but given all of them at the moment don't have that return type I'm inclined to do the following:

- Change AssimilationRunner.analysis() + StatelessAssimilationRunner.analysis() return annotations to tuple[pd.DataFrame | xr.DataArray, ...]
- Rewrite analysis_to_tensor to extract the first xr.DataArray from the tuple

This incurs zero changes in either pipeline — the call sites are unchanged, and no model changes for now. But it is def not ideal all our DA models don't obey the `AssimilationModel` protocol haha, to be addressed in follow-ups?

recipes/eval/src/predownload_utils.py
pzharrington: Yeah CPU only here

PR #1025, opened 2026-07-29
Async zarr sharding

pzharrington:
# Earth2Studio Pull Request

## Description
Adds sharding support to `AsyncZarrBackend`. Strategy, when enabled, is to permit sharding without blocking GPU work: let the chunks for a given shard accumulate in host memory and then flush/write when ready. Still handled by threads so the GPU can continue making forecasts and the backend remains async. Shard size would thus be limited by host memory capacity, best settings should be tuned per problem/system specifics.

Tested on an example problem with fairly realistic numbers: ~300 ms simulated model forward pass time, saving 73x721x1440 arrays each lead time, saving the lead times out with shard size of 8 and chunk size 1 (`write` called each forecast step, as they're generated). Results gave basically no overhead with sharding enabled, and reduced output file count by factor of 8 in the zarr store 🚀 

## Checklist


earth2studio/io/async_zarr.py
```diff
@@ -332,6 +568,77 @@ async def _initialize_arrays(
...
+
+            shard_size = self.shard_coords[key]
+            if shard_size % chunk_size != 0:
+                raise ValueError(
+                    f"Shard size {shard_size} for coordinate '{key}' of array '{name}' "
+                    + f"must be a multiple of its chunk size {chunk_size}. "
+                    + "Coordinates in `parallel_coords` have a chunk size of 1, others "
+                    + "default to their full length unless set via `chunked_coords`."
+                )
+            dim_size = array_coords[key].shape[0]
```
negin513, line 639: I can crash this with a config that validates fine. If a coord isn't in `parallel_coords`, `_write` always sends `slice(None)` for it, and `_stage_chunk` maps that to one shard (`s.start // h` = 0) and copies into that shard's buffer. So `chunked_coords={"lat": 8}` + `shard_coords={"lat": 16}` with lat of length 32 gets past this check (16 is a multiple of 8), then dies with a numpy broadcast error when the 32-wide slice hits a 16-wide buffer.

The existing chunked_coords test uses lat=16 with shard 16, which is the one case where it works. I'd just error out here when shard < dim length for non-parallel coords. Splitting the slice across shards would be the proper fix but probably not worth it until someone actually wants that layout. I have a test that reproduces the crash if you want it.

earth2studio/io/async_zarr.py
```diff
@@ -441,19 +748,311 @@ async def prepare_inputs(
...
+            current.set_result(None)
+
+    async def async_flush(self) -> None:
+        """Writes out every shard buffer that is still incomplete
+
+        Shards that never filled are written with the array fill value in the positions
+        that were never supplied, which reads back identically to an unwritten chunk.
+        """
+        with self._shard_lock:
+            pending = list(self._shard_buffers.items())
```
negin513, line 1038: There's a race here if someone calls this while an async_write is still going. Sync `flush()` drains writes first but this doesn't, and the buffers get cleared here while `_flushed_shards.add()` only happens later in the loop. If `_stage_chunk` lands in between, it finds no buffer, key not in `_flushed_shards`, nothing on disk yet -- so it opens a new buffer with `preexisting=False` and eventually overwrites whatever this flush wrote.

Easy fix: claim the buffers and add them to `_flushed_shards` under the same lock that clears the dict.

Also, these flushes run one at a time -- could just gather them, the slot gate already limits how many run at once. Would make close() faster when a few partial shards are left over.

earth2studio/io/async_zarr.py
```diff
@@ -441,19 +748,311 @@ async def prepare_inputs(
...
+        # flush on one thread cannot stall staging on another
+        with self._shard_lock:
+            buffer.mark(array_slice)
+            if buffer.remaining > 0 or buffer.claimed:
+                return
+            buffer.claimed = True
+            self._shard_buffers.pop(key, None)
+            self._flushed_shards.add(key)
+
+        await self._flush_buffer(name, zarray, key, buffer)
```
negin513: Something to think about: before this PR the `async_timeout` in `write()` covered one chunk write, now the write that completes a shard carries the whole flush (plus maybe waiting for a slot, plus maybe waiting on an earlier flush of the same shard). Big shard on a slow store and 600s isn't that much anymore.

And if it does time out, it's messy -- the buffer is already claimed and in `_flushed_shards`, so the data's gone from memory and later writes to that shard go down the merge path against a file that doesn't exist. You can recover by rewriting every chunk of the shard, but someone retrying just the failed step won't get their data back. Maybe just a note in the `async_timeout` docstring saying to scale it with shard size.

earth2studio/io/async_zarr.py
```diff
@@ -128,9 +334,36 @@ def __init__(
         self.overwrite = False  # Not formally supported
         self.parallel_coords = self._scrub_coordinates(parallel_coords)
         # Parameter to also chunk some of the other dims if needed
-        self.chunked_coords: dict[str, int] = {}
+        self.chunked_coords: dict[str, int] = dict(chunked_coords)
+        self.shard_coords: dict[str, int] = dict(shard_coords)
```
negin513, line 351: If a key in `shard_coords` doesn't match any coordinate (say someone types "leadtime"), everything silently comes out unsharded. Since sharding is the whole reason to set this, a warning for unmatched keys would save someone a confusing afternoon.

earth2studio/io/async_zarr.py
```diff
@@ -441,19 +748,311 @@ async def prepare_inputs(
...
+                    logger.debug(
+                        f"Preceding flush of shard {key[1]} of array {name} failed "
+                        + f"({e}), continuing with this flush"
+                    )
+
+            # Admitted only after the same shard ordering wait above, so a flush
+            # never holds a slot while waiting on another flush that needs one
+            await self._acquire_flush_slot(current)
+
+            if buffer.preexisting:
```
negin513, line 1008: nit: a preexisting shard whose buffer ends up fully filled gets written in one go with no read, but still logs the read-modify-write warning. Could gate it on `buffer.remaining > 0`.

earth2studio/io/async_zarr.py
```diff
@@ -619,15 +1220,21 @@ def close(self) -> None:
...
-        # Clean up process pool
-        self._limit_pool_size(0)
+        # Clean up process pool and write out any partially filled shards
+        self.flush()
 
     def __del__(self) -> None:
         if not hasattr(self, "io_futures"):
             return
-        if len(self.io_futures) > 0:
+        if len(self.io_futures) > 0 or len(self._shard_buffers) > 0:
```
negin513, line 1294: nit: `__del__` now goes through `flush()`, which actually writes partial shards instead of just draining futures. At interpreter shutdown the loop threads might already be dead, so this can fail weirdly. Probably fine since the warning tells people to call close(), just flagging that GC-time writes are best-effort.

earth2studio/io/async_zarr.py
```diff
@@ -65,6 +203,28 @@ class AsyncZarrBackend:
...
+    per process, since a flushing shard costs several times its own size once Zarr's
+    encoded copy is counted. Sharded writes are also slower than unsharded ones, which
+    is hidden as long as the model takes longer to produce a step than the store takes
+    to absorb it.
+
+    Warning
+    -------
+    When sharding, a shard must not contain data owned by more than one process. This
+    backend keeps every shard object to a single write by buffering its chunks, but that
+    only holds within a process, seperate ranks have seperate buffers. If two ranks each
```
negin513: typos: "seperate" x2 here, "campagins" in io.md, and a double space in "the `AsyncZarrBackend`,  none" (also io.md). Might also be worth a line somewhere that with `blocking=True` the flush is synchronous, so the "hidden behind model time" argument only holds in non-blocking mode.

negin513 (commented):
Nice work, this turned out really clean. I spent a while trying to poke holes in the concurrency and couldn't -- the lock never spans an await, and since data gets copied into the buffer before `mark()` is called, a flush can't catch a copy halfway. Also happy you fixed the swallowed exceptions in `_limit_pool_size`, that one affects people who never touch sharding.

FWIW I reran your benchmark setup on hsg too. 
Found a couple of I think minor issues, see inline.

earth2studio/io/async_zarr.py
negin513: Does it worth adding some more information on sharding best practices? Maybe for future PRs?

earth2studio/io/async_zarr.py
negin513: minor nit only.

earth2studio/io/async_zarr.py
negin513: Also another minor nit.

earth2studio/io/async_zarr.py
negin513: this is just minor nit too. Mostly adding a warning should resolve it. I think zarr also give out an error too.

negin513 (approved):
Marking this as approved. My only comment to double-check is around the async_flesh and pending writes.  All looks great otherwise.

earth2studio/io/async_zarr.py
pzharrington: `_compute_shards` now raises `ValueError` when a non-parallel coordinate has a shard size smaller than the dimension — these coords are always written as `slice(None)` so splitting across shards isn't supported

earth2studio/io/async_zarr.py
pzharrington: Good catch, this now adds all keys to `_flushed_shards` under the same lock that clears `_shard_buffers`, so any concurrent `_stage_chunk` that finds no buffer will see the key as flushed (and use the merge/chain path instead of creating a fresh `preexisting=False` buffer) Flushes are now gathered concurrently instead of running one at a time

earth2studio/io/async_zarr.py
pzharrington: Added a note to the `async_timeout `parameter docstring explaining that shard-completing writes carry the entire flush, so the timeout should be scaled with shard size

earth2studio/io/async_zarr.py
pzharrington: `_compute_shards` now warns when a `shard_coords` key doesn't appear in the array's coordinates

earth2studio/io/async_zarr.py
pzharrington: Ok, added the gate

earth2studio/io/async_zarr.py
pzharrington: Yeah I'm fine with just relying on the docstring/warning for users to call `close()` for now, hopefully won't be a major issue

earth2studio/io/async_zarr.py
pzharrington: Thanks, typos addressed and the nonblocking requirement is clarified. I think these docs should cover the basic usage for now and we can have more detailed recommendations emerge as people use it

earth2studio/io/async_zarr.py
```diff
@@ -441,19 +767,318 @@ async def prepare_inputs(
...
+                    + "count. Consider a smaller shard size or writing coordinates in "
+                    + "an order that completes shards sooner."
+                )
+
+        # Copy outside the lock, chunk regions are disjoint and numpy releases the GIL
+        local_slice = tuple(
+            slice(s.start - r.start, s.stop - r.start)
+            for s, r in zip(array_slice, buffer.region)
+        )
+        buffer.data[local_slice] = data
```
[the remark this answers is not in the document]
pzharrington: Good catch. Addressed with following changes: 

`_ShardBuffer` — two new fields:

- `pending_copies: int` — count of numpy copies into this buffer that have started but not yet called `mark()`. `async_flush` reads this under the lock.
- `pending_flush: bool` — set by `async_flush` when it claims a buffer that still has copies in flight. Tells the last completing copy to trigger the write.

`_stage_chunk` — **Phase 1 (get/create buffer)**: now increments `pending_copies` under the same lock that gets or inserts the buffer. This ensures `async_flush` can never observe `pending_copies == 0` while a copy is logically in flight.

`_stage_chunk` — **Phase 3 (mark and decide)**: after decrementing `pending_copies` and calling `mark()`, there are now three cases:

- `buffer.claimed` and `pending_copies == 0` and `pending_flush` → `async_flush` delegated the flush to us, flush now
- `buffer.claimed` otherwise → another thread owns the flush, return
- `remaining == 0` and not claimed → normal completion path, flush now

`async_flush`: for each buffer in the snapshot, sets `claimed = True` and adds to `_flushed_shards` (the atomicity fix from the previous PR comment). If `pending_copies == 0`, safe to flush immediately. If `pending_copies > 0`, sets `pending_flush = True` — the last copy will flush it.

PR #1035, opened 2026-07-31
Fix OPERA fill values

### LINKED ISSUE #1031: 🐛[BUG]: OPERA returns negative precipitation for pixels where the radar detected no rain
### Version

0.18.0a0

source

### Describe the issue

ODIM (Opera Data Information Model) is the HDF5 format these radar files
use. Every pixel is one of three things:

1. **Outside radar coverage** - no radar looks there. ODIM calls this `nodata`.
2. **Radar looked, saw nothing** - dry. ODIM calls this `undetect`.
3. **Radar measured something** - a real value.

Earth2Studio maps `nodata` to NaN (correct) and `undetect` to **-99.0** for
every quantity. For reflectivity (`refc`, dBZ) a negative marker is at least
plausible. For precipitation it is not:

| variable | ODIM quantity | what the caller gets | should be |
|---|---|---|---|
| `tprate` | RATE (mm/h) | **-99.0 mm/h** | 0 |
| `tp01` | ACRR (m) | **-0.099 m** | 0 |
| `refc` | DBZH (dBZ) | -99.0 dBZ | see below |

`tp01` is converted mm→m after the fill, so the sentinel arrives as `-0.099`
and anyone filtering on `== -99` misses it. Dry is the normal state, so this
hits roughly half of every returned field.

### Reproduction

```python
import numpy as np
from datetime import datetime
from earth2studio.data import OPERA

ds = OPERA(cache=False)
d = ds(datetime(2026, 7, 28, 12, 0), ["tprate", "tp01"])
for v in ["tprate", "tp01"]:
    a = np.asarray(d.sel(variable=v).values).squeeze()
    frac = np.isclose(a, -99.0).mean() + np.isclose(a, -0.099).mean()
    print(f"{v:7s} min={np.nanmin(a):9.4g}  mean={np.nanmean(a):9.4g}  filled={frac:.1%}")
```

```console
$ uv run python repro.py
tprate  min=    -99  mean=  -92.67  filled= 47.4%
tp01    min= -0.099  mean= -0.0881  filled= 45.0%
```

The archive is public (CloudFerro S3, no credentials).

### Where it comes from

[`opera.py` L197-198](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/data/opera.py#L197-L198)
hard-codes a reflectivity fill, and
[`_apply_linear_scaling` L241-L252](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/data/opera.py#L241-L252)
applies it to every quantity:

```python
# MRMS refc stores no-echo as -99 dBZ; fill OPERA no-detection as -99.0 to match.
_NO_DETECTION_FILL: float = -99.0
...
if undetect is not None:
    result[raw == undetect] = cls._NO_DETECTION_FILL
```

The
[docstring](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/data/opera.py#L161-L163)
states the same thing in dBZ units for all three variables. That reasoning was
aimed at `refc`. It does not transfer to `tprate` and `tp01`.

<details>
<summary>Why the MRMS justification does not hold up</summary>

MRMS uses -99 to mean **missing**, not "no echo"; its rain-rate product never
uses -99 in a mm/h field; `mrms.py` performs no sentinel translation; and this
repo's eval recipe documents the MRMS fill as `-9999`
(`recipes/eval/src/scoring.py:791`). The string `-99` appears nowhere else in
the repository except OPERA's own test.

</details>

### Why 0 is the right answer for RATE and ACRR

Inspecting the same file as the reproduction (2026-07-28 12:00Z), the lowest
real value in both precipitation quantities is already exactly 0:

```
RATE   undetect=-8888000.0  valid min=0    max=193.6   exact-zero valid px=2616
ACRR   undetect=-8888000.0  valid min=0    max=123.8   exact-zero valid px=12490
DBZH   undetect=-8888000.0  valid min=-32  max=69.5    exact-zero valid px=8902
```

So 0 is an in-range, already-present value for RATE and ACRR; a negative one
is not.

That matches the surrounding toolchain:

- BALTRAD RAVE encodes a zero accumulation *as* `undetect` in
  [`rave_acrr.c`](https://github.com/baltrad/rave/blob/master/librave/toolbox/rave_acrr.c)
  (`if (acrr <= 0.0) … setValue(..., self->undetect)`).
- [pysteps](https://github.com/pySTEPS/pysteps/blob/master/pysteps/io/importers.py)
  already branches: `offset` (0.0) for RATE/ACRR, `-30.0` for DBZH.
- ODIM_H5 v2.4 Table 13: `undetect` is "below the measurement detection
  threshold". In a mm/h or mm field, that is zero.

<details>
<summary>Nothing downstream catches it; the tests pin the bug</summary>

`prep_data_array` / `fetch_data` have no range checks. The eval recipe's
`valid_ranges` clamp covers `refc` / `refc_base` / `glm_density` but not
`tprate` / `tp01`. Threshold scores (FSS, Brier) happen to bin -0.099 with 0;
mean, RMSE, ACC and moments absorb the bias.

The only test that sets `undetect` uses reflectivity packing
(`gain=0.5, offset=-32.0`) and asserts `_NO_DETECTION_FILL == -99.0`. RATE and
ACRR tests never decode a value. Network tests are `@pytest.mark.xfail` with
`xfail_strict` unset.

</details>

## Expected behaviour

A pixel where the radar detected no precipitation should read as 0 mm/h and 0 m,
not as a negative value.

## Suggested fix

Make the fill depend on the quantity. Both decode branches already build
`what` from the matched group, so `what["quantity"]` is present inside
`_apply_linear_scaling`. This is a lookup in the existing function body, not a
signature change:

```python
_UNDETECT_FILL: dict[str, float] = {"RATE": 0.0, "ACRR": 0.0}
```

with `DBZH` falling back to the existing constant unless a maintainer decides
otherwise. Also update the L161-L163 docstring (the `nodata` gloss is wrong:
it says "no active radar echo", but `nodata` means never radiated / outside
coverage), add RATE/ACRR test cases, and note the behaviour change in the
CHANGELOG.

pzharrington:
# Earth2Studio Pull Request

## Description
Closes #1031 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed `OPERA` data source returning negative precipitation values (`-99.0 mm/h`
+  for `tprate`, `-0.099 m` for `tp01`) for pixels where the radar detected no rain.
+  Undetect pixels for RATE and ACRR quantities are now filled with `0.0` instead
+  of the reflectivity sentinel `-99.0 dBZ` (#1031).
 - Fixed `GHCNHourly` station discovery to use the published GHCNh station
   list (`ghcnh-station-list.csv`) instead of the GHCN-Daily station list.
 - Fixed `AIFS2` and `AIFS2ENS` assigning time-dependent forcing values to the wrong
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed `OPERA` data source returning negative precipitation values (`-99.0 mm/h`
   for `tprate`, `-0.099 m` for `tp01`) for pixels where the radar detected no rain.
   Undetect pixels for RATE and ACRR quantities are now filled with `0.0` instead
-  of the reflectivity sentinel `-99.0 dBZ` (#1031).
+  of the reflectivity sentinel `-99.0 dBZ`.
 - Fixed `GHCNHourly` station discovery to use the published GHCNh station
   list (`ghcnh-station-list.csv`) instead of the GHCN-Daily station list.
 - Fixed `AIFS2` and `AIFS2ENS` assigning time-dependent forcing values to the wrong
--- a/earth2studio/data/opera.py
+++ b/earth2studio/data/opera.py
@@ -158,9 +158,12 @@ class OPERA:
     pixel resolutions, a :exc:`ValueError` is raised — request each resolution
     group separately.
 
-    OPERA undetect values (active radar echo but no detectable precipitation) are
-    set to -99.0 dbZ, and nodata values are set to NaN (no active radar echo).
-    All other values are scaled by the gain and offset parameters in the ODIM HDF5 file.
+    ODIM pixels with no radar coverage (``nodata``) are set to NaN.  Pixels
+    where the radar was active but detected nothing (``undetect``) are filled
+    with a quantity-specific value: ``0.0`` for precipitation quantities (RATE,
+    ACRR) and ``-99.0 dBZ`` for reflectivity (DBZH) to match the MRMS refc
+    convention.  All other values are scaled by the gain and offset parameters
+    in the ODIM HDF5 file.
 
     Parameters
     ----------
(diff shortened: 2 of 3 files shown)


NickGeneva (approved):
lgtm

PR #972, opened 2026-07-15
Migrate remaining GRIB byte-range sources (HRRR, GEFS, CFS, NCAR ERA5) to obstore

negin513:
# Earth2Studio Pull Request

## Description

Completes the GRIB byte-range obstore migration ([physicsnemo-roadmap#2747](https://github.com/NVIDIA/physicsnemo-roadmap/issues/2747)) started in #964.

> **Stacked on #964** — this branch is based on `obstore-gfs`; the first commits shown here belong to that PR. Review only the three commits from `Migrate remaining GRIB byte-range sources...` onward; will rebase onto `main` once #964 merges.

### What changed

* Migrated the four remaining GRIB byte-range sources from s3fs/gcsfs to obstore via the shared `data/utils.py` helpers, following the GFS reference implementation:
  * `HRRR` / `HRRR_FX` — all three backends: S3 (`noaa-hrrr-bdp-pds`), the GCS mirror (`high-resolution-rapid-refresh`), and NOMADS over obstore `HTTPStore`.
  * `GEFS_FX` / `GEFS_FX_721x1440` — including a sentinel-free index parser where the trailing GRIB record is expressed directly as `byte_length=None` (read-to-EOF).
  * `CFS_FX` / `CFS_FX_Flux` — both `aws` and `nomads` sources.
  * `NCAR_ERA5` — with a new `_ObstoreBlockCachedIO` adapter serving h5py's scattered reads from LRU-cached 8 MB ranged GETs, and an explicit `us-west-2` region (obstore does not follow S3 region redirects; the naive port failed outright).
* **No fsspec fallbacks remain**: the GFS `ncep` source (fsspec FTP in #964) and the HRRR/CFS `nomads` sources now use obstore `HTTPStore` against the NOMADS HTTPS endpoint. None of the five modules import fsspec/s3fs/gcsfs.
* Download concurrency now autoscales: `async_workers=None` (new default) scales workers to the number of pending download tasks, capped at 64, via a shared `resolve_async_workers` helper; an explicit value still pins it. `retries` added to all migrated sources.
* `available()` probes use obstore listings/head requests instead of `s3fs.exists()`.

### Behavior notes

* Outputs are unchanged — every migrated source was verified **bit-identical** to its fsspec implementation against live stores.
* Existing warm caches remain valid (cache keys still hash the historical bucket-prefixed URIs).
* GFS `ncep` moves from FTP to HTTPS; environments that block outbound FTP (where the old source could not connect at all) now work.

[...]

PR #964, opened 2026-07-13
Migrate GFS byte-range GRIB fetches to obstore with shared helpers

negin513:
# Earth2Studio Pull Request

## Description

Implements the first phase of the GRIB byte-range obstore migration ([physicsnemo-roadmap#2747](https://github.com/NVIDIA/physicsnemo-roadmap/issues/2747)): shared byte-range utilities and the migration of `GFS` / `GFS_FX` as the reference implementation. HRRR, GEFS, and CFS will follow in subsequent PRs.

### What changed

* Added shared obstore utilities in `earth2studio.data.utils`:
  * `obstore_store_from_url(url, ...)` — creates stores for `s3://`, `gs://`, and `https://` URLs, including `skip_signature` configuration for anonymous access to public buckets.
  * `obstore_read_range(store, key, byte_offset, byte_length)` — supports ranged, whole-object, and read-to-EOF requests. Obstore `NotFoundError` exceptions are translated to `FileNotFoundError` to preserve existing caller behavior.
  * `obstore_fetch_to_cache(...)` — fetches objects into a SHA-256-keyed on-disk cache. An optional `cache_key` override allows migrated sources to preserve their existing cache-key scheme.
* Migrated `GFS` / `GFS_FX` `.idx` and byte-range GRIB reads from s3fs to obstore, removing the associated s3fs session management.
  * The `ncep` source remains on fsspec FTP because obstore does not provide an FTP backend.
* Updated the GFS fetch path to follow the CFS execution pattern:
  * bounded concurrency through `async_workers`,
  * per-task retries for transient failures through `retries`, and
  * `pygrib` decoding offloaded to a cancellation-aware worker thread.
* Refactored `UFSObsConv` / `UFSObsSat` to use the shared utilities instead of maintaining separate `S3Store` and `get_range_async` implementations.

This leaves a single shared implementation for obstore-backed byte-range reads.

### Behavior notes

* Outputs are unchanged.
* Existing warm caches remain valid because cache keys continue to hash the bucket-prefixed URI.
* `GFS` / `GFS_FX` now expose:
  * `async_workers=16`
  * `retries=3`
* `GFS.available()` now uses an obstore prefix listing instead of `s3fs.exists()`.

### Testing

* Added offline mock coverage for `GFS` / `GFS_FX`, following the existing CFS test pattern. These paths were previously covered only by network-dependent tests.
* Added unit tests for the three shared obstore utilities.
* Completed live smoke testing against `noaa-gfs-bdp-pds`, including:
[...]

PR #1042, opened 2026-08-04
Migrate GOES and GOES GLM data sources to obstore

negin513:
# Earth2Studio Pull Request

## Description

Migrates the GOES ABI (`GOES`) and GOES GLM (`GOESGLM`) data sources from `s3fs` to `obstore`, completing the GOES-family migration and following the patterns established in #955, #964, and #972.

The PR also:

- Adds asynchronous, memoized prefix listing for GOES and GLM.
- Eliminates redundant LIST requests and event-loop blocking.
- Introduces a reusable listing helper for future data-source migrations.
- Adopts [`obspec`](https://github.com/developmentseed/obspec) protocol typing for the shared object-store utilities.

## GOES ABI

- Whole-file NetCDF downloads and object listings now use the shared `obstore` helpers in `data/utils.py`.
- Cache keys remain derived from the historical bucket-prefixed paths, preserving compatibility with caches populated before this migration.
- Hour-directory listings are now consumed asynchronously. The previous implementation synchronously iterated the listing stream, blocking the event loop while downloads were in flight.
- Listings are memoized by hour prefix, and all unique hours are prefetched concurrently at the start of `fetch()`.

Fetching multiple timestamps from the same completed hour now requires one LIST request rather than one request per timestamp.

The current, still-filling hour is always re-listed so near-real-time polling can discover newly arriving files.

## GOES GLM

- Migrates GLM L2 LCFA access from `s3fs` to `obstore`.
- Creates one store per satellite bucket, initialized lazily and reused across calls.
- Removes the previous workaround that rebuilt the `s3fs` filesystem and `aiohttp` session on every call, including the `managed_session` wrapper.
- Memoizes listings for completed hour directories.

This is especially useful for `GOESGLMGrid`, which may fetch one five-minute bin at a time. Repeated requests within the same completed hour now share a single LIST operation.

The current hour remains uncached so near-real-time workflows continue to detect new files.

## Shared Prefix-Listing Helper

The asynchronous listing and memoization logic is implemented in:

```python
obstore_list_prefix()
````

The helper:

* Is typed against `AsyncListableStore`.
* Supports a `cacheable` flag for distinguishing completed and still-filling prefixes.
* Consumes each listing asynchronously.
* Centralizes listing-cache behavior for both GOES and GLM.

GOES and GLM now each perform their listing through a single helper call.

Future migrations of listing-based sources such as Himawari, JPSS, and MRMS can adopt the same behavior with minimal source-specific logic.

## `obspec` Protocol Typing

The following shared helpers and store attributes are now typed using [`[obspec](https://github.com/developmentseed/obspec)`](https://github.com/developmentseed/obspec) protocols:

* `GetAsync`
* `GetRangeAsync`
* `HeadAsync`
* `ListAsync`

These are composed into `AsyncReadableStore` and `AsyncListableStore` protocols in `data/utils.py`.

The helpers now call store methods directly rather than invoking `obstore` module-level functions. Because `obstore` stores satisfy these protocols structurally, production behavior is unchanged while tests can inject lightweight fake stores without patching `obstore` internals.

Affected helpers include:

* `obstore_read_range`
* `obstore_fetch_to_cache`
* `obstore_list_prefix`

## Performance

Benchmarks used a cold cache and public NOAA S3. Runs alternated between `main` and this branch to reduce network-ordering bias. Identical downloaded bytes and output arrays or rows were verified for every run.

| Workload                                                         | `s3fs` (`main`) | `obstore` (this PR) | Improvement |
| ---------------------------------------------------------------- | --------------: | ------------------: | ----------: |
| GOES: 12 CONUS timestamps × 16 bands within one hour             |          20.4 s |              13.9 s | ~30% faster |
| GLM: six sequential five-minute bins over a CONUS bounding box   |          24.8 s |               6.1 s |  ~4× faster |
| StormScope data preparation with GOES and GLM initial conditions |          20.8 s |              15.4 s | ~26% faster |

The raw `s3fs`-to-`obstore` transfer change is approximately performance-neutral for GOES because its whole-file downloads are primarily bandwidth-bound. The observed improvement comes from asynchronous listing, concurrent prefetching, and listing reuse.

GLM sees the largest gain because each hour directory contains approximately 180 small files. The previous implementation repeatedly paid both session-initialization and LIST costs for sequential five-minute fetches.

## Testing

* `test/data/test_goes.py`
* `test/data/test_goes_glm.py`
* Existing test suites for other shared-helper consumers also pass:

  * GFS
  * GEFS
  * HRRR
  * CFS

This work also uncovered and fixed an exception-translation issue in the GLM missing-file path.

`obstore_read_range()` exposes missing objects as `FileNotFoundError`, rather than `obstore`’s `NotFoundError`. As a result, the existing warn-and-skip path was not reached, and missing GLM files could raise after retries. The exception handling now correctly preserves the intended skip behavior.

## Checklist

* [x] I am familiar with the [[Contributing Guidelines](https://github.com/NVIDIA/earth2studio/blob/main/CONTRIBUTING.md)](https://github.com/NVIDIA/earth2studio/blob/main/CONTRIBUTING.md).
* [x] New or existing tests cover these changes.
* [x] Documentation is up to date.
* [x] [`[CHANGELOG.md](https://github.com/NVIDIA/earth2studio/blob/main/CHANGELOG.md)`](https://github.com/NVIDIA/earth2studio/blob/main/CHANGELOG.md) is up to date.
* [ ] An [[issue](https://github.com/NVIDIA/earth2studio/issues)](https://github.com/NVIDIA/earth2studio/issues) is linked to this pull request.
* [ ] Greptile feedback has been assessed and addressed where appropriate.

## Dependencies

Adds:

```text
obspec>=0.1
```

`obspec` contains pure-Python protocol definitions. Its only transitive dependency is `typing-extensions` on Python versions earlier than 3.12.

This PR removes the GOES family’s remaining uses of `s3fs`. Other Earth2Studio data sources continue to use `s3fs`.


CHANGELOG.md
```diff
@@ -83,6 +83,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   FTP); no fsspec fallbacks remain in the GRIB byte-range sources
 - Refactored UFS observation sources (`UFSObsConv`, `UFSObsSat`) onto the shared
   obstore byte-range helpers
+- Migrated GOES data source from s3fs to obstore; hour-directory listings are
```
NickGeneva: Merge got messed up, in the wrong section of the change log

pyproject.toml
```diff
@@ -31,6 +31,7 @@ dependencies = [
     "tqdm>=4.65.0",
     "xarray[parallel]>=2023.1.0",
     "zarr>=3.1.3",
+    "obspec>=0.1",
```
NickGeneva: Whats this for? Does it really need to be added into the core dependencies?

earth2studio/data/utils.py
```diff
@@ -834,6 +835,65 @@ def resolve_async_workers(
...
+        Memoization dict keyed by prefix; by default None (no memoization)
+    cacheable : bool, optional
+        Whether the result may be stored in ``cache``, by default True
+
+    Returns
+    -------
+    list[str]
+        Object keys (bucket-relative paths) under the prefix
+    """
+    if cache is not None and prefix in cache:
```
NickGeneva: Is there ever a possibility for cache to be populated but cachable is false? Should this trigger a refresh of the entry? Minor,,, maybe this behavior is fine.

NickGeneva (approved):
Few minor comments. LGTM otherwise

CHANGELOG.md
negin513: lol!! great catch.

pyproject.toml
negin513: It's a hard runtime dep rather than typing-only: `data/utils.py` imports `obspec` at module level and the `AsyncReadableStore`/`AsyncListableStore` protocols subclass its classes, which executes at import time via `earth2studio.data`'s eager imports. It also isn't pulled in transitively — obstore's only dependency is `typing-extensions` on <3.13.

Cost-wise it's about as light as a dep gets: pure-Python protocol definitions from the obstore maintainers (developmentseed), 15 kB wheel / ~130 kB installed, no transitive deps on Python ≥3.12 (only `typing-extensions` on 3.11, which torch already requires). What it buys: the shared helpers are typed against a vendor-neutral store interface instead of concrete obstore classes, so tests inject plain fake stores rather than patching obstore internals.

If you'd still rather keep core lean, I can guard the import + protocol definitions under `if TYPE_CHECKING:` with `Any` runtime fallbacks and move obspec to the dev group — everything keeps working since usage is annotation-only. In 909479b I moved the entry next to obstore and recorded it in the changelog Dependencies section.

earth2studio/data/utils.py
negin513: Good question — with the current callers it can't happen: an entry only enters the cache when its hour was complete (`hour_start + 1h <= utcnow`), and completeness is monotone in wall-clock time, so any later call for the same prefix recomputes `cacheable=True`. The only path to a populated cache with `cacheable=False` would be the system clock stepping backwards.

That said, no reason to rely on that argument — in 909479b `cacheable=False` now bypasses the cache read as well, i.e. it means "always re-list: never read from or store in the cache". No behavior change for GOES/GLM, and there's a test covering the pre-populated-cache case.

earth2studio/data/goes.py
```diff
@@ -379,26 +410,41 @@ async def fetch_array(
...
+            return self._hour_listing_cache[prefix]
+        if self.store is None:
+            raise ValueError("Object store is not initialized")
+
+        files = [
+            f"{self._bucket}/{entry['path']}"
+            async for chunk in obs.list(self.store, prefix=prefix)
+            for entry in chunk
+        ]
+        self._hour_listing_cache[prefix] = files
```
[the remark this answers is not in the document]
negin513: Fixed in c0effbb (current behavior at HEAD): `_list_hour_files` passes `cacheable=complete` to the shared listing helper, so only complete (past) hours are memoized and the still-filling hour is re-listed on every call. 909479b adds `test_goes_list_hour_files_memoization` to pin this at the datasource level.

earth2studio/data/goes.py
```diff
@@ -299,20 +315,30 @@ async def fetch(
...
+        async_tasks = [(i, t, variable) for i, t in enumerate(time)]
+        coros = [self.fetch_wrapper(task, xr_array=xr_array) for task in async_tasks]
+
+        await gather_with_concurrency(
+            coros,
+            max_workers=self._max_workers,
+            task_timeout=120.0,
+            desc="Fetching GOES data",
+            verbose=(not self._verbose),
         )
```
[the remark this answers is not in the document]
negin513: Good catch — fixed in 909479b by removing the gather-level `task_timeout`, as suggested. `async_retry`'s per-attempt 120s timeout still bounds every attempt (worst case (retries+1) × 120s + backoff), and the sync `__call__` path keeps its `async_timeout` cap. Also added regression tests: one pinning the call site (no gather-level timeout passed) and one covering the retry/timeout composition in both directions.

PR #1043, opened 2026-08-05
Migrate Himawari AHI data source to obstore

negin513:
# Earth2Studio Pull Request

## Description
Migrates the Himawari AHI (`HimawariAHI`) data source from s3fs to obstore, building on the shared helpers from #1042, and tunes the fetch path for its many-small-tiles access pattern.

### Migration

- Whole-file tile fetches go through `obstore_fetch_to_cache` (cache keys still hashed from the historical bucket-prefixed paths, so warm caches remain valid), listings through `obstore_list_prefix` with per-minute-directory memoization (scans older than an hour; fresher scans are always re-listed). Store attributes are obspec-typed (`AsyncListableStore`).
- The per-call s3fs session rebuild (`skip_instance_cache=True` + `managed_session`) is removed; the obstore store is built once and reused across calls.
- `available()` uses a one-entry obstore listing probe instead of a full `fs.ls`.

### Performance tuning

- Default `async_workers` raised 16 -> 32: a fetch is ~88 small tile objects per channel, so throughput is round-trip-latency bound and 16 workers left waves of idle latency (measured ~23% on its own).
- Tile NetCDF decode + block averaging moved off the event loop into a worker thread (`asyncio.to_thread`) so decode overlaps concurrent downloads. HDF5/netCDF4 is not thread-safe, so the NetCDF read itself is serialized behind a module-level lock while numpy post-processing runs concurrently.

### Benchmarks

Cold cache, public NOAA S3 (`noaa-himawari9`), alternating runs; byte-identical downloads and identical output arrays verified on every run.

| Workload                                        | s3fs (main) | this PR | Change       |
|-------------------------------------------------|-------------|---------|--------------|
| 1 call, 2 channels (176 tiles, 44 MB)           | 9.5 s       | 5.2 s   | ~1.8x faster |
| 4 sequential single-timestamp calls (352 tiles) | 28.3 s      | 12.3 s  | ~2.3x faster |

### Testing

- `test/data/test_himawari_ahi.py` passes including `--slow` network tests. Tests that patched `s3fs.S3FileSystem` internals now inject plain fake obspec stores.

## Checklist

- [x] I am familiar with the [Contributing Guidelines (https://github.com/NVIDIA/earth2studio/blob/main/CONTRIBUTING.md).

## Dependencies

No new dependencies beyond those introduced in #1042 (`obspec`). Removes another s3fs usage.


earth2studio/data/himawari_ahi.py
```diff
@@ -221,6 +225,16 @@ def _compute_pixel_roi(
     return int(rows.min()), int(rows.max()) + 1, int(cols.min()), int(cols.max()) + 1
 
 
+# HDF5 (netCDF4) is not thread-safe; serialize NetCDF reads across the
+# decode worker threads while numpy post-processing runs concurrently
+_HDF5_LOCK = threading.Lock()
+
+
+def _utcnow() -> datetime:
```
NickGeneva: Used in one spot, can this get inlined?

earth2studio/data/himawari_ahi.py
```diff
@@ -221,6 +225,16 @@ def _compute_pixel_roi(
     return int(rows.min()), int(rows.max()) + 1, int(cols.min()), int(cols.max()) + 1
 
 
+# HDF5 (netCDF4) is not thread-safe; serialize NetCDF reads across the
+# decode worker threads while numpy post-processing runs concurrently
+_HDF5_LOCK = threading.Lock()
```
NickGeneva, line 230: Is it not thread safe for reads? I can understand writes.

NickGeneva (approved):
Overall LGTM, two minor questions / nits. Approving to unblock!

earth2studio/data/himawari_ahi.py
negin513: Good call — inlined in 511639b6. It was a single call site and the tests don't patch it, so the seam was speculative.

earth2studio/data/himawari_ahi.py
negin513: Unfortunately reads too — the hazard isn't the file handle, it's the HDF5 C library's process-global state (error stacks, file-handle table, metadata cache), which is mutated on every open/read. Standard pip/conda HDF5 builds aren't compiled with `--enable-threadsafe`, so concurrent opens from two threads can segfault even on disjoint files, and netCDF4-python (the engine xarray resolves to here) has no internal locking of its own.

We could avoid the explicit lock by forcing `engine="h5netcdf"` and leaning on h5py's global API lock, but that's the same serialization hidden in a third-party internal plus a decode-behavior change. The lock only covers the NetCDF open/extract — the expensive `astype` + block-average downsampling run outside it in parallel across the worker threads.

negin513:
/ok to test [373ed66](https://github.com/NVIDIA/earth2studio/pull/1043/commits/373ed66ebb5e44240b0efae5e84a7149dd22b700)

PR #1046, opened 2026-08-06
Add hyperspectral IR sounder variables (AIRS, IASI, CrIS) to NNJAObsSat

negin513:
## Description

Adds the hyperspectral IR sounders — `airs`, `iasi`, `cris` — as variables on the existing `NNJAObsSat` source, reading the NCEP aggregate BUFR products from the NNJA archive at `s3://noaa-reanalyses-pds`. One source now covers all NNJA satellite data (microwave + IR) with a single entry point, matching how UFS sat data is structured; a mixed request like `["atms", "cris"]` returns a single frame.

All three sensors are returned in **brightness temperature (K)**:
- **AIRS** (Aqua, 2002–2023, `airs/nasa/aqua` route): stored as `TMBR` — no conversion needed. Per-channel wavenumbers read from the `LOGRCW` field in the BUFR.
- **IASI** (MetOp-A/B/C, 2008–present): integer-coded scaled radiance (`SCRA`) with per-channel-group scale exponents (`CHSF`); Planck inversion via the shared `radiance_to_bt`.
- **CrIS** (NPP/NOAA-20/NOAA-21, 2018–present, `crisf4` FSR): float radiance (`SRAD`); Planck inversion. Per-band `NFQF`/`NCQF` quality flags packed into the `quality` column; cross-track position from `FORN`.

Channel selection follows the JPSS/METOP `sensor_indices` interface (validated at construction: non-empty, unique, on the instrument grid); default returns all archived channels.

### New files

| File | Purpose |
|---|---|
| `data/utils_ir.py` | IASI/CrIS spectral grids, radiance unit conversions |
| `lexicon/nnja.py` | `airs`/`iasi`/`cris` variable ids (extends existing lexicon) |
| `test/data/test_utils_ir.py` | Grid, conversion, and cross-consistency tests |

### Modified files

- `data/utils_ncep.py` — adds `decode_ir_sounder()` alongside `decode_microwave()`, using the same `ProcessPoolExecutor`/batch pattern; widens `quality` to `uint32` (AIRS `ACQF` is 24-bit, CrIS packed band flags are 28)
- `data/nnja.py` — IR product routing with per-product naming conventions, platform maps, `sensor_indices` with validation
- `data/utils.py` — `radiance_to_bt`: `log1p` → `log(1+x)` (argument ≥16 over Earth scenes; measurably faster, identical accuracy) and CODATA docstring correction

### S3 paths (verified against live bucket)

```
AIRS: airs/nasa/aqua/YYYY/MM/bufr/airs_disc_final.YYYYMMDD.tHHz.bufr
IASI: iasi/mtiasi/YYYY/MM/bufr/gdas.YYYYMMDD.tHHz.mtiasi.tm00.bufr_d
CrIS: cris/crisf4/YYYY/MM/bufr/gdas.YYYYMMDD.tHHz.crisf4.tm00.bufr_d
```

### Review history

All four correctness items from [the review](https://github.com/NVIDIA/earth2studio/pull/1046#issuecomment-5270319520) are addressed:

1. AIRS routed to `airs/nasa/aqua` (2002–2023) with the `airs_disc_final` filename convention
2. CrIS cross-track position emitted: `scan_position` is `FORN` (1–30) for CrIS, `FOVN` for all other sensors
3. CrIS quality decoded: per-band `NFQF`/`NCQF` packed as `nfqf | ncqf<<19` per channel row; IASI comment reworded to name the footprint-level QGFQ-family scalars
4. `quality` widened to `uint32`

Plus the non-blocking dedup and hardening items: Planck inversion consolidated onto the shared `radiance_to_bt`; CrIS spectral grid pinned against `jpss_cris` (and IASI channel count against `metop_iasi`) by test; `sensor_indices` validation; `0-31-001` guard-spectrum marker handled explicitly with the second-`31002` break scoped to AIRS; footprints missing their cross-track position dropped rather than failing the file; wavenumbers resolved once per footprint; documentation for `elev` (CrIS-only; `SELV` is platform altitude), `scan_angle` (always NaN for IR), per-sensor `scan_position` semantics, and realistic per-cycle row magnitudes in the memory warning.

**Note:** the ISD/IBTrACS/CFS-reforecast/OPERA obstore migrations previously stacked on this branch were split out to #1058; this PR is now NNJA IR only.

## Follow-on PRs

- **#1059 (draft, stacked)** — performance/memory: vectorized per-footprint conversion and Arrow-per-batch accumulation, per §5.1–5.2 of the review analysis
- Emit ≤0-radiance rows with null observation instead of dropping (provenance; §4.4)
- `iasi_radiance`/`cris_radiance` variables alongside BT (§2.2)
- Optional `airs_gdas` variant exposing the `airsev` route (collocated AMSU-A + TOCC; §2.3)
- Extract a shared MW/IR batch-decode harness (the two paths are deliberately parallel today)

## Checklist

- [x] S3 paths verified against live `noaa-reanalyses-pds` bucket
- [x] Output schema shared with the microwave sensors (mixed MW+IR requests concat cleanly)
- [x] Unit tests — 67 passing (mocked BUFR decode, grid cross-consistency vs `jpss_cris`/`metop_iasi`)
- [x] Docs — `NNJAObsSat` entry in `datasources_dataframe.rst` renders the updated docstring
- [x] `ruff` clean
- [x] `mypy` — only pre-existing `pandas` stubs warning (identical to `nnja.py`, `ufs.py`, etc.)


aayushg55:
@negin513 Some initial high level comments:

- I don't think we need a separate NNJAObsIRSat. It would be nice to match the UFS sat data in how the single "source" contains all the satellite data (IR + MW) and there is a single entry point. We can just have a separate decode path for MW and IR. But open to push back here
- For channel selection, we shouldn't add the csv we created (the channel ranking file). Our HealDA specific things should be largely kept orthogonal to the E2S datasource. Channel selection should happen through the existing interface /api that exists for UFS, JPSS CRIS, and others. (i.e. no `ir_channels` argument). Default should return all channels
- not sure if we need airs_wavenumbers.csv as of yet.

negin513:
Thanks @aayushg55, good points — went ahead and restructured:

- Dropped `NNJAObsIRSat` entirely. The IR sounders are now just variables on `NNJAObsSat` (`airs`, `iasi`, `cris`) and `_decode_file` routes them to the IR decode path internally. So one source for all the sat data like UFS, and a mixed request like `["atms", "cris"]` comes back as a single frame.
- Removed `ir_channels` and the ranking csv. Default is all channels now. I added a `sensor_indices` arg for narrowing since that's the interface JPSS_CRIS and the METOP sources already use, but you only get subsetting if you ask for it.
- On `airs_wavenumbers.csv`, what would be the best way to handle it? Concidering missing wavwenumber data?

negin513:
Update on the airs csv: turns out it's not needed at all. The airsev BUFR encodes the centre wavenumber per channel (LOGRCW, log10 of the wavenumber in m⁻¹), so I now read it straight from the file and deleted the csv. Checked against a real 2019 file, all 281 channels match the old table to <0.01 cm⁻¹.

Doing that also surfaced a decode bug. The AMSU-A/HSB blocks that follow the AIRS channels aren't always separated by another replication marker, so a few microwave channels were leaking through as AIRS rows. The decoder now uses the declared replication count as the block boundary.

aayushg55:
@negin513 Thanks for the restructure.

I had my agent review the NNJA IR based on our current work/findings in converting the same bufr to parquet. It produced a rather lengthy document which I think has some highly relevant information which would be hard to distill otherwise, so attaching in its full here as documentation/reference. The summary and the suggested changes are both at the top, everything after that is the additional evidence.
[e2studio_pr1046_nnja_ir.md](https://github.com/user-attachments/files/30991403/e2studio_pr1046_nnja_ir.md)

The four correctness items, repeated here:
1. **AIRS route.** `airsev` covers 2007-02-21 → 2020-12-07, but `airs/nasa/aqua` in the same bucket
   covers a longer 2002-08-31 → 2023-01-18. The two are structurally identical and mostly agree on the infrared
   values. Worth defaulting to the NASA route, or exposing both somehow. (We use the NASA route)
2. **CrIS `FORN` is decoded into `scalars` and then never emitted.** CrIS is the one sensor where
   `FOVN` is not the cross-track position — `FOVN` is 1–9 within the field of regard, `FORN` is 1–30
   across track, and all 270 pairs occur. So CrIS currently comes out of this source with no
   cross-track position at all, which is what edge-of-scan screening and scan-angle bias correction
   key on. `jpss_cris.py` already carries the two as separate columns.
3. **CrIS quality is hard-coded to "none".** `crisf4` carries five quality fields. `NFQF` is the one
   that flags the rows damaged by the NOAA-21 neon event, and those radiances are finite, so a
   consumer can't spot them any other way. The long layout makes this tractable — resolve the band
   for each channel and pack that band's flags.
4. **`quality` is `uint16` but AIRS `ACQF` is a 24-bit flag table.** A value above 65535 doesn't
   truncate, it raises, and that happens outside the per-message `try`, so one such value fails the
   whole request. `uint32` covers it.

The rest is mainly nice to have code deduplication and some comments worth correcting.

There are also some points on performance/memory, but this is not really specific to this implementation but rather generally applicable for all these raw high-resolution MW/IR sounders data sources, and particularly for IR which has hundreds of channels. These would probably be best addressed in future PRs. TLDR, as is, 48hr of satellite data might consume a lot of memory.

aayushg55 (approved):
Thanks @negin513 for following through with all of the changes!

negin513:
/ok to test [f4f0679](https://github.com/NVIDIA/earth2studio/pull/1046/commits/f4f0679de5e3abe48842893de7b4dcaabdc69500)

negin513:
Thanks @aayushg55 . All issues are addressed and started a  perf PR on #1059 for vectorized conversion.

negin513:
Will bring this in once all tests pass.

PR #1054, opened 2026-08-11
Documentation Upgrade / Refactor

NickGeneva:
# Earth2Studio Pull Request

## Description

https://nickgeneva.github.io/earth2studio/

## Checklist


NickGeneva:
@greptile-apps

.github/workflows/e2s-docs-full.yml
```diff
@@ -43,17 +44,19 @@ env:
 
 jobs:
   build-docs:
-    name: Build full docs
+    name: Build and publish full docs
     if: github.repository == 'NVIDIA/earth2studio' && github.ref_type == 'branch' && (github.ref_name == github.event.repository.default_branch || endsWith(github.ref_name, '-rc'))
+    permissions:
+      contents: write
+      packages: read
```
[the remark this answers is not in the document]
NickGeneva: This is by design because of mike versioning and how it works.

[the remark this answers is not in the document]
NickGeneva: Seperate stages would mean manual upload process

negin513 (approved):
Great job!! This looks great. I have so many positives to say; especially great work on user guide and install guide. 


Here are some top level comments, I think we should address: 
### Aesthetics 
- Let’s use **NVIDIA Sans** consistently throughout and increase the size of the NVIDIA logo on top left. 
- I don't know if we need to stick with a certain color theme for NVIDIA docs or not. If so, I prefer a clean no color gradient look, such as zarr docs (https://zarr.dev/) or [this](https://negin513.github.io/earth2studio-landing/recipes/)) for docs. (This is very personal preference though). 
- Let’s refine the landing-page styling to move away from the current simpler, particularly some of the square/card elements.  I want to use font awesome elements to make it look fancier. ;-) -- Absolute nit that I can take care of. 

### Content
- Let’s keep the blog hidden for now until we get a +1 on publishing it? @pzharrington what do you think? 
- @negin513 will update the recipes. I like to have one tab for recipes and examples. Both ercipes and examples are very empty on the left side. 

Overall, this is in very good shape. Fantastic work!! :-)

p.s. this comment is very human written.

docs/assets/stylesheets/nvidia-material.css
```diff
@@ -0,0 +1,2404 @@
...
+  width: min(100%, var(--e2s-content-width), 52rem);
+  margin: 1.8rem auto 0;
+  transform: translate3d(0, var(--e2s-hero-panel-y, 0), 0);
+  will-change: transform;
+}
+
+.e2s-eyebrow,
+.e2s-section-kicker {
+  color: var(--nv-green-2) !important;
+  font-family: "Cascadia Code", "SFMono-Regular", Consolas, monospace;
```
negin513: Suggest using NVIDIA Sans for the section kickers/eyebrows ("Agent ready", etc.) to match the rest of the page typography. 

```suggestion
  font-family: "NVIDIA Sans", sans-serif, Consolas, monospace;
```

docs/griffe_extensions.py
```diff
@@ -0,0 +1,131 @@
...
+            output.append(line)
+            index += 1
+            continue
+
+        indent = line[: len(line) - len(line.lstrip())]
+        output.append(f"{indent}```pycon")
+        while index < len(lines):
+            current = lines[index]
+            if _is_doctest_line(current):
+                output.append(current)
```
negin513, line 89: `_convert_doctest_blocks` closes the ` ```pycon ` fence at the first non-prompt line, so doctest *expected output* lines are pushed outside the fence and render as stray prose between fragmented code blocks.

Reproduced with `tile_coords` from `earth2studio/utils/coords.py`:

````
```pycon
>>> x_tiled.shape
```
torch.Size([2, 3, 3, 4])
```pycon
>>> list(out.keys())
```
['a', 'b']
````

Expected-output lines that follow a prompt line should stay inside the fence; the fence should only close on a blank line / end of the doctest block.

File: docs/griffe_extensions.py [hunk 4]
```diff
@@ -0,0 +1,131 @@
...
+    return f"`{escaped}`"
+
+
+def _convert_inline_roles(docstring: str) -> str:
+    """Render supported Sphinx inline roles for Markdown API docs."""
+
+    def replace(match: re.Match[str]) -> str:
+        raw_target = match.group("target").strip()
+        label, target = _legacy_role_label(raw_target)
+        if raw_target.startswith("~earth2studio."):
```
negin513, line 59: This only emits an autorefs link when the role target literally starts with `~earth2studio.` — everything else degrades to unlinked inline code. Grepping the source tree, none of the ~154 `:class:`/`:func:`/`:meth:` role usages in docstrings use that prefix (they're bare same-module refs like `` :class:`Aurora1p5` ``), so effectively all existing docstring cross-references lose their hyperlinks in the new build. The old Sphinx build resolved bare refs via current-module lookup.

docs/generate_api.py
```diff
@@ -0,0 +1,964 @@
...
+        start = getattr(node, "lineno", None)
+        end = getattr(node, "end_lineno", None)
+    elif source is not None:
+        docstring = module_docstring(source)
+        start = end = None
+    else:
+        docstring = ""
+        start = end = None
+    badges = tuple(dict.fromkeys(BADGE_RE.findall(docstring)))
+    summary = clean_rst_roles(first_paragraph(docstring))
```
negin513, line 724: `clean_rst_roles` bakes markdown link syntax `` [`X`][x] `` into the front-matter `summary`, but downstream consumers treat summary as plain text and HTML-escape it (`mkdocs_hooks.py` `escape(summary)`, `e2s-catalog.js` `escapeHtml(item.summary)`). So e.g. the `Aurora1p5Ensemble` catalog card displays the literal text ``[`Aurora1p5`][Aurora1p5]`` instead of prose. Summaries probably want a plain-text role conversion (label only, no link syntax).

.github/workflows/tmp-fork-docs-api.yml
```diff
@@ -0,0 +1,153 @@
...
+          for plugin in ("mike",):
+              config = remove_plugin(config, plugin)
+
+          (root / "mkdocs.fork-api.yml").write_text(config, encoding="utf-8")
+          PY
+
+      - name: Build docs preview
+        run: uv run zensical build --clean --config-file mkdocs.fork-api.yml
+
+      - name: Fix API cross-references
```
negin513: This step has a `name:` but no `run:`/`uses:`, which makes the workflow file invalid. Looks like a placeholder that should be filled in or deleted??

.github/workflows/e2s-docs-full.yml
```diff
@@ -90,116 +90,51 @@ jobs:
...
           echo "- Branch: \`${GITHUB_REF_NAME}\`" >> "$GITHUB_STEP_SUMMARY"
           echo "- DOC_VERSION: \`${doc_version}\`" >> "$GITHUB_STEP_SUMMARY"
-          echo "- Publish path: \`${publish_path}\`" >> "$GITHUB_STEP_SUMMARY"
 
       - name: Checkout repository
         uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5
         with:
           fetch-depth: 0
-          persist-credentials: false
+          persist-credentials: true
```
negin513, line 102: Privilege-isolation regression vs `main`: this job now holds `contents: write` and all docs secrets, checks out with `persist-credentials: true`, and then executes example code (plus the git-sourced gallery tool) during the build. On `main`, the build job ran examples with a read-only token and `persist-credentials: false`, and publishing was a separate job gated by the `docs-publish` environment. With this change, any compromised dependency/checkpoint executed during an example run can read the write-scoped token from `.git/config` and push to `gh-pages` (same pattern in `e2s-docs-light.yml`). Consider keeping the build read-only and doing the mike deploy in a separate gated job, or at least `persist-credentials: false` + an explicit token only for the deploy step.

docs/blog/index.md
```diff
@@ -0,0 +1,15 @@
+# Blog
+
+Earth-2 product and engineering blog posts and updates.
+
+## Latest Updates
+
+### [MkDocs Material Migration](posts/mkdocs-upgrade/)
```
negin513: These hardcoded `posts/mkdocs-upgrade/` links 404 on the deployed site. `mike deploy` (what CI publishes) rebuilds with plain mkdocs-material, whose blog plugin publishes the post at `blog/2026/08/06/mkdocs-material-migration/` — `blog/posts/mkdocs-upgrade/` is never generated. (The links do work under the local `zensical` build, which doesn't relocate posts — so this is easy to miss in local preview.) Same issue in `blog/archive.md:5`, `blog/categories.md:5` (doubly broken — relative resolution gives `/blog/categories/categories/documentation/`), and `blog/categories/documentation.md:6`. Verified by building the site with mkdocs; the build also logs 'unrecognized relative link' warnings for these but exits 0 with `strict: false`.

docs/assets/javascripts/e2s-catalog.js
```diff
@@ -0,0 +1,231 @@
...
+      button.addEventListener("click", () => {
+        state.kind = button.dataset.catalogKind;
+        state.filters = {};
+        state.page = 1;
+        render();
+      });
+    });
+    const searchInput = root.querySelector("[data-catalog-search]");
+    if (searchInput) searchInput.addEventListener("input", (event) => {
+      state.query = event.target.value.trim();
```
negin513: The `.trim()` here plus the full re-render with caret forced to the end makes multi-word search impossible: typing `sea ice`, the trailing space is stripped, `render()` rebuilds the input with the trimmed value, and the caret jumps to the end — so the next keystroke produces `seai`. It also breaks mid-string editing (caret jumps on every keystroke). Suggest keeping the raw value in the input (trim only when *filtering*) and not re-rendering the input element itself on keystrokes.

negin513 (commented):
Automated review pass (multi-agent, each finding adversarially verified — several reproduced by building the site). Inline comments below; happy to be corrected on any of them.

examples/README.md
```diff
@@ -1,34 +1,34 @@
-Examples
-========
+# Examples
```
negin513, line 1: This file was rewritten in Markdown (`#` headings, `??? info` collapsibles, fenced code blocks) but keeps the `.rst` extension, so GitHub renders it as garbled reStructuredText. And since the new gallery builder derives content from the example scripts rather than README files, this content doesn't appear in the docs site either. Should probably be renamed to `README.md` (with the `??? info` collapsible replaced, since GitHub won't render that) or reverted to valid rst.

docs/userguide/advanced/lexicon.md
```diff
@@ -76,11 +74,10 @@ data store.
...
 The following snippet is part of the lexicon for the GFS dataset.
 The class has a `metaclass=LexiconType`, which is present in
-{py:mod}`earth2studio.lexicon.base` used for type checking.
+`earth2studio.lexicon.base` used for type checking.
 
-```{literalinclude} ../../../earth2studio/lexicon/gfs.py
-    :lines: 24-60
-    :language: python
+```python
+--8<-- "earth2studio/lexicon/gfs.py:288:315"
```
negin513: The Sphinx `start-after`/`end-before` marker includes were converted to hardcoded absolute line ranges, which has caused several problems across the userguide:

- Here: `lexicon/gfs.py:288:315` shows the GFS *modifier function*, but the surrounding text describes the `GFSLexicon` class (`metaclass=LexiconType`, `VOCAB`) — the old page included lines 24–60. The identical modifier snippet then repeats ~20 lines below.
- Ranges start/end exactly on the marker lines, so `# sphinx - ... start/end` comments leak into rendered snippets (here, `components/io.md:50`, `components/datasources.md:99`).
- `components/data_assimilation.md:23` (`da/base.py:28:171`) opens on a dangling `except ImportError:` fragment and cuts off mid-docstring, dropping the required `to()` method; `components/diagnostic.md:24` also drops `to()` vs the old `:lines: 25-`.

Beyond the current inaccuracies, absolute line numbers will silently drift as the source files change. pymdownx snippets supports section markers (`--8<-- "file.py:section-name"` with `; --8<-- [start:name]` comments), which would preserve the old marker-based behavior.

docs/userguide/about/install_options.yml
```diff
@@ -0,0 +1,404 @@
...
+          - Depends on flash-attn, which can take a long time to build.
+      - id: aifs2ens
+        label: AIFS 2.0 Ensemble
+        extra: aifs2ens
+        summary: ECMWF AIFS 2.0 ensemble model dependencies.
+        api_refs:
+          - earth2studio.models.px.AIFS2ENS
+        warnings:
+          - Depends on flash-attn, which can take a long time to build.
+      - id: atlas
```
negin513, line 105: The generated install selector drops the mandatory pip preinstall steps the old `install.md` documented — only `fcn3` and `sfno` kept `preinstall` blocks. E.g. `atlas` previously required `pip install --no-build-isolation "torch-harmonics @ git+..."` first, and `cbottle` needed hatchling + earth2grid + cbottle git installs; a pip user following the new selector gets only `pip install "earth2studio[atlas]"` and hits a build failure (or import failure at model load). Affects atlas/cbottle/dlesym and likely stormscope/cosmo/graphcast/orbit.

.github/workflows/e2s-docs-full.yml
negin513: ????

docs/griffe_extensions.py
```diff
@@ -0,0 +1,131 @@
...
+            if _is_doctest_line(current):
+                output.append(current)
+                index += 1
+                continue
+            if not current.strip() and index + 1 < len(lines):
+                if _is_doctest_line(lines[index + 1]):
+                    output.append(current)
+                    index += 1
+                    continue
+            break
```
negin513, line 97: Suggested fix for the fragmented-fence issue flagged above: treat any non-blank line as part of the doctest block (prompt or expected output), and close the fence only at a blank line that is not followed by another prompt. Tested against the `tile_coords` docstring — output/prompt lines now render as one clean `pycon` block, blank lines between consecutive doctest groups keep the fence open, and trailing prose still lands outside the fence.

```suggestion
        while index < len(lines):
            current = lines[index]
            if current.strip():
                output.append(current)
                index += 1
                continue
            if index + 1 < len(lines) and _is_doctest_line(lines[index + 1]):
                output.append(current)
                index += 1
                continue
            break
```

docs/userguide/about/install_options.yml
negin513: ???

docs/griffe_extensions.py
```diff
@@ -0,0 +1,131 @@
...
+    """Render supported Sphinx inline roles for Markdown API docs."""
+
+    def replace(match: re.Match[str]) -> str:
+        raw_target = match.group("target").strip()
+        label, target = _legacy_role_label(raw_target)
+        if raw_target.startswith("~earth2studio."):
+            return f"[`{label}`][{target}]"
+        return _inline_code(label)
+
+    return RST_INLINE_ROLE_RE.sub(replace, docstring)
```
negin513, line 63: Suggested fix for the lost cross-references flagged above: since `on_object` has the griffe `obj` in scope, bare refs can be resolved the way Sphinx did — via current-scope lookup — using `obj.resolve()`. Tested against the actual package with griffe: 53 of the 67 role usages in the loaded modules now resolve to fully-qualified `earth2studio.*` autorefs targets (including dotted refs like `` :meth:`GHCNDaily.get_stations_bbox` ``); the remaining 14 are third-party/builtin refs (`~numpy.ndarray`, `pandas.DataFrame`, `ValueError`) that correctly keep the current inline-code fallback. Needs the companion one-line change at the call site below (line 130) to pass `obj` through.

One caveat worth checking: if a resolved target points at an object that has no rendered API page, autorefs will warn and leave the raw `[label][target]` text — worth a full docs build to confirm no such refs exist before applying.

```suggestion
def _resolve_ref_target(obj: griffe.Object, target: str) -> str | None:
    """Resolve a cross-reference name in the object's scope to a full path."""
    if target.startswith("earth2studio."):
        return target
    first, _, rest = target.partition(".")
    try:
        resolved = obj.resolve(first)
    except Exception:
        return None
    if not resolved.startswith("earth2studio."):
        return None
    return f"{resolved}.{rest}" if rest else resolved


def _convert_inline_roles(docstring: str, obj: griffe.Object | None = None) -> str:
    """Render supported Sphinx inline roles for Markdown API docs."""

    def replace(match: re.Match[str]) -> str:
        raw_target = match.group("target").strip()
        label, target = _legacy_role_label(raw_target)
        if obj is not None:
            resolved = _resolve_ref_target(obj, target)
            if resolved is not None:
                return f"[`{label}`][{resolved}]"
        elif target.startswith("earth2studio."):
            return f"[`{label}`][{target}]"
        return _inline_code(label)

    return RST_INLINE_ROLE_RE.sub(replace, docstring)
```

docs/griffe_extensions.py
```diff
@@ -0,0 +1,131 @@
...
+
+class StripBadgesSection(griffe.Extension):
+    """Strip the package badge section from rendered docstrings."""
+
+    def on_object(self, *, obj: griffe.Object, **kwargs: object) -> None:
+        """Update collected docstrings in place before mkdocstrings renders them."""
+        if not obj.docstring:
+            return
+        docstring = _strip_badges_section(obj.docstring.value)
+        docstring = _convert_inline_roles(docstring)
```
negin513, line 130: Companion to the suggestion at lines 53–63 — pass the object through so bare refs resolve in its scope:

```suggestion
        docstring = _convert_inline_roles(docstring, obj)
```

docs/assets/stylesheets/nvidia-material.css
```diff
@@ -0,0 +1,2404 @@
...
+  font-size: 0.68em;
+  font-weight: 900;
+  opacity: 0.86;
+  transform: translateY(-0.06em);
+}
+
+.md-header__button.md-logo img,
+.md-header__button.md-logo svg {
+  height: 1.5rem;
+}
```
negin513: Suggest making the header nvidia logo bigger (1.5rem → 2.25rem, 1.5×) without changing the header height.

```suggestion
.md-header__button.md-logo {
  margin: 0;
  padding: 0.075rem 0.6rem;
}

.md-header__button.md-logo img,
.md-header__button.md-logo svg {
  height: 2.25rem;
}
```

docs/assets/stylesheets/nvidia-material.css
```diff
@@ -0,0 +1,2404 @@
...
+  padding: 0 1rem;
+  border: 1px solid rgba(118, 185, 0, 0.34);
+  border-radius: 0.32rem;
+  background: rgba(5, 8, 6, 0.4);
+  box-shadow:
+    inset 0 0 0 1px rgba(118, 185, 0, 0.08),
+    0 0.65rem 2rem rgba(0, 0, 0, 0.34),
+    0 0 1.8rem rgba(118, 185, 0, 0.1);
+  color: var(--nv-green-2) !important;
+  font-family: "Cascadia Code", "SFMono-Regular", Consolas, monospace;
```
negin513: Same NVIDIA Sans treatment for the hero "Get Started" button (this `.e2s-hero__start` rule is its winning font-family — the label span inherits from the anchor):

```suggestion
  font-family: "NVIDIA Sans", sans-serif;
```

mkdocs.yml
```diff
@@ -0,0 +1,435 @@
...
+exclude_docs: |
+  _build/**
+  overrides/**
+site_dir: site
+strict: false
+copyright: Copyright &copy; 2024-2026 NVIDIA Corporation & Affiliates
+theme:
+  name: material
+  custom_dir: docs/overrides
+  logo: _static/NVIDIA-Logo-V-ForScreen-ForDarkBG.png
```
negin513: Suggest using the horizontal NVIDIA logo lockup here, matching the current production docs site. Note this needs the two SVGs added to `docs/_static/` — they aren't in the repo (the production site serves them from the old Sphinx build's static files): `nvidia-logo-horiz-rgb-wht-for-screen.svg` and `nvidia-logo-horiz-rgb-blk-for-screen.svg`, both downloadable from https://nvidia.github.io/earth2studio/_static/. Previewed locally with the enlarged header sizing suggested in `nvidia-material.css` — the horizontal lockup at 2.25rem reads much better in the header than the vertical mark.

```suggestion
  logo: _static/nvidia-logo-horiz-rgb-wht-for-screen.svg
```

docs/assets/stylesheets/nvidia-material.css
```diff
@@ -0,0 +1,2404 @@
...
+  border-radius: 0.35rem;
+  background: var(--e2s-install-panel);
+}
+
+.e2s-install-step legend,
+.e2s-install-select span {
+  display: block;
+  margin-bottom: 0.55rem;
+  color: var(--e2s-install-link);
+  font-family: "Cascadia Code", "SFMono-Regular", Consolas, monospace;
```
negin513: NVIDIA Sans for the install-selector step labels too ("1. Choose package manager", "2. Choose source", "3. Choose component", and the "Choose sub-component" span — this selector matches exactly those four labels). The command/code output pane and option-button labels match neither selector part, so they keep their current treatment.

```suggestion
  font-family: "NVIDIA Sans", sans-serif;
```

negin513:
Consolidation idea for the label typography: the `"Cascadia Code", "SFMono-Regular", Consolas, monospace` stack is repeated verbatim in 13 rules across `docs/assets/stylesheets/nvidia-material.css` (kickers, hero button, install-step labels, catalog filters, gallery labels, …). Several of my inline suggestions are already swapping individual occurrences to NVIDIA Sans one rule at a time. Rather than piecemeal edits, consider defining the label font once in the `:root` block:

```css
:root {
  --md-text-font: "NVIDIA Sans";
  --e2s-label-font: "NVIDIA Sans", sans-serif;
  ...
}
```

and having all 13 rules use `font-family: var(--e2s-label-font);`. That makes the mono-vs-sans decision a one-line change site-wide, keeps the label treatment consistent, and avoids the drift where some labels get updated and others don't.

(Also worth noting: Cascadia Code has no `@font-face`/webfont import in the stylesheet, so visitors without it installed have been seeing SF Mono/Consolas fallbacks anyway.)

docs/assets/stylesheets/nvidia-material.css
NickGeneva: Fixed!

mkdocs.yml
NickGeneva: Fixed!

docs/assets/stylesheets/nvidia-material.css
NickGeneva: I think I got this fixed.

docs/assets/stylesheets/nvidia-material.css
NickGeneva: Just the code blocks should be the different font now

docs/griffe_extensions.py
NickGeneva: Yeah, trying to do full preplacement of these class, func, meth references causes like 80% to just not work (logic is potentially complicated when the full import path isnt provided). So I constrained it to just the tilde ones that have the full reference. So rather go back and update later.

.github/workflows/tmp-fork-docs-api.yml
NickGeneva: Place holder for fork deployment will delete workflow before merge. 👍

docs/griffe_extensions.py
NickGeneva: Right now skipping this, tried this originally but reverted due to i having problems with various function / method references.

examples/README.md
NickGeneva: Good spot, fixed.

docs/userguide/advanced/lexicon.md
NickGeneva: Remove the old sphinx comments, the fragile line numbers are just part of how Zensical/Mkdocs. Can convert them to code blocks down the line probably... since the interfaces arent changing.

docs/assets/javascripts/e2s-catalog.js
NickGeneva: Fixed, I think (lol this is all agent code)

docs/griffe_extensions.py
NickGeneva: skipping.

docs/assets/stylesheets/nvidia-material.css
NickGeneva: Tried to make this larger, kinda worked. Probably just need to edit the svg. Better than what it was, maybe

docs/griffe_extensions.py
NickGeneva: The doc string should probably get updated to a more markdown friendly form. Since this is only a single instance not a huge concern.

docs/griffe_extensions.py
NickGeneva: Similar comment as above, prefer to update the docstring rather than parser I think.

.github/workflows/e2s-docs-full.yml
NickGeneva: Regarding this, the difference is that right now in the PR im using mike to do the upload to the gh-pages branch... which is different from before because mike builds and pushes in one step. Could have a manual upload step but wanted to try the upload with mike. to permissions are present in the build env. Can change if it doesnt really work well.

examples/02_medium_range/_gallery.toml
pzharrington, line 1: What's up with this file?

NickGeneva: setting of like env variables in sub directories for these example... since the example sets the env var anyway it should be removed. was having a lot of trouble with atlas so kinda left over from debugging

pyproject.toml
```diff
@@ -369,6 +369,7 @@ nvidia-physicsnemo = [
 ]
 
 climate-learn = { git = "https://github.com/NickGeneva/ORBIT-2", rev = "5b2d80a8ba4dc95029211ef2b8530d3663f65d39" }
+earth2studio-gallery = { git = "https://github.com/NickGeneva/earth2studio-gallery.git", rev = "da2fbfc4258e7ac88a1c0f5ea0a6153ea8150f9c" }
```
pzharrington: Just a thought...a bit weird to have this in a separate repo, maybe we could fold the tooling into the `docs` or similar folder somewhere here eventually?

examples/02_medium_range/_gallery.toml
NickGeneva: removed.

pzharrington (approved):
Send it!

pyproject.toml
NickGeneva: Basically replicating sphinx gallery from ground up, so quite a bit of code. So keeping in this other repo so theres some dedicated tests and user guide for the agents.

pzharrington: Yeah I figured the volume of it would be the main thing. Fine for now, but can you make sure we have write access to it?

PR #1056, opened 2026-08-11
Fix CorrDiffCosmoEra5 package cache collisions

gertln:
## Description

`CorrDiffCosmoEra5.load_default_package()` used `same_names=True`, causing
same-named files under `rea2/` and `rea6/` to share one cache path. Consequently,
loading a resolution could read metadata cached for the other resolution and fail
with errors such as `KeyError: 'rea6_diffusion'`.

Changes:

- Use `BasenameCacheMapper(directory_levels=1)` to include the resolution in cache
  names while preserving filename extensions.
- Pin the default Hugging Face package to an immutable revision.
- Add a hermetic end-to-end regression test that caches and reads same-named REA2
  and REA6 files using the model’s actual cache configuration.

### DIFF
--- a/earth2studio/models/dx/corrdiff_cosmo_era5.py
+++ b/earth2studio/models/dx/corrdiff_cosmo_era5.py
@@ -53,6 +53,7 @@ from typing import Literal
 import numpy as np
 import torch
 import xarray as xr
+from fsspec.implementations.cache_mapper import BasenameCacheMapper
 from loguru import logger
 
 from earth2studio.lexicon import CosmoLexicon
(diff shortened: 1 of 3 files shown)


earth2studio/models/dx/corrdiff_cosmo_era5.py
```diff
@@ -1609,7 +1612,10 @@ def load_default_package(cls) -> Package:
             DEFAULT_PACKAGE_URI,
             cache_options={
                 "cache_storage": Package.default_cache("corrdiff_cosmo_era5"),
-                "same_names": True,
+                # Include the resolution directory to distinguish files with
+                # matching basenames, such as rea2/metadata.json and
+                # rea6/metadata.json.
+                "cache_mapper": BasenameCacheMapper(directory_levels=1),
```
NickGeneva, line 1618: For ref in future: 
https://github.com/fsspec/filesystem_spec/blob/669de5e157cd3ce3f9bf71c7bceb7bb31c8c7a55/fsspec/implementations/cache_mapper.py#L32

NickGeneva (approved):
LGTM, amazing work

PR #1060, opened 2026-08-12
Add SamudrACE coupled prognostic model 

elynnwu:
# Earth2Studio Pull Request

## Description

Adds `SamudrACE`, a coupled atmosphere-ocean prognostic model (ACE2 atmosphere + Samudra ocean), with the data sources, lexicon, and packaging needed to run it.

`earth2studio/models/px/samudrace.py` is a thin adapter around fme's `CoupledStepper`: one `predict` call advances one coupled (5 day) step, and all coupling logic — SST prescription, flux exchange and averaging, ocean-fraction prediction, masking — stays owned by fme. The wrapper does coordinate/tensor conversion, assembles the forcing window, and buffers inner-step predictions. Only public fme APIs are used; variable lists, timesteps, and forcing requirements come from the checkpoint.

`create_iterator` is the interface: it yields the initial condition then 6 hour atmosphere steps, with ocean fields advancing once per coupled cycle and held constant in between (ocean diagnostics are NaN until the first boundary). The model is iterator-only — `__call__` raises `NotImplementedError`, because the smallest unit the stepper can advance is a full coupled cycle.

On GPU the wrapper enables `torch.backends.cudnn.benchmark` for the duration of each coupled step and restores it on exit, matching fme's inference entrypoints. With it set, a 120 day forecast through this wrapper is bit-identical to the same forecast through fme.

`earth2studio/data/samudrace.py` adds two data sources. `SamudrACEData` serves the published CM4 initial conditions from [`allenai/SamudrACE-CM4-piControl`](https://huggingface.co/allenai/SamudrACE-CM4-piControl) and raises on any other timestamp. `SamudrACEForcingData` serves per-scenario forcing: one year of 6-hourly data on the no-leap CM4 calendar, matched by month/day/hour with the year ignored, static fields broadcast. February 29 and off-grid times raise.

Tests build a real tiny `CoupledStepper` on CPU and cover the `PrognosticModel` seam (handshakes, iterator cadence, batching, device movement), plus a parity test that the iterator trajectory matches a direct `CoupledStepper.predict` trajectory. A `package`-marked test drives the real checkpoint through one coupled cycle.

Changes to existing behaviour worth maintainer attention:

- `to_time_array` no longer casts unconditionally to `datetime64[ns]`. New `normalize_time_precision` (`utils/time.py`) casts to nanoseconds only when all times are in range (1678–2262), else keeps the array's precision. Real-world dates are unchanged; CM4 model years previously wrapped silently (year 311 became 2064). `fetch_data` and `datasource_to_file` offset times through a matching `_offset_times` helper so out-of-range times are not re-wrapped. New tests in `test/utils/test_time.py` and `test/data/test_data_utils.py`.
- The `fme` git source in `[tool.uv.sources]` is scoped to the `ace2` extra, so `ace2` resolves unchanged while `samudrace` takes `fme==2026.4.0` from PyPI. The two are declared conflicting.
- Like `ace2`, `samudrace` is excluded from `all` and conflicts with `atlas`, `fcn3`, `perturbation`, and `sfno`: `fme` pins `torch-harmonics==0.8.0`, which cannot co-resolve with the git-pinned 0.8.1 those extras use.
- `E2STUDIO_VOCAB` gains ten entries: `ewss`, `nsss`, `sithick`, `siconc`, `zos`, `mttwp`, `thetao`, `so`, `uo`, `vo`.

A `test-px-models-samudrace` tox environment mirroring `test-px-models-ace2` is still needed before CI exercises these tests with `fme` installed; without it they skip.

## Checklist

## Dependencies

New optional extra `samudrace`: `fme==2026.4.0`, plus `pandas>=2.3.2` and `scipy>=1.15.2`. Conflicts with `ace2`, `atlas`, `fcn3`, `perturbation`, and `sfno`; not part of `all`.


elynnwu:
### SamudrACE sanity check
A minimal inference run using the published checkpoint, pulled directly from [Hugging Face](https://huggingface.co/allenai/SamudrACE-CM4-piControl).

samudrace_sanity_contours

```python
"""Run SamudrACE from the published checkpoint and plot forecast contours.

Sanity check for the SamudrACE coupled atmosphere-ocean emulator: runs a
120 day deterministic forecast from a CM4 preindustrial-control initial
condition and plots the atmosphere and ocean state at four lead times, the SST
change from the initial condition, and the global evolution of both components.

Everything -- checkpoint, initial condition, and forcing -- is downloaded from
the ``allenai/SamudrACE-CM4-piControl`` HuggingFace repository on first run
(roughly 1 GB, cached afterwards).

This review-only script is intentionally not part of the pull request.
"""

from pathlib import Path

import matplotlib
import numpy as np
import torch
import xarray as xr

matplotlib.use("Agg")
import matplotlib.pyplot as plt  # noqa: E402
from matplotlib.gridspec import GridSpec  # noqa: E402

import earth2studio.run as run  # noqa: E402
from earth2studio.data.samudrace import SamudrACEData  # noqa: E402
from earth2studio.io import ZarrBackend  # noqa: E402
from earth2studio.models.px import SamudrACE  # noqa: E402

# The SamudrACE checkpoint was trained on a GFDL CM4 preindustrial-control run,
# so the initial condition is a CM4 model-year timestamp, not a real date.
# Model years overflow datetime64[ns], hence second precision throughout.
INITIAL_TIME = np.datetime64("0311-01-01T00:00:00", "s")
FORCING_SCENARIO = "0311"

# The atmosphere steps every 6 hours and the ocean every 5 days, so a coupled
# cycle is 20 atmosphere steps. 24 cycles is 120 days.
STEPS_PER_CYCLE = 20
N_CYCLES = 24
NSTEPS = N_CYCLES * STEPS_PER_CYCLE

# Fields to run and plot: one atmosphere field and one ocean field.
RUN_VARIABLES = ("t2m", "sst")

# Contour rows. The SST anomaly row is what shows the ocean actually
# evolving: the raw SST field barely changes by eye over 120 days.
CONTOUR_ROWS = (
    ("t2m", "2 m temperature (atmosphere)", "K", "RdYlBu_r", (0, 30, 60, 120), False),
    ("sst", "Sea surface temperature (ocean)", "K", "viridis", (0, 30, 60, 120), False),
    (
        "sst",
        "SST change from initial condition",
        "K",
        "RdBu_r",
        (30, 60, 90, 120),
        True,
    ),
)
N_COLUMNS = 4

OUTPUT_PATH = Path("samudrace_sanity_contours.png")
STORE_PATH = Path("samudrace_sanity.zarr")


def run_forecast(device: torch.device) -> xr.Dataset:
    """Run the SamudrACE forecast and return its output as a dataset.

    Parameters
    ----------
    device : torch.device
        Device to run inference on.

    Returns
    -------
    xr.Dataset
        Forecast with dims (time, lead_time, lat, lon), one variable per
        plotted field.
    """
    package = SamudrACE.load_default_package()
    model = SamudrACE.load_model(package, scenario=FORCING_SCENARIO)
    data = SamudrACEData()

    io = ZarrBackend(
        file_name=str(STORE_PATH),
        chunks={"time": 1, "lead_time": 1},
        backend_kwargs={"overwrite": True},
    )
    run.deterministic(
        [INITIAL_TIME],
        NSTEPS,
        model,
        data,
        io,
        output_coords=OrderedDict({"variable": np.array(RUN_VARIABLES, dtype=object)}),
        device=device,
    )
    return xr.open_zarr(STORE_PATH, consolidated=False).isel(time=0).load()


def global_mean(field: xr.DataArray) -> xr.DataArray:
    """Area-weighted global mean, ignoring masked (NaN) points.

    Parameters
    ----------
    field : xr.DataArray
        Field with lat and lon dimensions.

    Returns
    -------
    xr.DataArray
        Mean over lat and lon.
    """
    weights = np.cos(np.deg2rad(field["lat"]))
    return field.weighted(weights.fillna(0.0)).mean(dim=("lat", "lon"))


def plot_forecast(forecast: xr.Dataset) -> None:
    """Plot forecast contours at several lead times and the global series.

    Parameters
    ----------
    forecast : xr.Dataset
        Forecast returned by :func:`run_forecast`.
    """
    lead_days = forecast["lead_time"] / np.timedelta64(1, "D")

    def at_lead(name: str, day: float) -> xr.DataArray:
        return forecast[name].isel(lead_time=int(np.abs(lead_days - day).argmin()))

    fig = plt.figure(figsize=(16, 12), constrained_layout=True)
    grid = GridSpec(4, 3 * N_COLUMNS, figure=fig, height_ratios=[1, 1, 1, 0.8])

    for row, (name, title, units, cmap, days, anomaly) in enumerate(CONTOUR_ROWS):
        fields = [at_lead(name, day) for day in days]
        if anomaly:
            fields = [field - at_lead(name, 0) for field in fields]
            bound = max(float(np.nanpercentile(np.abs(f), 99.0)) for f in fields)
            levels = np.linspace(-bound, bound, 21)
        else:
            lower = min(float(np.nanpercentile(f, 1.0)) for f in fields)
            upper = max(float(np.nanpercentile(f, 99.0)) for f in fields)
            levels = np.linspace(lower, upper, 21)

        for column, (day, field) in enumerate(zip(days, fields)):
            ax = fig.add_subplot(grid[row, 3 * column : 3 * column + 3])
            # Land (and, for ocean fields, anything off the ocean grid) is NaN.
            ax.set_facecolor("0.85")
            contour = ax.contourf(
                field["lon"],
                field["lat"],
                field,
                levels=levels,
                cmap=cmap,
                extend="both",
            )
            label = f"{name} anomaly" if anomaly else name
            ax.set_title(f"{label} -- lead +{day} days", fontsize=10)
            if column == 0:
                ax.set_ylabel(f"{title}\nLatitude", fontsize=9)
            if row == len(CONTOUR_ROWS) - 1:
                ax.set_xlabel("Longitude", fontsize=9)
            if column == N_COLUMNS - 1:
                fig.colorbar(contour, ax=ax, label=units)

    # Global series: the coupled system should evolve smoothly, with the
    # ocean stepping once per 5 day cycle and the atmosphere every 6 hours.
    series = (
        ("Global mean 2 m temperature (K)", global_mean(forecast["t2m"])),
        ("Global mean SST (K)", global_mean(forecast["sst"])),
    )
    for index, (label, values) in enumerate(series):
        ax = fig.add_subplot(grid[3, 6 * index : 6 * index + 6])
        ax.plot(lead_days, values, linewidth=1.2)
        ax.set_xlabel("Lead time (days)")
        ax.set_ylabel(label, fontsize=9)
        ax.grid(alpha=0.3)

    fig.suptitle(
        f"SamudrACE {N_CYCLES} coupled cycles ({NSTEPS} atmosphere steps, "
        f"120 days) from CM4 piControl {INITIAL_TIME}",
        fontsize=13,
    )
    fig.savefig(OUTPUT_PATH, dpi=110, bbox_inches="tight")
    print(f"Wrote {OUTPUT_PATH}")


def main() -> None:
    """Run the forecast and write the sanity-check figure."""
    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
    print(f"Running SamudrACE on {device}")

    forecast = run_forecast(device)
    print(f"Forecast variables: {sorted(forecast.data_vars)}")
    print(f"Lead times: {forecast.sizes['lead_time']}")
    for name in RUN_VARIABLES:
        field = forecast[name]
        finite = np.isfinite(field.values)
        print(
            f"  {name}: min {np.nanmin(field.values):.3f} "
            f"max {np.nanmax(field.values):.3f} "
            f"finite {finite.sum() / finite.size:.1%}"
        )
        if not finite.any():
            raise RuntimeError(f"{name} is entirely non-finite")

    plot_forecast(forecast)


if __name__ == "__main__":
    main()

```

earth2studio/data/samudrace.py
```diff
@@ -0,0 +1,546 @@
...
+    def __init__(self, cache: bool = True, verbose: bool = True):
+        self._cache = cache
+        self._verbose = verbose
+
+    @property
+    def cache(self) -> str:
+        """Return the local cache path for downloaded files."""
+        cache_location = os.path.join(datasource_cache_root(), "samudrace")
+        if not self._cache:
+            cache_location = os.path.join(cache_location, "tmp_samudrace")
```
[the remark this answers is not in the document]
elynnwu: Claude: Good catch, fixed in 9197ba4.

`_SamudrACEBase.cache` now routes through the existing `datasource_cache_dir(subdir, persistent, tmp_hash)` helper in `earth2studio/data/utils.py` with a lazily-generated per-instance hash, so cache-disabled sources get `.../samudrace/tmp_samudrace_<hash>` and can no longer clean up each other's downloads. This matches the pattern already in `arco.py`, `gfs.py`, and `wb2.py`.

Two related things in the same change:

- Clean-up moved into a `finally` block (each `__call__` is now a thin wrapper around a new `_fetch_array`, as in `arco.py`), so an interrupted download no longer leaves a partial file behind. Previously clean-up only ran on the success path.
- The open NetCDF handles are closed before their files are removed, rather than just dropped for the GC to collect.

Three tests added in `test/data/test_samudrace_data.py`: temp dirs are distinct per instance and stable across accesses, and a failed download leaves nothing behind (parametrized over both sources). All three fail against the previous code.

One thing intentionally left as-is: when `cache=True` the files land in the HuggingFace hub cache via `hf_hub_download`, not under `datasource_cache_root()`, so `EARTH2STUDIO_CACHE` does not move them. That is deliberate — it dedups with the checkpoint download from the same repo — but happy to note it in the class docstrings if maintainers would rather it were explicit.

earth2studio/data/samudrace.py
```diff
@@ -0,0 +1,614 @@
...
+    list[datetime]
+        List of datetimes.
+    """
+    if isinstance(time, datetime):
+        return [time]
+    if isinstance(time, np.datetime64):
+        time = np.array([time])
+    if isinstance(time, np.ndarray):
+        return list(time.astype("datetime64[s]").astype(object))
+    return list(time)
```
negin513: A plain Python list of `np.datetime64` values falls through to `return list(time)` unconverted, and then crashes downstream: `_validate_time` calls `time.isoformat()` and `SamudrACEForcingData._match_time` accesses `time.month`/`.hour`, both of which raise `AttributeError` on `np.datetime64`.

```python
SamudrACEData()([np.datetime64("0151-01-06T00:00:00")], "t2m")
# AttributeError: 'numpy.datetime64' object has no attribute 'isoformat'
```

This is a supported calling pattern elsewhere in the repo — the core `prep_data_inputs` in `earth2studio/data/utils.py` handles `list[np.datetime64]` per element — and the docstring here recommends `np.datetime64` inputs.

```suggestion
    return [
        t.astype("datetime64[s]").item() if isinstance(t, np.datetime64) else t
        for t in time
    ]
```

(`.astype("datetime64[s]").item()` round-trips model years like 0151 correctly.)

earth2studio/lexicon/base.py
```diff
@@ -385,6 +385,16 @@ def __contains__(cls, val: object) -> bool:
     "t10": "temperature at 10 hPa (K)",
     "z10": "geopotential at 10 hPa (m2 s-2)",
     "w10": "vertical wind at 10 hPa (Pa s-1)",
+    "ewss": "eastward turbulent surface stress (N m-2 s)",
+    "nsss": "northward turbulent surface stress (N m-2 s)",
+    "sithick": "sea ice thickness (m)",
+    "siconc": "sea ice area fraction as seen by the ocean (0-1)",
+    "zos": "sea surface height above geoid (m)",
+    "mttwp": "tendency of total water path due to advection (kg m-2 s-1)",
+    "thetao": "sea water potential temperature (K)",
```
negin513, line 394: The vocab entry declares Kelvin, but SamudrACE serves these in °C: the published HF initial-condition file stores `thetao_*` with `units: 'degC'` (mean ~12.8 °C, while `sst` in the same file is in K, ~286), and `SamudrACELexicon` applies an identity modifier — so users get degC under a vocab promising K, a ~273 discrepancy vs. sources that follow the vocab (cf. `CMIP6Lexicon`, which explicitly converts degC→K for `sst`/`tos`). Since the model itself consumes degC, declaring degC here seems like the right fix rather than converting:

```suggestion
    "thetao": "sea water potential temperature (degC)",
```

Related, same block: `siconc` on line 391 is defined as ocean-view fraction (0-1), but the pre-existing `CMIP6Lexicon` serves E2S `siconc` in percent (it multiplies fractions by 100) — the same canonical name now has a 100x unit difference across sources. Worth reconciling one way or the other while touching this table.

pyproject.toml
```diff
@@ -208,6 +208,11 @@ pangu = [
     "onnxscript",
     "onnxruntime-gpu>=1.21.0",
 ]
+samudrace = [
+    "fme==2026.4.0",
+    "pandas>=2.3.2",
+    "scipy>=1.15.2",
```
negin513, line 213: `SamudrACEForcingData._open_dataset` uses `decode_times=xr.coders.CFDatetimeCoder(use_cftime=True)` (`earth2studio/data/samudrace.py:573`), but the `xarray.coders` namespace — and passing a coder object to `decode_times` at all — only exists in xarray >= 2025.01. The project-wide floor is still `xarray[parallel]>=2023.1.0` and fme's own xarray dependency is unconstrained, so an environment legally resolved to an older xarray raises `AttributeError: module 'xarray' has no attribute 'coders'` the first time forcing data is opened.

Declaring the real requirement on the extra is the least invasive fix (the alternative — `decode_times=True, use_cftime=True` — is the spelling xarray is deprecating in favor of coder objects):

```suggestion
    "scipy>=1.15.2",
    "xarray>=2025.1.0",
```

earth2studio/utils/time.py
```diff
@@ -19,6 +19,33 @@
...
+    ----------
+    time : np.ndarray
+        Numpy datetime64 array
+
+    Returns
+    -------
+    np.ndarray
+        The array as ``datetime64[ns]`` when representable, otherwise unchanged
+    """
+    if time.size == 0 or np.all((time >= _NS_TIME_MIN) & (time <= _NS_TIME_MAX)):
```
negin513: NaT elements silently defeat the ns normalization: every comparison against `NaT` is False (like NaN), so a single `NaT` in the array makes `np.all(...)` False and the function takes the out-of-range branch even when all real timestamps are ordinary. Since `to_time_array` previously did `.astype("datetime64[ns]")` unconditionally, this is a behavior change in shared code:

```python
to_time_array([np.datetime64("NaT"), np.datetime64("2024-01-01")])
# before: dtype datetime64[ns]
# after:  dtype datetime64[D]
```

`NaT` is representable in ns — it's incomparable, not out of range — so this also contradicts the docstring ("as ``datetime64[ns]`` when representable"). Masking NaT out of the range check restores the previous behavior:

```suggestion
    valid = ~np.isnat(time)
    if time.size == 0 or np.all((time[valid] >= _NS_TIME_MIN) & (time[valid] <= _NS_TIME_MAX)):
```

test/models/px/test_samudrace.py
```diff
@@ -0,0 +1,763 @@
...
+
+import numpy as np
+import pytest
+import torch
+import xarray as xr
+
+from earth2studio.models.px.samudrace import SamudrACE
+from earth2studio.utils import handshake_dim
+
+pytest.importorskip("fme")
```
negin513, line 28: As written, this suite never executes in any CI environment — the model ships with zero executed test coverage:

- `samudrace` is (deliberately, per the pyproject note) excluded from the `all` extra, so the `test-px-models` tox env (`extras = all`) has no `fme` installed and this `importorskip` skips the entire module.
- The `test-px-models-conflicts` env in `tox.ini` — which exists exactly for extras excluded from `all`, and already has a dedicated `uv sync --extra ace2` + `test_ace2.py` block for the analogous fme-based model — was not extended with a samudrace block.
- `test/conftest.py`'s `_TEST_DEPENDENCIES` has an entry for `test_ace2.py` but none for this file, so the version-aware dependency gate sibling optional-dep tests use doesn't apply either.

So a regression in `earth2studio/models/px/samudrace.py` merges green (all tests report as *skipped*, which reads as passing). Suggested fix, mirroring the ace2 pattern:

```ini
# tox.ini [testenv:test-px-models-conflicts]
    uv sync --extra samudrace
    pytest {posargs:-s --cov --cov-append --slow --package --testmon} \
        test/models/px/test_samudrace.py
```

```python
# test/conftest.py _TEST_DEPENDENCIES
"test/models/px/test_samudrace.py": ["samudrace"],
```

(Related data point: I ran the suite's usage pattern against the real checkpoint on a GPU machine — the model itself works well, but the CPU-parametrized device tests in this file fail on any CUDA machine because fme pins its normalizer/mask tensors to `fme.get_device()`; `model.to("cpu")` raises `RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!`. Worth knowing before wiring the suite into CI, since GPU runners would go red — details in a separate comment if useful.)

negin513:
### Independent sanity run (GB200, CUDA)

Ran one full coupled cycle (20 × 6 h = 5 days) from the published year-0151 piControl initial condition, loading the checkpoint through `load_default_package()` (`hf://allenai/SamudrACE-CM4-piControl`) and the IC/forcing through `SamudrACEData`/`SamudrACEForcingData`:

SamudrACE PR1060 sanity run

Observations:

- Stable integration: global-mean t2m oscillates ~274.3–275.3 K with no drift or blow-up; min/max stay physical (~220/318 K).
- The Δt2m map over the cycle is concentrated in NH winter midlatitudes (synoptic systems) with near-zero change in the tropics — the expected January pattern.
- Coupled semantics behave as designed: SST is constant through the 19 inner steps and updates at the cycle boundary (285.98 → 286.01 K at step 20); ocean fields are NaN over land at a constant 0.740 finite fraction with no NaN creep into atmosphere fields.
- Performance: model load 6.5 s (warm HF cache), one 5-day coupled cycle in 6.7 s on a single GB200.
- The production `hf://` package-resolution path works against the real repo (65 s cold / 6.5 s warm).

One environment note for aarch64 users: with pip torch 2.12.1+cu130, the bundled `nvidia-cudnn-cu13==9.20` wheel doesn't ship `libcudnn_engines_tensor_ir.so`, so cuDNN dlopens the system copy and fails with `CUDNN_STATUS_SUBLIBRARY_VERSION_MISMATCH` if the OS carries a different cuDNN (9.21 here). Upgrading to `nvidia-cudnn-cu13>=9.24` resolved it — nothing to change in this PR, just a heads-up for the docs if it comes up.

negin513 (approved):

negin513:
### Resolved merge with `main` available

This PR now conflicts with `main` (mostly the docs refactor in #1054). I resolved the merge and validated it, but GitHub doesn't allow maintainer pushes to a fork branch when the push touches `.github/workflows` (which merging current `main` unavoidably does), so I can't push it here directly. The resolved merge is ready to pull:

```bash
git pull https://github.com/negin513/earth2studio.git samudrace-merge-main
git push
```

What the resolution does (single merge commit, no changes to your feature commits):

- **`CHANGELOG.md`** — kept both sides' Added/Fixed entries.
- **`docs/modules/*.rst`** — deleted on `main` by the docs refactor; the `SamudrACE` / `SamudrACEData` / `SamudrACEForcingData` entries are ported to the new `models_px.md` and `datasources_analysis.md` autosummary lists.
- **`docs/userguide/about/install.md`** — took `main`'s rewritten page; the SamudrACE install notes (fme 2026.4.0 pin, ace2/torch-harmonics conflicts) moved into `install_options.yml` as a `samudrace` entry following the new schema.
- **`pyproject.toml`** — kept the ace2/samudrace exclusion note for the `all` extra; dropped the cosmo note since `main` re-added cosmo to `all`.
- **`uv.lock`** — regenerated from `main`'s current lock (`uv lock`); minimal diff, just the fme 2026.4.0 / samudrace additions.

Validation on the merged tree: `test/utils/test_time.py`, `test/data/test_samudrace_data.py`, `test/data/test_data_utils.py`, and `test/models/px/test_samudrace.py` — 78 passed. The only failures (6) are the pre-existing fme device-pinning behavior on CUDA machines discussed in the review comments, unrelated to the merge.

negin513:
/ok to test [7c0611f](https://github.com/NVIDIA/earth2studio/pull/1060/commits/7c0611fb9595fad9a4e18471fdf183051389b074)

PR #1058, opened 2026-08-12
Optimize and migrate ISD, IBTrACS, CFS reforecast, and OPERA data sources on obstore

negin513:
# Earth2Studio Pull Request

## Description

Closes NVIDIA/physicsnemo-roadmap#2883
closes NVIDIA/physicsnemo-roadmap#2886
closes NVIDIA/physicsnemo-roadmap#2887
closes NVIDIA/physicsnemo-roadmap#2888

Migrates four data sources from fsspec/s3fs/HTTP to obstore, continuing the migration pattern established by the GOES (#1024), GOES GLM, and Himawari AHI (#1043) PRs:

- **ISD** — station CSV and station-history S3 reads move from s3fs to obstore
- **IBTrACS** — HTTP downloads and Last-Modified cache validation move to obstore
- **CFS reforecast** — full-file HTTP fetching moves to obstore, preserving historical cache keys
- **OPERA** — HTTP object reads move to obstore, retaining URL-derived cache keys

Each migration is a single-file commit. Cache-key formats are unchanged, so existing local caches remain valid.

Three additional changes surfaced by benchmarking this PR:

- **CFS reforecast NCEI URL fix** — NCEI retired the `/data/climate-forecast-system/access/reforecast` path (it now serves only a readme), breaking `CFS_Reforecast_FX` / `CFS_Reforecast_FX_Flux` on `main` as well. Both classes now point at the relocated archive (`https://www.ncei.noaa.gov/oa/prod-cfs-reforecast`) with its renamed product subdirs; file naming and date-path layout are unchanged.
- **Single-pass GRIB decode** — CFS reforecast decoding called `pygrib.select()` once per variable, and each call rescans every message in the file (~1.2 s on a 524-message pgbf file), costing ~100 s for a full-lexicon request while the 22 MB download takes ~2 s. Variables are now resolved in one pass over the file's messages, preserving `select()`'s first-match semantics. Decoded arrays verified bit-identical to the old path for all 81 pgbf variables.
- **Shared GRIB decode helper** — GFS, HRRR, GEFS, and CFS each carried a byte-identical private `_decode_*_grib` helper; all four are consolidated into `decode_grib_message` in `earth2studio.data.utils` (net −77 lines). Pure refactor: identical decode logic and thread dispatch, no performance or behavior change for those sources.

### Performance

Each source fetched its full lexicon with `cache=False`, 3 reps, same host and time window; medians shown. CFS rows used the relocated NCEI endpoint on both sides for an apples-to-apples comparison.

[...]

PR #1062, opened 2026-08-12
Migrate JPSS VIIRS, ATMS, and CrIS data sources to obstore

negin513:
# Earth2Studio Pull Request

## Description

Closes NVIDIA/physicsnemo-roadmap#2879, closes NVIDIA/physicsnemo-roadmap#2880, closes NVIDIA/physicsnemo-roadmap#2881.

Migrates the three JPSS data sources from s3fs to obstore as one batched effort — they share the NOAA NESDIS S3 bucket layout (`noaa-nesdis-{n20,n21,snpp}-pds`) and the granule-discovery-by-listing pattern, so the migration and test-fixture work amortizes across all three. Continues the pattern established by GOES/GLM (#1042), Himawari AHI (#1043), and #1058.

- **JPSS (VIIRS)** — single-bucket store per instance. Day-directory listings now go through `obstore_list_prefix` with per-instance memoization: previously every `(time, variable)` pair re-listed the same day directory (plus one more per geolocation fetch); now a completed day is listed once per product folder, while in-progress days are always re-listed so near-real-time polling still sees new granules. `available()` moves from sync s3fs to a sync obstore listing that verifies both the data and geolocation prefixes (one listed entry proves existence).
- **JPSS_ATMS** — instances span multiple satellites, so `_async_init` builds one store per bucket. The `_ls` + `FileNotFoundError` handling becomes an empty-listing check (obstore lists missing prefixes as empty). Download retry loop, backoff, and exception semantics are unchanged; only the transfer line moved to `obstore_read_range`.
- **JPSS_CRIS** — same per-bucket store layout. The dual SDR/GEO day-listing (`asyncio.gather` at the old jpss_cris.py:839) stays concurrent as two gathered `obstore_list_prefix` coroutines, and the entire two-level `FileNotFoundError` fallback block is deleted — empty listings subsume it. Completed-day listings are memoized per bucket.

**Warm-cache compatibility:** cache-key hashing is byte-identical everywhere — VIIRS hashes the same bucket-prefixed `bucket/key` paths s3fs used to return, and ATMS/CrIS rebuild the exact `s3://bucket/key` URI strings that `_cache_path` hashes — so local caches populated before the migration remain valid.

**Tests:** each source gains offline tests that inject fake obspec stores (the `_FakeStore` pattern from test_himawari_ahi.py) rather than patching obstore internals — covering granule discovery/time-window selection, SDR↔GEO pairing and missing-GEO skip (CrIS), empty-listing/data-gap branches, `available()`, and a full VIIRS `__call__` mock over synthetic HDF5 granules. Combined suite: 57 passed offline, plus the two live `available()` network tests xpass through the new obstore path.

`earth2studio/data/utils.py` is untouched — all helpers needed already exist — so there is no overlap with #1058.

## Checklist

- [x] Changelog updated.

## Performance

Cold-fetch medians (`cache=False`, same host; interleaved A/B where noted):

| Source | File type | Benchmark | main | PR | Speedup | Source of the win |
|---|---|---|---|---|---|---|
| VIIRS | HDF5 (SDR/EDR granules, ~1 data + 1 GEO file per time/var, S3) | 2 times × 2 M-band vars | 3.7s | 2.4s | 1.6× | obstore transfer path + day-listing memoization |
| ATMS | BUFR (one scan line per file, ~37 small files for ±10m, S3) | n20 ±10m, 931k rows | 2.8s | 0.9s | 3.1× | obstore + vectorized decode + fetch/decode pipelining |
| VIIRS | HDF5 (M-band SDR granules, S3) | 1 time × full M-band lexicon (16 vars, 17 files) | ~6.4s | ~7.1s | ~1.0× | parity — 17 granule downloads dominate (high S3 variance, 6 reps/side) |
| CrIS | HDF5 (SCRIF + GCRSO pairs, ~16 MB each, S3) | n20 ±2m, 3 channels | ~8.2s | ~4.4s | 1.9× | chunked parallel byte-range downloads |
| CrIS | HDF5, full spectrum | n20 ±2m, all 2211 channels (16M rows) | ~7.0s | ~4.7s | ~1.5× | chunked downloads; decode share grows so the transfer win dilutes |

**Variable-space coverage:** ATMS is inherently the full space (the single `atms` variable carries all 22 channels via `sensor_index`). VIIRS is covered for the full M-band lexicon (16 variables); I-band (5 variables, ~4× larger granules) and L2 EDR (14 variables) run the identical code path with different file sizes/counts and were not separately benchmarked. CrIS is covered at both the subset (3 of 2211 channels) and full-spectrum extremes.

VIIRS full M-band reads as parity because 17 granule downloads dominate and S3 variance is high (worst rep 36s on a slow granule — bucket-side, both code paths affected). CrIS full-spectrum dilutes from 1.9× to ~1.5× because decode's share grows relative to transfer at 16M rows; its remaining warm-path costs are S3 listing latency (~60%) and the Planck inversion in `radiance_to_bt`, both out of scope here.

Beyond the migration itself, the follow-up commits:

- **Vectorize ATMS BUFR decode** — numpy column assembly + one Arrow table per BUFR message instead of ~1M per-(FOV, channel) dict appends and eleven `astype` passes; warm decode+compile 1.8s → 0.6s for 37 granules. Verified **bit-identical** via `assert_frame_equal(check_exact=True)` on the full 931k-row frame (values, dtypes, row order, NaN retention, NaT skips).
- **Pipeline ATMS decode behind downloads** — each file decodes in a worker thread (serialized by an eccodes lock; eccodes is not thread-safe) as soon as its download lands; decode deduplicated to once per unique file. Same exact-equality parity.
- **VIIRS decode off the event loop** — h5py blocks move to `asyncio.to_thread` under an HDF5 lock (himawari pattern). No measured end-to-end change at small request sizes (interleaved A/B within noise); prevents decode from serializing downloads on large fan-outs. Output verified identical.
- **CrIS chunked downloads** — large granules fetch as 1 MiB × 8 concurrent byte ranges instead of one single-stream GET (chunk size tuned empirically; 1 MiB clearly beat 2–8 MiB). sha256 byte parity verified against the single-range path; also reduces variance (4.5–5.8s vs 6.0–12.6s spread).

CrIS decode was measured at 0.14s warm and left alone — it is transfer-bound, which is why its win comes from the download path.


earth2studio/data/jpss.py
```diff
@@ -295,27 +315,41 @@ async def fetch_data(
...
-        # NetCDF file - use xarray
-        with h5py.File(local_file, "r") as h5:
-            # Get dataset from HDF5 file
-            if self._product_type in ["I", "M"]:
-                ds = h5["All_Data"][folder + "_All"][dataset_name]
-            else:
-                ds = h5[dataset_name]
+        # HDF5 decode is CPU-bound; run in a worker thread so it overlaps
+        # with concurrent granule downloads instead of blocking the event loop
+        processed_data = await asyncio.to_thread(
```
NickGeneva, line 323: maybe stupid question, but does the to_thread do much here if theres a H5 lock? Is there a perf difference if _decode_data is just sync?

earth2studio/data/jpss.py
```diff
@@ -529,11 +573,26 @@ async def _get_s3_path(
...
+        # Day directories that can no longer gain files are memoized so
+        # repeated (time, variable) requests issue a single LIST; the
+        # bucket-prefixed paths match the historical cache-key scheme.
+        day_end = datetime(time.year, time.month, time.day) + timedelta(days=1)
+        prefix = base_url.split(f"{bucket}/", 1)[1]
+        keys = await obstore_list_prefix(
+            self.store,
+            prefix,
+            cache=self._day_listing_cache,
+            cacheable=day_end + timedelta(hours=1)
```
NickGeneva, line 597: Is plus 1 hour here just picked here as a sort of heuristic. Or a specific reason. Just curious, thats all

earth2studio/data/jpss_cris.py
```diff
@@ -903,10 +1008,14 @@ async def _fetch_remote_file(self, s3_uri: str) -> None:
...
 
+        if self.stores is None:
+            raise ValueError("Object stores are not initialized")
+        bucket, key = s3_uri.removeprefix("s3://").split("/", 1)
+
         last_exc: Exception | None = None
         for attempt in range(1, self._retries + 1):
             try:
-                data = await self.fs._cat_file(s3_uri.replace("s3://", "", 1))  # type: ignore[union-attr]
+                data = await _read_object_chunked(self.stores[bucket], key)
```
NickGeneva: should this section be replaced with the obstore save to cache util?

earth2studio/data/jpss_cris.py
```diff
@@ -387,6 +391,90 @@ def _iet_to_utc(
...
+# gather_with_concurrency(max_workers) and the per-bucket store keeps
+# max_pool_connections == max_workers (24); a typical +/-2m request touches
+# ~4 files, so files x chunks (~32) stays near the pool size. For large
+# requests the worst case is max_workers x 8 in-flight ranges, but
+# pool_max_idle_per_host only caps idle connection reuse (not concurrency)
+# and with many files in flight aggregate bandwidth is already saturated.
+_MAX_CHUNK_CONCURRENCY: int = 8
+
+
+async def _read_object_chunked(
```
NickGeneva: This was done before the other PC pr with the streamable util, but maybe this should be part of that function can a potential set up would be streaming to bytes or temp files. so for large files you can use temp. Dont need to do this in this PR, just an idea

NickGeneva (approved):
Over all looks good, main item is the streaming util in the cris streaming, if that should be something thats more general util. Approving to unblock.

earth2studio/data/jpss.py
negin513: Not a stupid question at all! You're right that the lock serializes the decodes themselves, so threading doesn't help decode-vs-decode. What it buys us is keeping the event loop free — if `_decode_data` ran sync, every in-flight granule download would stall while each decode runs. With `to_thread`, downloads keep overlapping with whichever decode holds the lock. I expanded the comment to spell this out.

earth2studio/data/jpss.py
negin513: Yep, just a heuristic :) Granules land in a day's directory with some processing/upload lag after the observations themselves, so treating a day as closed exactly at midnight UTC risks memoizing a still-filling listing. The extra hour is a conservative buffer for that lag. Added a comment explaining it.

earth2studio/data/jpss_cris.py
negin513: Good call — done! `_fetch_remote_file` now goes through `obstore_fetch_to_cache`, with a new `chunked=True` option so it keeps the concurrent byte-range download. It passes an explicit `cache_key` so the historical `sha256(s3_uri)` cache file names stay the same and warm caches remain valid.

earth2studio/data/jpss_cris.py
negin513: Since this was your main item I went ahead and did it here: `_read_object_chunked` is now `obstore_read_chunked` in `earth2studio/data/utils.py`, and `obstore_fetch_to_cache` grew a `chunked` flag that routes through it (whole-object fetches only — chunked + byte-range raises). Streaming to temp files for really large objects is a nice idea, saving that for the follow-up utils work. Tests moved to `test_data_utils.py` next to the other obstore util tests. Thanks for the review!

earth2studio/data/jpss.py
negin513: Followed up with an actual measurement to be sure: cold-cache fetch of 9 M-band granules (3 times x 3 variables, NOAA-20), 4 alternating rounds each way. Sync decode (inlined on the event loop): median 5.2s. Threaded (`to_thread`, current code): median 4.1s — about 20% faster, with 3 of 4 threaded runs landing below the fastest sync run. So it's a modest but real win from downloads overlapping the locked decode, not a dramatic one — worth keeping since it's free.

negin513:
/ok to test [384a75a](https://github.com/NVIDIA/earth2studio/pull/1062/commits/384a75a29b4383b2b7d672ca8a04fa1203e62d30)

PR #1024, opened 2026-07-28
Update GitHub CI workflow

NickGeneva:
# Earth2Studio Pull Request

## Checklist

- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/earth2studio/blob/main/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.
- [ ] The [CHANGELOG.md](https://github.com/NVIDIA/earth2studio/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/earth2studio/issues) is linked to this pull request.
- [ ] Assess and address Greptile feedback (AI code review bot for guidance; use discretion, addressing all feedback is not required).

PR #1063, opened 2026-08-13
Optimize and migrate GHCN data sources on obstore with by_station fetching

negin513:
# Earth2Studio Pull Request

## Description

Closes NVIDIA/physicsnemo-roadmap#2884.

Migrates the GHCN daily and hourly data sources from s3fs/fsspec to obstore, continuing the migration series (GOES #1024, Himawari AHI #1043, ISD/IBTrACS/CFS-reforecast/OPERA #1058, MRMS/NClimGrid #1061):

- **GHCNDaily** — async s3fs replaced by an obstore S3 store: partition listings via `obstore_list_prefix` (past years' `by_year` partitions are immutable and memoized per instance; the current year is always re-listed), parquet reads via `obstore_read_range`, catalog files and `available()` via `obs.get`/`obs.list`. Cache hash strings unchanged, warm caches remain valid.
- **GHCNHourly** — the fsspec `https` filesystem replaced by an obstore HTTP store for station-year parquet reads and the station-list CSV.
- **GHCNDaily by_station fetching** — the old path downloaded *global* `by_year` year/element partitions (~100 MB each) regardless of how few stations were requested. Requests with ≤500 unique stations now fetch `parquet/by_station/STATION=<id>/ELEMENT=<el>/` files (~50–100 KB each, full period of record); larger station lists fall back to `by_year`, where request count is bounded by years × elements rather than stations × elements.
- **Mutable partition cache expiry** — `by_station` files and the current year's `by_year`/station-year partitions receive new observations on NOAA's roughly daily update cadence, so their disk-cached copies now expire after 24h (`_cache_file_fresh`) instead of being served indefinitely; immutable past-year partitions still cache forever.
- **`GHCNHourly` blocking I/O** — `_fetch_station_year`'s parquet parse + disk-cache write now runs via `asyncio.to_thread`, matching `GHCNDaily._fetch_partition`, so it no longer stalls the event loop while other station-year fetches are in flight.
- **`async_workers` default raised 16 → 32** — GHCN fetches are small independent HTTP/S3 requests dominated by round-trip latency rather than bandwidth, so throughput scales with concurrency well past the old default (see Performance below).

### Correctness note

The `by_station` layout is also more complete than `by_year`: the 2022/2023 `by_year` TAVG partitions are missing TAVG rows for stations that carry them in `by_station` (verified by scanning every file of the 2023 TAVG partition — 2.16 M rows, zero rows for the benchmark stations). Station queries therefore now return `t2m` observations the old path silently dropped. Every row common to both paths is value-identical.

### Performance

Full lexicon, `cache=False`, 2 stations, 3 reps, same host and time window:

| Data source | Layout | Before (main) | After (PR) | Speedup |
|---|---|---|---|---|
| GHCNDaily (12 vars) | by_year global partitions → by_station files | 55.4–60.0 s | 2.0–2.4 s | **~26×** |
| GHCNHourly (8 vars) | per station-year parquet, HTTPS | 0.9–1.1 s | 0.9–1.5 s | parity |

The obstore migration alone benchmarked at parity on both classes (GHCNDaily ~57 s both ways — the global-partition download dominated); the ~26× comes from the by_station redesign. GHCNHourly was already minimal: one small parquet per station-year.

**Concurrency (`async_workers`):** with per-request latency dominating (not per-task CPU work), throughput scales with concurrency well past the old default of 16. Measured live against NOAA with `cache=False`:

| Data source | Workload | 16 workers | 32 workers | 64 workers |
|---|---|---|---|---|
| GHCNHourly | 20 stations × 20 years (200 fetches) | 14.5–17.4 s | 12.1–18.5 s* | 6.5–7.0 s |
| GHCNDaily (by_station) | 20 stations × 6 vars (120 fetches) | 6.6–10.7 s | 5.1–7.5 s | 4.9 s |

\* one 32-worker GHCNHourly run hit an 18.5s network outlier; the rest cluster with the 64-worker numbers, well below the 16-worker baseline.

Default raised to 32 — it captures most of the available speedup (roughly halving wall time for large station lists) without doubling concurrent request load against NOAA's public endpoints again for the smaller additional gain 64 offers. Callers with large station counts can pass a higher `async_workers` explicitly for further headroom.

## Checklist

- [x] New or existing tests cover these changes (`test_ghcn_daily.py`, `test_ghcn_hourly.py` — 61 offline tests pass; mocks updated to patch the shared obstore helpers).
- [x] An issue is linked to this pull request (physicsnemo-roadmap task issue above).

## Dependencies

No new dependencies; obstore and obspec are already core dependencies.


CHANGELOG.md
```diff
@@ -26,6 +26,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   hours are memoized per instance while the current hour is always re-listed
 - Migrated Himawari AHI data source from s3fs to obstore with memoized
   minute-directory listings (scans older than an hour)
+- Migrated GHCNDaily and GHCNHourly data sources from s3fs/fsspec to obstore;
+  by_year partition listings of past years are memoized per instance
+- GHCNDaily now fetches per-station parquet files (`parquet/by_station`) for
```
NickGeneva: nit: too verbose

earth2studio/data/ghcn.py
```diff
@@ -1007,41 +1077,40 @@ async def fetch(
...
+        )
+
+        coros = [
+            async_retry(
+                self._fetch_station_year,
+                station_id,
+                year,
+                retries=self._retries,
+                backoff=1.0,
+                task_timeout=60.0,
```
NickGeneva, line 1135: is there a async_timeout parameter for this data frame source? could that be used here?

earth2studio/data/ghcn.py
```diff
@@ -506,137 +503,214 @@ async def fetch(
...
+
+        s3_path = (
+            f"{self._S3_BUCKET}/parquet/by_station/"
+            f"STATION={station}/ELEMENT={element}/"
+        )
+        cache_hash = hashlib.sha256(s3_path.encode()).hexdigest()
+        parquet_path = os.path.join(self.cache, f"{cache_hash}.parquet")
+
+        if self._cache and os.path.isfile(parquet_path):
+            return await asyncio.to_thread(pd.read_parquet, parquet_path)
```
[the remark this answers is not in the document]
NickGeneva: I think greptile is correct here, the cachable for the actual data is limited by if the obs are not in the current year so prevents us having a stale cache for files getting updated for current year.

but if these files are getting updated, i think we need a similar condition?

NickGeneva (commented):
LGTM over all.

Main item is greptiles comment, my agent complained about this too... can you double check this. Since we didnt cache that before we didnt have to worry about it on main

earth2studio/data/ghcn.py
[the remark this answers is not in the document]
negin513: Yeah, you're both right — thanks for catching this! And it went a little deeper: the `year < now().year` condition only protects the LIST memoization, but the disk-cache check at the top of `_fetch_year_element` (and the GHCNh station-year one) has no year guard, so a current-year partition cached in January would be served stale all year too. That part was already on main, but it made sense to fix it all together here.

Fix (68348f73): mutable partitions — by_station files plus current-year slices — now expire from the disk cache after 24h, matching NOAA's update cadence; past-year partitions still cache forever. One heads-up on greptile's `if not self._cache` suggestion: that branch could never actually run, since the tmp cache is wiped after every call and pairs are dedup'd within one.

earth2studio/data/ghcn.py
negin513: Good question! `async_timeout` is the overall budget on the whole fetch (600s default), so using it per-task would let one hung download starve the retries — they're different layers. But your instinct flushed out a real problem: the hard-coded 60s was too tight for the ~100 MB by_year partitions (anything under ~1.7 MB/s would time out on every retry). Now it scales with the layout: 60s for the small by_station files, 300s for by_year.

CHANGELOG.md
negin513: Fair! Condensed to one bullet :)

earth2studio/data/ghcn.py
NickGeneva: Got it, makes sense. Sounds reasonable enough. I have encountered instances where hard coded timeouts despite being conservative causes issues for some with really slow networks or different parts of the world.

But I think this is reasonable enough such that we can change it if ever needed. Thanks for updates.

NickGeneva (approved):
LGTM

negin513:
ok to test

negin513:
/okay to test [79f73ec](https://github.com/NVIDIA/earth2studio/pull/1063/commits/79f73ec75f8eac1f19884c81deb19217f703c023)

negin513:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

negin513:
/ok to test [317396d](https://github.com/NVIDIA/earth2studio/pull/1063/commits/317396dab6055bb4b88749e3cb9524c8b7728b6e)

PR #1065, opened 2026-08-13
Migrate Planetary Computer data sources to obstore with GRIB byte-range fetches

negin513:
# Earth2Studio Pull Request

## Description

Closes the Planetary Computer obstore migration task (NVIDIA/physicsnemo-roadmap#2750).

Replaces `planetary_computer.sign()` + httpx whole-file streaming with obstore `AzureStore` + `PlanetaryComputerCredentialProvider` across all five Planetary Computer data sources, and adds GRIB index byte-range fetches to `PlanetaryComputerECMWFOpenDataIFS`.

**Commit 1 — obstore migration (all 5 sources).** Asset downloads go through the shared `obstore_fetch_to_cache` + `async_retry` helpers (same pattern as the GOES/Himawari migrations); cache filenames are unchanged so warm caches stay valid. SAS tokens are fetched and renewed automatically by the credential provider, with one `AzureStore` cached per (storage account, container). The provider is constructed with explicit `account_name`/`container_name` — constructing it from the full asset URL roots the store at the blob path and doubles the key. STAC search via `pystac-client` is unchanged.

**Commit 2 — GRIB byte-range fetches for ECMWF Open Data IFS.** ECMWF Open Data items carry their GRIB index as a first-class `index` STAC asset (JSON-lines with `_offset`/`_length` per message). The IFS source now resolves the messages serving the requested variables and issues a single multi-range read instead of downloading the whole ~120 MB analysis file. Concatenated GRIB messages form a valid GRIB file, so pygrib extraction is unchanged and values are bit-identical (asserted in a new benchmark test). Whole-file fallback remains for items without an index asset; subset cache entries are keyed by the variable set. The base class now exposes an overridable `_fetch_asset(store, plan)` hook (whole-file default) so other collections can adopt partial fetches without touching the download pipeline.

### Performance per migrated data source

Cold-cache A/B against the live services (2026-08-12, best of 3 runs, `cache=False`):

| Data source | Request | Before (sign + httpx) | After (obstore) | Change |
|---|---|---|---|---|
| `PlanetaryComputerOISST` | 6 times × 2 vars (12 × ~1.5 MB files) | 2.6 s | 2.9 s | parity |
| `PlanetaryComputerSentinel3AOD` | 1 time × 2 vars | 10.4 s | 11.0 s | parity |
| `PlanetaryComputerMODISFire` | 2 times × 2 vars | 2.5 s | 3.3 s | parity |
| `PlanetaryComputerGOES` | 2 times × 1 var (~2 × 70 MB MCMIP) | 12.0 s | 12.1 s | parity |
| `PlanetaryComputerECMWFOpenDataIFS` | 1 time × 5 vars (sfc + pl + soil) | 33.3 s (whole 117.6 MB file) | 5.5 s (3 MB ranged) | **6x** |

The four NetCDF/COG sources are transfer-bound whole-file downloads in both stacks, so parity is expected — the migration is a robustness/consistency/dependency win for them. IFS is the performance change, and it scales with request size:

| Variables requested (of 160) | Data fetched | Time | Speedup vs whole file |
|---|---|---|---|
| 1 | 0.03 MB | 0.14 s | ~240x |
| 3 | 0.7 MB | 0.28 s | ~115x |
| 8 | 3.4 MB | 0.9 s | ~38x |
| 16 | 5.0 MB | 1.4 s | ~23x |
| 32 | 15.5 MB | 3.8 s | ~9x |
| 64 | 39.4 MB | 9.7 s | ~3.4x |
| 100 | 68.6 MB | 18.0 s | ~1.8x |
| 160 (all) | 117.6 MB | 33.0 s | parity |

Speedup tracks the transfer cut almost exactly (no per-range request penalty), and there is no crossover where whole-file wins — worst case is parity at all 160 messages. Typical few-variable requests land in the 20–200x range; full ~70-variable model state vectors get ~2–3x per timestep.

### Verification

- All 5 sources smoke-tested end-to-end against the live services; IFS ranged values verified bit-identical to the whole-file path (`test_planetary_computer_ifs_ranged_benchmark`, slow-marked, reproduces the A/B).
- New offline unit tests for index-entry matching, blob-href parsing, and subset cache keying.
- All 160 IFS lexicon variables validated against a live index; the 3 time-maxima variables (`max_t2m`, `min_t2m`, `max_fg10m`) are absent from step-0h files entirely — the previous whole-file path could not fetch them either (no regression; possible lexicon follow-up).

## Checklist

### Post-review hardening (commit 3)

An adversarial review pass surfaced four correctness issues at the retry/cache seam, all fixed:

- **Retry classification**: obstore exceptions subclass `Exception` (not `OSError`), so transient failures were never retried while permanent `FileNotFoundError`s were retried 4x then masked. The retry tuple now includes `obstore.exceptions.BaseError`, and a new `async_retry(no_retry=...)` parameter fails permanent errors fast with their informative message (missing-variable requests: ~15 s masked → 0.5 s clear).
- **Atomic cache writes**: `obstore_fetch_to_cache` now streams whole objects to disk (bounded memory, was fully RAM-buffered) and publishes via temp file + rename; the IFS ranged writer uses a shared `atomic_write_bytes` helper. Interrupted or concurrent downloads can no longer leave a partial file as a poisoned cache entry — this hardens **all** obstore-migrated sources, not just Planetary Computer.
- **Stall bound**: each fetch attempt is capped at 300 s so one stalled transfer cannot consume the whole `async_timeout` budget without retrying.
- **Async credential provider**: swapped the sync requests-based provider for `PlanetaryComputerAsyncCredentialProvider`, so SAS fetch/renewal never blocks the concurrent download tasks.

Plus smaller cleanups: collapsed the duplicated dataset-key/level-type tables, fixed the stale IFS lexicon docstring (referenced the removed `planetary_computer` package), corrected `request_timeout` docstrings, and updated the Himawari test mock to the full obspec `GetResultAsync` protocol.

## Dependencies

Removed from the `data` extra: `planetary-computer`, `httpx`. Added: `aiohttp>=3.9.0` + `aiohttp-retry>=2.8.3` (used by obstore's `PlanetaryComputerAsyncCredentialProvider`; aiohttp previously transitive via core `s3fs`/`gcsfs`). No other new packages.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


earth2studio/data/planetary_computer.py
```diff
@@ -989,6 +1070,94 @@ def __init__(
...
+                for entry in matches
+            )
+        sorted_ranges = sorted(ranges)
+        buffers = await store.get_ranges_async(
+            plan.key,
+            starts=[start for start, _ in sorted_ranges],
+            ends=[end for _, end in sorted_ranges],
+        )
+        data = b"".join(bytes(buffer) for buffer in buffers)
+        await asyncio.to_thread(atomic_write_bytes, plan.local_path, data)
```
NickGeneva, line 1139: why does the write need to be atomic here? Is a plans local path not guaranteed to be unique?

earth2studio/data/utils.py
```diff
@@ -997,6 +1007,34 @@ async def obstore_read_range(
     return bytes(data)
 
 
+def atomic_write_bytes(path: str | os.PathLike, data: bytes) -> None:
+    """Writes bytes to a file atomically via a same-directory temp file.
+
+    Readers never observe a partially written file: content is written to a
+    unique temporary file in the destination directory and published with an
+    atomic ``os.replace``. Concurrent writers of the same path are safe (the
```
NickGeneva, line 1076: when is this useful? since the last process just wins and overwrites all other ones, what is this preventing? Seems to me like if this is occurring it should trigger some error rather than silently overwritting like this but I dont understand its use at the moment.

earth2studio/data/utils.py
```diff
@@ -1038,10 +1081,30 @@ async def obstore_fetch_to_cache(
     cache_path = os.path.join(cache_dir, cache_key)
     if Path(cache_path).is_file():
         return cache_path
-    data = await obstore_read_range(
-        store, key, byte_offset=byte_offset, byte_length=byte_length
-    )
-    await asyncio.to_thread(Path(cache_path).write_bytes, data)
+    if byte_offset == 0 and byte_length is None:
```
NickGeneva: Are there any potential down sides to always streaming to tmp files versus just in memory. I understand the need for large files but what about data thats maybe like a few MB or less?

earth2studio/data/utils.py
```diff
@@ -1038,10 +1081,30 @@ async def obstore_fetch_to_cache(
...
+        # Whole object: stream chunks straight to a temp file so large assets
+        # are never buffered fully in memory, then publish atomically
+        try:
+            resp = await store.get_async(key)
+        except (FileNotFoundError, obs.exceptions.NotFoundError):
+            raise FileNotFoundError(f"Object {key} not found in store") from None
```
NickGeneva: the "from None" is odd, why not raise the normal error if ob store already has a Not Found exception? Is it confusing?

earth2studio/data/utils.py
```diff
@@ -1038,10 +1081,30 @@ async def obstore_fetch_to_cache(
...
+            with os.fdopen(fd, "wb") as file:
+                async for chunk in resp:
+                    file.write(chunk)
+            await asyncio.to_thread(os.replace, tmp_path, cache_path)
+        except BaseException:
+            with contextlib.suppress(OSError):
+                os.unlink(tmp_path)
+            raise
+    else:
+        data = await obstore_read_range(
```
NickGeneva, line 1229: If theres a byte length, like reading a grib file, this means streaming bytes is just skipped?

earth2studio/data/utils.py
```diff
@@ -1038,10 +1081,30 @@ async def obstore_fetch_to_cache(
...
+        # are never buffered fully in memory, then publish atomically
+        try:
+            resp = await store.get_async(key)
+        except (FileNotFoundError, obs.exceptions.NotFoundError):
+            raise FileNotFoundError(f"Object {key} not found in store") from None
+        fd, tmp_path = await asyncio.to_thread(
+            tempfile.mkstemp, dir=cache_dir, suffix=".tmp"
+        )
+        try:
+            with os.fdopen(fd, "wb") as file:
```
NickGeneva: Yeah, in general just seems kinda odd to me to be doing all the time / by default.

earth2studio/lexicon/planetary_computer.py
```diff
@@ -149,15 +149,19 @@ def get_item(cls, val: str) -> tuple[str, Modifier]:
...
-    For available variables, inspect one of the index files:
+    For available variables, inspect the GRIB index sidecar exposed as the
+    ``index`` asset on any item of the ``ecmwf-forecast`` collection (one
+    JSON-lines entry per GRIB message):
 
     .. highlight:: python
     .. code-block:: python
 
-        import planetary_computer
+        from pystac_client import Client
```
NickGeneva: package change?

pyproject.toml
```diff
@@ -65,11 +65,14 @@ data = [
     "gribberish>=1.5.0; python_version>='3.12'",
     "icechunk>=2.0.0; python_version>='3.12'",
     "intake-esgf>=2026.1.26; python_version>='3.12'",
-    "planetary-computer>=0.5.0",
     "pcodec>=1.0.2; python_version>='3.12'",
     "pybufrkit>=0.2.22",
     "pystac-client>=0.9.0",
-    "httpx==0.28.1",
+    # aiohttp + aiohttp-retry are used by obstore's
```
NickGeneva, line 71: nit: remove these comments, verbose

earth2studio/data/planetary_computer.py
```diff
@@ -437,52 +447,110 @@ def _prepare_asset_plans(
...
-                temp_path.replace(plan.local_path)
+                await async_retry(
+                    self._fetch_asset,
+                    store,
+                    plan,
+                    retries=self._max_retries,
+                    backoff=1.0,
+                    # Bound each attempt so a stalled transfer cannot consume
+                    # the whole async_timeout budget without retrying
+                    task_timeout=300.0,
```
NickGeneva, line 514: Could this this be the self.async timeout?

NickGeneva (commented):
Took a look. The plantary computer updates look good to me (not familiar with azure blobs myself)

I have some concerns with the obs store util changes. Bit worried the obs store cache changes are too focused on attempting to optimize planetary computer data source rendering it overly complex / potential regression for other data sources

Not sure

negin513:
Fair point — I've made the atomic cache write opt-in instead. Default behavior for all the other obstore sources is exactly what they had before; only the planetary computer sources use it, since the ranged IFS fetches can actually race on the same cache file across processes. I also looked at how similar tools handle this — ecmwf-opendata and Herbie both do the same index-based byte-range fetching for this data, and neither hardens their cache writes, so keeping this narrow to the one place with a real race felt like the right call.

Also rebased on main.

negin513:
For reference: the byte-range approach here is essentially the same as what ECMWF's own ecmwf-opendata client does for this dataset — same .index sidecar, same per-message offset/length ranged downloads. Should be ready to go.

negin513:
@NickGeneva this should be ready for another look when you get a chance — the cache changes are now opt-in and scoped to just the planetary computer sources, everything else keeps its old write path. Rebased on main, CI running on the latest commit.

uv.lock
```diff
@@ -1607,63 +1619,51 @@ wheels = [
 
 [[package]]
 name = "cuda-toolkit"
-version = "13.0.3.0"
+version = "13.0.2"
```
NickGeneva: Just as a heads up you're down grading the torch version.

Its a bit annoying, earth2grid will basically pull it down if you have it on the machine. Copy the current one on main and attempt `uv lock`

If it keeps pulling down the version, you might need to delete your uv cache because of earth2grid build keeps forcing the version...  😞 (Peter might of found a cleaner way).

NickGeneva (approved):
LGTM, one additional comment wit the lock file to make sure its nots lowering the torch version, im adding a step in the CI to check for this but its not implemented yet

NickGeneva:
```
test/data/test_planetary_computer.py::test_planetary_computer_parse_blob_href PASSED [ 77%]
test/data/test_planetary_computer.py::test_planetary_computer_local_asset_path_discriminator PASSED [ 77%]
test/data/test_planetary_computer.py::test_planetary_computer_ifs_index_entry_matches PASSED [ 77%]
test/data/test_planetary_computer.py::test_planetary_computer_ifs_ranged_benchmark XPASS [ 77%]
test/data/test_planetary_computer.py::test_planetary_computer_base_extract_not_implemented PASSED [ 78%]
test/data/test_planetary_computer.py::test_planetary_computer_base_invalid_spatial_dims PASSED [ 78%]
test/data/test_planetary_computer.py::test_planetary_computer_base_init PASSED [ 78%]
test/data/test_planetary_computer.py::test_planetary_computer_goes_exceptions[goes16-C-valid_time0-invalid_times0] PASSED [ 78%]
test/data/test_planetary_computer.py::test_planetary_computer_modis_fire_cache[False] XPASS [ 78%]
test/data/test_planetary_computer.py::test_planetary_computer_oisst_cache[False] XPASS [ 78%]
test/data/test_planetary_computer.py::test_planetary_computer_oisst_cache[True] XPASS [ 78%]
test/data/test_planetary_computer.py::test_planetary_computer_oisst_fetch[time1-sst] XPASS [ 78%]
test/data/test_planetary_computer.py::test_planetary_computer_modis_fire_cache[True] XPASS [ 78%]
test/data/test_planetary_computer.py::test_planetary_computer_modis_fire_fetch[time0-variable0-h35v10] XPASS [ 78%]
test/data/test_planetary_computer.py::test_planetary_computer_oisst_fetch[time0-variable0] XPASS [ 79%]
test/data/test_planetary_computer.py::test_planetary_computer_goes_exceptions[goes19-F-valid_time1-invalid_times1] PASSED [ 79%]
test/data/test_planetary_computer.py::test_planetary_computer_modis_fire_fetch[time1-fmask-h34v10] XPASS [ 79%]
test/data/test_planetary_computer.py::test_planetary_computer_sentinel3_cache[False] XPASS [ 79%]
test/data/test_planetary_computer.py::test_planetary_computer_sentinel3_cache[True] XPASS [ 79%]
test/data/test_planetary_computer.py::test_planetary_computer_sentinel3_fetch[time1-s3sy02aod] XPASS [ 79%]
test/data/test_planetary_computer.py::test_planetary_computer_goes_fetch[goes16-C-time0-variable0] XPASS [ 79%]
test/data/test_planetary_computer.py::test_planetary_computer_sentinel3_fetch[time0-variable0] XPASS [ 79%]
test/data/test_planetary_computer.py::test_planetary_computer_goes_fetch[goes19-C-time1-variable1] XPASS [ 79%]
test/data/test_planetary_computer.py::test_planetary_computer_ifs_fetch[time1-t2m] XPASS [ 79%]
test/data/test_planetary_computer.py::test_planetary_computer_ifs_fetch[time0-variable0] XPASS [ 79%]
```

PR #1077, opened 2026-08-18
Online scoring in model eval recipe

pzharrington:
# Earth2Studio Pull Request

## Description
Adds an **online scoring** mode to `recipes/eval`: forecasts are reduced to
sufficient statistics inside the inference loop instead of being written to
`forecast.zarr` and re-read by `score.py`. At campaign scale the raw store
is the binding constraint — a representative 50-member, 730-IC, 14-day
global campaign is ~170 TB of raw fields to produce ~20 MB of scores.
Online scoring collapses that to a ~400 MB `stats.zarr`, converting an
O(100 TB) problem into something much more tractable.

```bash
torchrun --nproc_per_node=$NGPU --standalone main.py \
    campaign=fcn3_2024_monthly \
    scoring.mode=online output.retain=none
```

Validated across all six pipelines in the recipe (`ForecastPipeline`,
`AssimilationForecastPipeline`, `AssimilationPipeline`,
`DiagnosticPipeline`, `StormScopePipeline`, `DLESyMPipeline`).

## What's new

- **`src/online.py`** (new) — group-aware process groups, per-step
  reduction kernels, mergeable accumulators (per-member MSE, ensemble-mean
  MSE, spread/SSR, fair CRPS via member exchange, rank histogram,
  bias/correlation, ACC), `stats.zarr` I/O, and a cheap single-process
  `finalize_stats` step that derives a `scores.zarr` byte-compatible with
  the existing offline schema.
- **Ensemble groups** (`src/work.py`) — ranks are partitioned into groups
  that jointly own one IC at a time; rank `g` in a group carries members
  `[g*K, (g+1)*K)`. With the default `members_per_rank=1`, per-member
  seeding is untouched and online runs reproduce offline forecasts
  bit-for-bit. `members_per_rank > 1` (`K > 1`) batches members onto the
  model's batch axis via a new `Pipeline.run_item_batched`, shrinking a
  group to fit inside one node (NVLink instead of inter-node fabric for
  the CRPS exchange).
- **`Pipeline` contract additions** (`src/pipelines/base.py`) —
  `supports_online_scoring` (opt-in, default `False`), `run_item_batched`,
  and `seed_member` / `stochastic_components()` to unify the four
  previously-divergent per-pipeline seeding mechanisms (`set_rng`,
  sample-indexed generative models, global `manual_seed`) into one path.
- **Per-pipeline enablement**: `ForecastPipeline`, `AssimilationPipeline`,
  `AssimilationForecastPipeline`, `DiagnosticPipeline`, `StormScopePipeline`,
  and `DLESyMPipeline` all set `supports_online_scoring = True`. Generative
  diagnostics (e.g. CorrDiff-shaped models) get their own `sample` axis
  normalized to `ensemble` inside the pipeline rather than in shared
  structural-dim handling. `DLESyMPipeline` gets mask-aware accumulation so
  NaN-masked ocean gridpoints (outside their validity cadence) are excluded
  from every weighted sum instead of silently polluting the rank histogram
  or being scored as zero.
- **`score.py`** — in online mode, scoring is just the cheap finalize step
  (`stats.zarr` -> `scores.zarr`); an online run can still be cross-checked
  against the offline path when `output.retain=all` by overriding
  `scoring.mode=offline` with a distinct `store_name`.
- **Resume / failure handling** — durability unit is `(IC, group)` with its
  own marker namespace (`.online_progress`); a dead rank fails its whole
  group cleanly via a configurable NCCL timeout rather than hanging.
- **Config surface** — `scoring.mode: online`, `output.retain: none|all`,
  and `scoring.online.*` (`ensemble_group_size`, `members_per_rank`,
  `pairwise_comm_dtype`, `defer_pairwise_one_step`, `variable_chunk`,
  `pairwise_variables`, `stats_store`, `scores_store`, `climatology`,
  `nccl_timeout_s`).

## Important design notes

- All group reductions run on **residuals** (`d = f - y`), not raw fields —
  algebraically identical (verification is already local, ensemble variance
  is shift-invariant) but keeps wire values at forecast-error magnitude
  instead of field magnitude, which avoids catastrophic cancellation in
  moment sums and makes `bfloat16` safe for the pairwise CRPS exchange.
- The CRPS pairwise term (`E|X-X'|`) is the only genuinely expensive piece
  of communication (~25x everything else); it's mitigated via one-step
  deferral (overlaps the model forward), `members_per_rank` (moves the
  exchange onto NVLink), configurable comm dtype, and an optional variable
  subset.
- The pairwise exchange is a chunked all-gather rather than a ring (an
  earlier `pairwise_exchange: ring` mode shipped briefly and was removed —
  a manual ring can't be deferred and cost more to maintain than it bought).
- `scoring.mode: both` and `pairwise_exchange` overrides now raise a
  migration error rather than silently falling back, since a silent
  fallback to offline would write the raw store the campaign was sized to
  avoid.

## Testing

- `recipes/eval/test/test_online.py` — accumulator correctness, grouping
  invariance (`M` fixed, `G` in `{1, 2, 4}`), mask-aware accumulation,
  fair CRPS, member batching, config validation.
- `recipes/eval/test/test_diagnostic_inference.py`,
  `recipes/eval/test/test_multigpu.py`,
  `recipes/eval/test/_multigpu_worker.py` — multi-GPU `torchrun`-subprocess
  validation and generative-diagnostic sample-axis handling.
- `recipes/eval/README.md` has a full "Online scoring" section with
  configuration reference, the CRPS cost/mitigation table, and validation
  instructions.

## Checklist


recipes/eval/src/pipelines/forecast.py
```diff
@@ -364,14 +582,42 @@ def run_item(
...
-            dx_ic = self._dx_input_coords[id(dx)]
-            x_in, coords_in = map_coords(x, coords, dx_ic)
-            y, y_coords = dx(x_in, coords_in)
-            x_combined, coords_combined = cat_coords(
-                (x_combined, y), (coords_combined, y_coords), "variable"
+        yield self._run_diagnostics(x, coords, np.array([item.ensemble_id]))
+
+    def run_item_batched(
+        self,
+        items: list[WorkItem],
```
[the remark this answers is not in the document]
pzharrington: Fixed

pzharrington:
@greptileai

pzharrington:
@greptileai

recipes/eval/src/pipelines/diagnostic.py
pzharrington, line 1: Just migrated from the `forecast.py` file for clarity, otherwise changes are minimal and mostly resemble the updates in the assimilation pipeline supporting online scoring, `run_item_batched`, etc.

recipes/eval/README.md
```diff
@@ -675,6 +676,171 @@ per-(time, lead_time) RMSE values.  To compute aggregate RMSE across
...
+
+Members meet via a single all-gather per variable chunk, which is what
+makes the exchange deferrable; `variable_chunk` bounds both the payload
+and the float64 working set.
+
+**Re-deriving scores.**  `stats.zarr` holds *sums*, not scores, so regional
+breakdowns, or bootstrap CIs over ICs never require re-running inference.
+However, new metrics or more extensive changes will require re-running the
+inference portion, which is the main downside of online scoring. To
+consolidate the summation results in `stats.zarr` into the final scores,
```
Dibyajyoti-Chakraborty: "Regional breakdowns" here is slightly confusing. The next line says that it needs re-running for new-metrics. But even a new regional eval for some metrics will need a rerun right ?

recipes/eval/src/online.py
```diff
@@ -0,0 +1,2554 @@
...
+weighted rank counts, anomaly cross-moments — from which
+:func:`finalize_stats` derives a ``scores.zarr`` that is schema-compatible
+with the offline scorer's output (``{metric}__{variable}`` arrays).  Two
+properties follow:
+
+* Non-linear aggregations stay correct.  ``RMSE = sqrt(mean_t(MSE))`` and
+  ACC are not linear in ICs, so the *sums* rather than the per-IC scores
+  are the durable artifact.
+* Derived metrics, regional breakdowns and bootstrap CIs over ICs remain
+  possible after the fact without re-running inference.
```
Dibyajyoti-Chakraborty: Can a different regional metric be computed post-hoc?

recipes/eval/src/online.py
```diff
@@ -0,0 +1,2554 @@
...
+class WeightSum:
+    """The weighted-mean normalizer ``W = sum_s w``.
+
+    Stored per ``(IC, lead, variable)`` — rather than assumed constant —
+    precisely so a masked / regional accumulation can vary it without
+    changing the store schema or the finalizers: a fully NaN-masked
+    (lead, variable) gets ``W = 0`` here, matching the zeroed numerator
+    every other statistic's :meth:`StepContext.wsum` produces for it, so
+    ``finalize_stats`` divides ``0 / 0 = NaN`` rather than a wrong,
+    finite value against the unmasked grid weight.
```
Dibyajyoti-Chakraborty, line 942: This can not be changed once computed, right? Which is fine; just contradicts some lines in the readme about regional breakdowns not needing rerun.

recipes/eval/README.md
```diff
@@ -675,6 +676,171 @@ per-(time, lead_time) RMSE values.  To compute aggregate RMSE across
...
+
+**Failure domain.**  A dead rank hangs its whole group rather than merely
+forfeiting its own items.  `scoring.online.nccl_timeout_s` (default 30 min)
+turns that into a clean job failure that `resume=true` picks up from.
+
+**Caveats.**
+
+- `verification.zarr` must be predownloaded; coverage of every
+  `(IC + lead)` valid time is checked before model weights load.
+- Online scoring priortizes support for a fixed set of important metrics
```
Dibyajyoti-Chakraborty: Typo.
priortizes -> prioritizes

Dibyajyoti-Chakraborty (commented):
Overall LGTM. Some minor comments mentioned below.

recipes/eval/README.md
pzharrington: Good point yeah. Regional breakdowns was something I wanted to leave the door open to, but as of now there's now regional config for online scoring. Spatial weighting (currently just`lat_weights`, on/off globally) is baked into the sums during the run and the per-gridpoint data is gone afterward. I think we should be able to extend this to spatial sums/regional breakdowns without too much trouble in a follow-up PR. So you're right, a different/new region always needs a rerun (same as a genuinely new metric) for now, and only metrics derivable from the already-stored sums (e.g. rank reliability from stored rank counts) and bootstrap CIs over ICs are free. Clarified the wording on it here and in the `online.py` docstring.

recipes/eval/src/online.py
pzharrington: Yeah, good catch

recipes/eval/src/online.py
pzharrington: Not as of now....follow-up PR ;)

Dibyajyoti-Chakraborty (approved):
LGTM

PR #1079, opened 2026-08-20
Scorecard for E2S models

Dibyajyoti-Chakraborty:
Adds **Scorecards** to the documentation: one page per model (FCN3, Aurora, to start) showing
its forecast skill on a common campaign — 24 ERA5 initial conditions of 2025, 14-day horizon —
with interactive lead-time plots for RMSE, MAE, CRPS, spread, and log spectral distance, plus
variable tables, citations, etc. Pages are generated at docs build time (`docs/generate_scorecard.py`)
 from minified JSON score exports produced by the new
`recipes/eval/scorecard/` — self-contained campaign configs and a driver/exporter that reuse the
eval recipe's own inference, scoring, and metrics end to end.

Example build - https://dibyajyoti-chakraborty.github.io/earth2studio/scorecard/generated/

## Checklist


recipes/eval/scorecard/cfg/campaign/fcn3_2025_scorecard.yaml
```diff
@@ -0,0 +1,295 @@
...
+#   ./run_scorecard.sh fcn3_2025_scorecard
+# ---------------------------------------------------------------------------
+defaults:
+    - override /model: null   # self-contained: the model is defined below
+
+project: fcn3
+run_id: 2025_scorecard
+
+# NGC rather than the HuggingFace default: anonymous HF pulls from the
+# cluster's shared egress IP hit a 429 rate limit. Same Apache-2.0 checkpoint.
```
pzharrington, line 36: Mild preference for HF instead of NGC just for stability -- time for you to make a HF account ;)

recipes/eval/scorecard/utils/pipelines.py
```diff
@@ -0,0 +1,334 @@
...
+    every lead is written. A history model does not: Aurora, with input
+    ``lead_time = [-6h, 0h]``, yields the t-6h history frame and then jumps
+    to +6h, so lead 0 is never written and would score as NaN.
+
+    This pipeline restores the expected behaviour: the initial condition is
+    written at lead 0 (that is what ``forecast[lead=0]`` is by definition),
+    yields at negative lead times are dropped, and a model that emits its own
+    lead 0 is not written twice. Keeping lead 0 real is what makes the
+    scorecard's sanity check -- error exactly zero at lead 0 -- meaningful.
+    """
```
pzharrington: At first I was surprised at the need for this, but some digging revealed I think that it's just a bug with Aurora. The general convention is prognostic models with history should return the IC from the first yield in `create_iterator`, and it's up to the model wrapper to strip off the history and return a clean IC. DLWP, FengWu, and FuXi all do this, for example.

`Aurora` has a bug in that `_default_generator` slices the tensor (`x[:, :, 1:]`) but yields coords unmodified, so the first yield has a tensor with lead-dim size 1 paired with `coords["lead_time"]` still `[-6h, 0h]` (length 2). So I think the more consistent fix would be patching `Aurora._default_generator` to slice `coords["lead_time"]` the same way `aurora1p5.py` and DLWP/FengWu/FuXi do, rather than special-casing it in the eval recipe. Then this whole abstraction can be removed

Dibyajyoti-Chakraborty: Ya, I faced it last year too in the MoWE training. Let me try to do that fix first.

recipes/eval/scorecard/run_scorecard.py
pzharrington, line 1: There are some things in here that are a bit cluster-/env-specific:
 - `if socket.gethostname().startswith("login")`
 - `os.environ.setdefault("EARTH2STUDIO_CACHE", "/opt/venv/.e2s_cache")`
 - etc. 
 
I suppose we have a bit of a decision to make on how general we want this to be vs. something that we just know how to use ourselves as the maintainers. If the latter, I think it should be marked as non-public (prepend a `_` to the filename) and that should be noted in the README.

pzharrington: It's also forced to be single-node (`--standalone` in the torchrun) which we may need to expand beyond in the future. When the campaigns become larger-scale, it may need to be more async/incremental -- the inference jobs could take days to get through queue and complete

pzharrington: The nuclear option: we maintain our own isolated runscripts/automation/things like this in a private repo designed to work with a checkout of e2studio; then this could be removed entirely. In general running these should be able to be done easily by any of the E2 team members on internal clusters, but we should try not to commit things related to internal clusters publicly. @NickGeneva any thoughts?

docs/_static/scorecard/eval_scores_fcn3.json
pzharrington, line 1: Thinking ahead here...these are already approaching the 250K max file size limitation in pre-commit, basically these are the biggest files in the repo besides the lockfile. Every new export of the json will fully rewrite it so every re-score is another 100-200K blob added to git history, and if/when we add regional/temporal breakdowns that will already push us over the 250K probably.

Not saying this should block the current PR merging but we probably should more seriously consider either externally hosting these assets (e.g. in @NickGeneva 's E2Studio Assets repo on HF -- this would open up the pathways to fancier visualization plots in the scorecard too) or some other leaner approach here (maybe a different file format? combining json and npy?)

recipes/eval/scorecard/run_scorecard.py
Dibyajyoti-Chakraborty: Updated!

recipes/eval/scorecard/utils/pipelines.py
Dibyajyoti-Chakraborty: Updated!

recipes/eval/scorecard/cfg/campaign/fcn3_2025_scorecard.yaml
Dibyajyoti-Chakraborty: Let's keep this one NGC for now ;) I'm having some issues with HF. I'll try it with the next models.

pzharrington (approved):
Some items to address in follow-ups but this is great for now imo!

recipes/eval/scorecard/run_scorecard.py
NickGeneva: +1 thanks for updating, all for having in house utility repo for orchestration on our systems to keep public side clean.

docs/_static/scorecard/eval_scores_fcn3.json
NickGeneva: Agreed! HF is a cleaner spot most likely, but this works for now 👍

NickGeneva (approved):
Over all LGTM!

PR #1086, opened 2026-08-21
Fix SamudrACE run-to-run non-determinism from per-cycle cudnn.benchmark toggling

negin513:
## Description

`SamudrACE` inference was not reproducible run-to-run: two runs of the identical script, checkpoint, initial condition, and forcing produced bit-identical output through the first coupled (5 day) cycle, then diverged starting at the first ocean-coupling boundary (~0.01-0.02 K), chaotically amplifying to full decorrelation (~30-40 K max t2m diff) by ~day 25.

**Root cause:** `_cudnn_benchmark()` toggled `torch.backends.cudnn.benchmark` on and off around every coupled cycle. Each toggle re-triggers cuDNN's GPU-timing-based algorithm autotuning search, which is subject to real timing noise and can select a different (numerically non-identical) algorithm on different runs. The docstring claimed this matched fme's inference entrypoints, but fme sets the flag once, globally, and never restores it — the reference `fme.coupled.inference` CLI is perfectly bit-reproducible run-to-run on the same hardware.

**Fix:** remove the per-cycle toggle and leave the flag at its default. The two tests asserting the old toggling contract are removed.

## Verification (GB200, 24- and 72-cycle forecasts from the published `allenai/SamudrACE-CM4-piControl` checkpoint)

- **Before:** two identical runs diverge from each other starting at atmosphere step 21 (max t2m diff 0.021 K at 126 h, ~34 K by 600 h).
- **After:** two independent runs are bit-identical (`np.array_equal`, full 180x360 fields, every step) over both 120-day (481 steps) and 360-day (1441 steps) forecasts.
- **Parity with reference:** output is bit-for-bit identical to the raw `fme.coupled.inference` CLI run on the same checkpoint/IC/forcing when the CLI also runs without benchmark autotuning — all 1440 atmosphere steps and 72 ocean cycles of a 360-day run, including derived diagnostics (global-mean diurnal cycle statistics match exactly). Setting `benchmark=True` once globally on both sides does **not** achieve cross-process parity (the autotune is a per-process timing lottery), confirming heuristic (default) algorithm selection is the right choice for reproducibility.
- **Performance:** ~35% faster end-to-end (~22 s vs ~35 s per 120-day forecast) since the repeated per-cycle autotuning overhead is gone.

## Checklist

- [x] The CHANGELOG.md is up to date with these changes.


negin513:
Sanity figures for the fix: the same 120-day (24 coupled cycles) forecast from the CM4 piControl 0311-01-01 initial condition, run through both paths with the patched code. The output fields are bit-for-bit identical (`np.array_equal` over every t2m step and sst cycle), so the two figures below are pixel-identical — posting both for completeness.

**Earth2Studio `SamudrACE` (this PR):**

e2s SamudrACE sanity contours

**Reference `fme.coupled.inference` CLI (same checkpoint/IC/forcing, benchmark autotuning off):**

fme CLI sanity contours

Physical checks visible in both: correct seasonal NH warming for a Jan 1 start, the diurnal cycle in global-mean t2m (period exactly 1.00 day, ~0.29 K std, warmest at 06Z — driven by the `DSWRFtoa` forcing and matching the reference exactly), and global-mean SST stepping once per 5-day coupled cycle as designed.

negin513:
**Extended validation: out-of-the-box fme reference config, full duration**

To rule out anything specific to a hand-built reference config, the comparison was repeated against the inference configuration **shipped in the HuggingFace repo itself**, run at its full published length:

```bash
hf download allenai/SamudrACE-CM4-piControl --local-dir SamudrACE-CM4-piControl
cd SamudrACE-CM4-piControl
python -m fme.coupled.inference inference_config.yaml   # as shipped: n_coupled_steps=140 (700 days),
                                                        # coupled_steps_in_memory=20, forcing n_repeats=3,
                                                        # IC 0311-01-01, relative paths in the repo dir
```

The only deviation for the bit-for-bit test: `torch.backends.cudnn.benchmark` pinned `False` in the fme process (fme hard-codes it `True` with no config option; benchmark autotuning is a per-process timing lottery, so no two processes — including two stock fme runs vs each other's algorithm choices, or stock fme vs anything — can be expected to match bitwise under it).

The Earth2Studio side ran this PR's `SamudrACE` for 84 coupled cycles (420 days, 1680 atmosphere steps) — the maximum from this IC, since the forcing lookup rejects proleptic-Gregorian 0312-02-29 (~day 424) on the no-leap forcing calendar by design.

**Result over the full 420-day overlap window:**

| Field | Extent compared | `np.array_equal` (full 180×360 arrays, `equal_nan=True`) |
|---|---|---|
| `t2m` / `TMP2m` | all 1680 atmosphere steps | **bit-identical** |
| `sst` | all 84 ocean cycles | **bit-identical** |

Notably this spans the **forcing year boundary (day 365)**: fme extends past one year via `n_repeats: 3` (verbatim annual tiling) while `SamudrACEForcingData` tiles by month/day/hour matching — bfb holding straight through the boundary confirms the two tiling semantics are exactly equivalent, not merely close.

Timing on GB200: shipped 700-day stock run 107.5 s (26.1 steps/s); benchmark-off 700-day run 141.9 s (19.9 steps/s); e2s 420-day run ~90 s end-to-end. For reference, the stock (benchmark=True) run vs e2s shows the expected autotune-lottery divergence: exact through step 20, max 0.0138 K at step 21, ~34 K by step 101.

negin513:
@pzharrington mind taking a look at this one when you get a chance? Small diff but it fixes SamudrACE giving different results on every run; turned out to be the cudnn benchmark toggling around each coupled cycle. 

Details and the bit-for-bit checks against the fme CLI are in the comments above.

CHANGELOG.md
```diff
@@ -84,6 +84,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed `SamudrACE` inference being non-reproducible run-to-run: toggling
+  `torch.backends.cudnn.benchmark` on and off around each coupled cycle
+  re-triggered cuDNN's GPU-timing-based algorithm search every cycle
```
pzharrington, line 89: Arguably not necessary since we just added it, up to you if you want to keep it

pzharrington (approved):

negin513:
/ok to test [0faab5a](https://github.com/NVIDIA/earth2studio/pull/1086/commits/0faab5ad23e894d34ee165515ff9e30f7da569e6)

PR #1088, opened 2026-08-24
CI: verify test container Torch version

NickGeneva:
## Summary

- capture the installed Torch version while warming the test image uv cache
- expose the captured version to test-job shells without inspecting the cache at runtime
- fail immediately when the container Torch version differs from `uv.lock`

## Validation

- `uv run pre-commit run --files .github/actions/run-ci-test/action.yml test/Dockerfile`
- verified the guard passes for Torch 2.13.0
- verified an intentional 0.0.0 mismatch fails before environment setup or tests


test/_ci/check_torch_version.sh
```diff
@@ -0,0 +1,56 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -Eeuo pipefail
+
+if [[ -z "${TORCH_VERSION:-}" ]]; then
```
NickGeneva, line 20: if TORCH_VERSION not set in container, just log a warning

test/_ci/check_torch_version.sh
```diff
@@ -0,0 +1,56 @@
...
+    fi
+    exit 0
+fi
+
+if [[ -z "${UV_PYTHON:-}" ]]; then
+    echo "UV_PYTHON is not set by the test container."
+    exit 1
+fi
+
+locked_torch_version="$(
```
NickGeneva, line 35: Pull out the torch version from uv tree command (which will parse the pipelines uv.lock file)

pzharrington (approved):
LGTM 👍

PR #1092, opened 2026-08-24
Fix GFS precipitation lookup for historical data

### LINKED ISSUE #1030: 🐛[BUG]: GFS `tp` returns an all-zero field for dates before 2021-03-23
### Version

0.18.0a0

source

### Describe the issue

# 🐛[BUG]: GFS_FX `tp` returns an all-zero field for supported dates before 2021-03-23

**Version:** 0.18.0a0
**Installation method:** source

---

## Describe the issue

Asking `GFS_FX` for total precipitation (`tp`) at a positive lead time can
quietly give you a field of all zeros for any date from 2021-01-01 to
2021-03-22. Zeros look like no precipitation occurred, so nothing further down
the line notices anything is wrong.

<details>
<summary>How the GRIB index lookup works</summary>

Before downloading anything, the code reads a small text file that lists what is
inside each GRIB file. Each line looks like this:

```
596:418593746:d=2024010100:APCP:surface:0-3 hour acc fcst:
 ^      ^                    ^      ^          ^
 |      |                    |      |          how much time the precipitation accumulates over
 |      |                    |      level
 |      |                    variable
 |      where it starts in the file
 record number
```

The code builds a search string and looks for it in that list. If it finds
nothing, it skips the variable and leaves that part of the output untouched.

</details>

### Problem 1: the search uses a record number, and the number moved

The lexicon looks up `tp` by its record number:

```python
# earth2studio/lexicon/gfs.py L51
"tp": "596::APCP::surface",  # 3 hour acc
```

This is the only one of 246 entries that does this. Every other entry searches
by variable name and level.

Record 596 is right for files from 2021-03-23 onwards. In older files the
corresponding accumulation record is **450**, so the search finds nothing and
you get zeros.

Those older dates are supported on purpose. `GFS` accepts dates back to
2021-01-01, and the code already has a
[special branch](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/data/gfs.py#L484)
for the older directory layout. So the library knows the old files exist, but
looks for them using a record number that only works for the new ones.

Both layouts, side by side:

```console
$ curl -s https://noaa-gfs-bdp-pds.s3.amazonaws.com/gfs.20210101/00/gfs.t00z.pgrb2.0p25.f003.idx | grep ':APCP:surface:'
450:236579998:d=2021010100:APCP:surface:0-3 hour acc fcst:
451:236880680:d=2021010100:APCP:surface:0-3 hour acc fcst:

$ curl -s https://noaa-gfs-bdp-pds.s3.amazonaws.com/gfs.20240101/00/atmos/gfs.t00z.pgrb2.0p25.f003.idx | grep ':APCP:surface:'
596:418593746:d=2024010100:APCP:surface:0-3 hour acc fcst:
597:418886486:d=2024010100:APCP:surface:0-3 hour acc fcst:
```

Same record descriptions and roles, different record numbers.

### Reproduction

Save as `repro.py`:

```python
import numpy as np
from earth2studio.data import GFS_FX

ds = GFS_FX(cache=False)
for t in [datetime(2021, 1, 1), datetime(2021, 6, 1)]:
    a = np.asarray(ds(t, timedelta(hours=3), ["tp"]).values).squeeze()
    print(f"{t:%Y-%m-%d}  min={a.min():.6g} max={a.max():.6g} allzero={bool((a==0).all())}")
```

```console
$ uv run python repro.py
WARNING | earth2studio.data.gfs:_create_tasks:325 - Variable tp not found in
index file for time 2021-01-01 00:00:00 at 3:00:00, values will be unset
2021-01-01  min=0 max=0        allzero=True
2021-06-01  min=0 max=0.111375 allzero=False
```

The same request five months apart. One of them comes back empty.

<details>
<summary>Why a record number was used, and what the two records mean</summary>

Each forecast file contains two `APCP:surface` records. Record 596 represents
the current six-hour accumulation bucket, while record 597 represents
accumulation since initialization. Their interval labels distinguish them after
f006, but the current lexicon selects record 596 by a layout-dependent record
number:

| lead | record 596 (recent window) | record 597 (since the start) |
|---|---|---|
| f001 | `0-1 hour` | `0-1 hour` |
| f006 | `0-6 hour` | `0-6 hour` |
| f007 | `6-7 hour` | `0-7 hour` |
| f012 | `6-12 hour` | `0-12 hour` |
| f024 | `18-24 hour` | `0-1 day` |
| f120 | `114-120 hour` | `0-5 day` |
| f384 | `378-384 hour` | `0-16 day` |

From f007 onwards the two have different time windows. That window text is the
same in old and new files, which makes it a better thing to search for than the
record number.

</details>

### Why the tests do not catch it

No test requests `GFS_FX.tp` at a positive lead time using the older layout.
The existing
[`2021-01-01 tp test`](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/test/data/test_gfs.py#L29-L46)
uses `GFS`, which reads f000 and therefore exercises a different case.

## Expected behaviour

`GFS_FX.tp` should resolve for every supported date at valid positive lead
times.

## Suggested fix

Stop searching by record number. Search by the time window instead, because that
text is the same in old and new files.

For `h >= 1`, the window follows a simple pattern: it restarts every 6 hours.

```
start = 6 * ((h - 1) // 6)
key   = f"APCP::surface::{start}-{h} hour acc fcst"
```

For example lead 7 gives `6-7 hour acc fcst`, and lead 120 gives
`114-120 hour acc fcst`.

<details>
<summary>What was actually checked against live index files</summary>

- 14 lead times (f001, f002, f003, f004, f005, f006, f007, f009, f012, f024,
  f120, f240, f384) at **00z on 2024-01-01**, new layout. f121 returns 404,
  because GFS runs hourly up to f120 and 3-hourly after that.
- The **00z, 06z, 12z and 18z** runs, but only at **f009 and f015**, on
  2024-01-01. This is what shows the 6-hour pattern follows the lead time and
  not which run of the day it is.
- The old layout on **2021-01-01** (f000, f003, f006, f009, f012, f024) and on
  **2021-02-15 and 2021-03-01** (f003 and f009 only), all at 00z.

Not every combination was tested. What this does show is that the window text is
the same in both layouts, while the record number is not.

</details>

<details>
<summary>Implementation notes: HRRR's rule does not transfer, and the duplicate records</summary>

**1. The HRRR rule does not work here.** HRRR restarts its accumulation every
hour, GFS every 6 hours. Copying the HRRR code would give the wrong window at
every lead except 1, 7, 13 and so on.

**2. At leads 1 to 6 both records show the same window text**, so the search
matches both and the first one wins. That is safe at the leads checked, but
worth a code comment so nobody later assumes it is an accident.

At the two leads compared, both on 2024-01-01 00z, the duplicate records are
byte for byte identical, so it makes no difference which one is picked:

| lead (2024-01-01 00z) | record 596 | record 597 |
|---|---|---|
| f003 | `6cad64b0d2967761d899e4683496c470` | same |
| f006 | `8a1cc394334afec57d4271cf4fbe6b5c` | same |

Leads 1, 2, 4 and 5 were not compared, so treat those as untested rather than
confirmed.

</details>

Because `GFSLexicon` is shared by `GFS` and `GFS_FX`, the `tp` entry should stay.
Its value can become a plain `APCP::surface` key that `GFS_FX` completes with
the correct positive-lead window.

## Related

The zero filling that hides all of this is #1028. If those arrays were filled
with NaN instead, the existing "no NaNs" assertions in the tests would have
caught this when the file layout changed.

NickGeneva:
## Summary

TLDR: GFS has a random layout shifts part way through the archive and also changes the index of the tp value

- stop identifying GFS total precipitation by layout-dependent GRIB record number
- retain forecast interval descriptions in parsed index keys
- select the six-hour precipitation accumulation window for each positive lead time
- preserve the existing kg/m�-to-meters precipitation conversion
- add compact offline coverage for historical and current record layouts

http://cpc.ncep.noaa.gov/products/wesley/wgrib2/ncep_norm.html

Fixes #1030

## Testing

- `uv run pytest test/data/test_gfs.py -q` (13 passed, 11 slow tests skipped)
- `make lint`
- live `GFS_FX` f003 fetch for 2021-01-01: max=0.09825 m, allzero=False
- live `GFS_FX` f003 fetch for 2024-01-01: max=0.12325 m, allzero=False


earth2studio/lexicon/gfs.py
```diff
@@ -49,7 +49,7 @@ class GFSLexicon(metaclass=LexiconType):
         "sp": "PRES::surface",
         "msl": "PRMSL::mean sea level",
         "tcwv": "PWAT::entire atmosphere (considered as a single layer)",
-        "tp": "596::APCP::surface",  # 3 hour acc
+        "tp": "APCP::surface",
```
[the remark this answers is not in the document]
NickGeneva: Fixed in 1e12efac. The modifier now keys off the Earth2Studio variable ID (`val == "tp"`), preserving the kg/m^2-to-meters conversion after removing the record-number prefix. The compact regression test also asserts that 1000 kg/m^2 converts to 1 m; live old/new-layout f003 values are now 0.09825 m and 0.12325 m.

NickGeneva:
Manual archive-boundary plotting check (kept uncommitted):

```python
from pathlib import Path

import numpy as np

from earth2studio.data import GFS_FX

DATES = [
    datetime(2021, 3, 22, 6),
    datetime(2021, 3, 22, 12),
]
LEAD_TIME = timedelta(hours=3)
OUTPUT = Path(__file__).with_name("gfs_tp_old_new.jpg")


def main() -> None:
    source = GFS_FX(cache=False, verbose=False)
    fields: list[np.ndarray] = []
    lat = lon = None

    for date in DATES:
        data = source(date, LEAD_TIME, ["tp"])
        fields.append(np.asarray(data.values).squeeze() * 1000.0)
        lat = data.coords["lat"].values
        lon = data.coords["lon"].values

    vmax = max(float(np.nanpercentile(field, 99.5)) for field in fields)
    fig, axes = plt.subplots(1, 2, figsize=(14, 5), constrained_layout=True)

    for axis, date, field in zip(axes, DATES, fields):
        image = axis.pcolormesh(
            lon,
            lat,
            field,
            shading="auto",
            cmap="Blues",
            vmin=0,
            vmax=vmax,
        )
        axis.set(
            title=f"{date:%Y-%m-%d %H}z + 3 h",
            xlabel="Longitude",
            ylabel="Latitude",
        )

    fig.colorbar(image, ax=axes, label="3-hour total precipitation (mm)", extend="max")
    fig.suptitle("GFS total precipitation across the archive layout boundary")
    fig.savefig(OUTPUT, dpi=180, format="jpg", pil_kwargs={"quality": 90})
    plt.close(fig)
    print(f"Saved {OUTPUT}")


if __name__ == "__main__":
    main()
```

Run from the repository root with:

```console
uv run --group docs python plot_gfs_tp_old_new.py
```

This regenerates `gfs_tp_old_new.jpg` on a shared color scale. The local run
successfully fetched and rendered 2021-03-22 06z f003 from the legacy path and
2021-03-22 12z f003 from the `atmos/` path.

gfs_tp_old_new

earth2studio/lexicon/gfs.py
gertln:
```suggestion
        "tp": "APCP::surface", # accumulation over the current 6 h bucket; window length grows 1 h to 6 h then restarts
```

earth2studio/data/gfs.py
gertln: not introduced here but GFS layout cutoff is 12 hours too late, 2021-03-22 12z and 18z fail for all variables

gertln:
```suggestion
        if time < datetime(2021, 3, 22,12):
```

gertln:
```suggestion
        if time < datetime(2021, 3, 22,12):
```

gertln (commented):
The correctness for lead 0 depends on #1095 
f000 contains no APCP:surface record so with the lead_hour > 0 guard it falls through to the missing-record path and will result in all zeros. With #1095 merged it beomes all-nan.

earth2studio/data/gfs.py
NickGeneva: Good catch, fixed.

NickGeneva:
@greptile-apps

earth2studio/lexicon/gfs.py
```diff
@@ -49,7 +49,7 @@ class GFSLexicon(metaclass=LexiconType):
         "sp": "PRES::surface",
         "msl": "PRMSL::mean sea level",
         "tcwv": "PWAT::entire atmosphere (considered as a single layer)",
-        "tp": "596::APCP::surface",  # 3 hour acc
+        "tp": "APCP::surface",  # accumulation over the current 6 h bucket; window length grows 1 h to 6 h then restarts
```
pzharrington, line 52: So this basically accumulates in 6h increments relative to the start of the forecast (where it starts at 0)?

pzharrington (approved):

earth2studio/lexicon/gfs.py
NickGeneva: Yeah my understanding is that basically when you go sub 6 hourly... things get a little wacky because its not exactly a rolling window. Basically the accumulation resets every 6 hours.

http://cpc.ncep.noaa.gov/products/wesley/wgrib2/ncep_norm.html

So technically, I think at 6 hour intervals its actually tp06, ugh kinda confusing. Dont have a good answer for this.

pzharrington: Yeah, annoying but I think `tp` makes sense for the variable name, and behavior can be documented in the data source

PR #1093, opened 2026-08-24
Fix HRRR GRIB index lookups

### LINKED ISSUE #1029: 🐛[BUG]: HRRR GRIB index lookup misses variables
### Version

0.18.0a0

source

### Describe the issue

Five HRRR variables fail to resolve against the GRIB index.
No fetch task is created, that variable's region of the pre-allocated output
array is never written, and the caller receives it without an exception.

| Class | Variable | Affected lead times |
|---|---|---|
| `HRRR_FX` | `sd`, `sde`, `snowc`, `tcc` | every lead > 0 |
| `HRRR_FX` | `tp` | lead 0 |
| `HRRR` | `tp` | all times |

**The data is present in the file.** These are lookup-key defects, not missing
records.

### Deterministic evidence

The records exist, indexed under `1 hour fcst`:

```console
$ curl -s https://noaa-hrrr-bdp-pds.s3.amazonaws.com/hrrr.20260720/conus/hrrr.t00z.wrfsfcf01.grib2.idx \
    | grep -E ':(WEASD|SNOWC|SNOD):surface:1 hour fcst:|:TCDC:entire atmosphere:1 hour fcst:'
68:44839760:d=2026072000:WEASD:surface:1 hour fcst:
69:44842125:d=2026072000:SNOWC:surface:1 hour fcst:
70:44843752:d=2026072000:SNOD:surface:1 hour fcst:
116:82872841:d=2026072000:TCDC:entire atmosphere:1 hour fcst:
```

But [`HRRRFXLexicon` L203-L206](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/lexicon/hrrr.py#L203-L206)
pins them to an explicit `::anl` forecast-validity token.

This is correct only for `f00`. Positive-lead files label these records as
`<lead> hour fcst`, so the hard-coded `anl` token prevents any match:

```python
"tcc": "wrfsfc::TCDC::entire atmosphere::anl",
"sd": "wrfsfc::WEASD::surface::anl",      # will be zero
"sde": "wrfsfc::SNOD::surface::anl",      # will be zero
"snowc": "wrfsfc::SNOWC::surface::anl",   # will be zero
```

`anl` appears **only** in the `f00` file, `grep -c ':anl:'` on `wrfsfcf01.idx`,
`wrfprsf01.idx` and `wrfnatf01.idx` returns `0`. So the lookup misses at every
forecast lead, and
[`_create_tasks` L423-L426](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/data/hrrr.py#L423-L426)
skips the variable:

```console
$ uv run python -c "
from earth2studio.data import HRRR_FX
HRRR_FX(cache=False)(datetime(2026,7,20), timedelta(hours=1), ['sd'])"

WARNING | earth2studio.data.hrrr:_create_tasks:424 - Variable sd not found in
index file for time 2026-07-20 00:00:00 at 1:00:00, values will be unset
```

This warning confirms that the lookup found no matching index entry. The code
then skips creating a download task, leaving the requested variable's output
array unwritten.

The other 643 of 648 entries in `HRRRFXLexicon` carry no forecast-validity
token and are completed from the lead time by the generic branch, which produces
exactly the `anl` / `<lead> hour fcst` strings that match. These four (plus
`tp`, below) are the only ones that do not.

### `tp` at lead 0 looks for `-1-0 hour` instead of the available `0-0 day` record

[`data/hrrr.py` L406](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/data/hrrr.py#L406):

```python
hrrr_key = f"{variable_name}::{level}::{hours-1:d}-{hours:d} hour acc fcst"
```

At lead 0 this yields `APCP::surface::-1-0 hour acc fcst`, which matches no
record. The `f00` file records the zero-length accumulation as
`APCP:surface:0-0 day acc fcst` (note: **day**, not hour).

This line also overwrites the lexicon's own token unconditionally, so the
analysis class `HRRR` is affected at every date: its lexicon entry
`wrfsfc::APCP::surface::0-0 hour acc` is dead text and never used.

```console
$ uv run python -c "
import numpy as np
from datetime import datetime
from earth2studio.data import HRRR
a = np.asarray(HRRR(cache=False)(datetime(2026,7,20), ['tp']).values)
print(a.min(), a.max())"

WARNING | earth2studio.data.hrrr:_create_tasks:424 - Variable tp not found in
index file for time 2026-07-20 00:00:00 at 0:00:00, values will be unset
0.0 0.0
```

### Observed consequence

Because `HRRR_FX` never writes the requested variable into its output array,
the returned values depend on whatever data is already present in the allocated
memory.
`HRRR_FX` allocates with
[`np.empty`](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/data/hrrr.py#L840),
so its contents are whatever was previously at that address. The values are not
defined and do not reproduce. Two identical calls in a single process returned
different arrays; across runs we have observed means of `2.5e270`, `-97.5`,
and exact zeros, none of them NaN, so nothing downstream flags them:

```
call 1: min=0      max=4.826e+276  mean=2.533e+270  nan=False
call 2: min=-134.1 max=6.932e-310  mean=-97.51      nan=False
identical across calls: False
```

`HRRR` allocates with `np.zeros`, so `tp` is deterministically all-zero , 
indistinguishable from "it did not rain".

That fill behaviour is tracked separately in #1028; it is why this bug went
unnoticed, but it is not this bug.

## Expected behaviour

All five variables resolve, in both classes and at every lead time, and return
the records that are present in the file.

## Suggested fix

For `sd`, `sde`, `snowc` and `tcc`, build the index key from the requested lead
time: use `anl` at lead 0 and `<lead> hour fcst` at positive leads. For `tp`,
handle lead 0 explicitly by selecting `APCP::surface::0-0 day acc fcst` instead
of calculating the invalid `-1-0 hour` key.

**Note:** the `HRRRLexicon` (analysis) entries at
[L60-L63](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/lexicon/hrrr.py#L60-L63)
carry the same `::anl` token and are **correct**, that class only ever fetches
lead 0. They should not be changed.

**Also note:** with the lookup fixed, `sd` returns the correct `WEASD` record but
in kg/m², whereas
[`base.py`](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/lexicon/base.py)
specifies metres water equivalent; the ×1000 conversion is separately missing.
`snowc` likewise returns 0-100 where a 0-1 fraction appears intended. Both are
pre-existing unit defects that were invisible while the field was unwritten and
should be tracked in a separate follow-up issue.

NickGeneva:
## Summary

- derive HRRR forecast-validity keys for snow depth, snow cover, and total cloud cover from the requested lead time
- select the HRRR `0-0 day acc fcst` precipitation record at lead zero
- keep hourly precipitation-window selection for positive leads
- add offline regression coverage for all affected variables in both analysis and forecast paths

Fixes #1029

## Testing

- `uv run --extra data pytest test/data/test_hrrr.py -q` (14 passed, 11 slow tests skipped)
- `make lint`
- live 2026-07-20 HRRR lead-zero `tp` byte-range fetch and decode
- live 2026-07-20 HRRR_FX f01 fetches for `sd`, `sde`, `snowc`, and `tcc`; all decoded values finite

## Scope

The unit-conversion observations and missing-output initialization behavior described in the issue are separate concerns and are intentionally unchanged.


earth2studio/lexicon/hrrr.py
```diff
@@ -197,13 +197,13 @@ def build_vocab() -> dict[str, str]:
...
-            "tp": "wrfsfc::APCP::surface::x-x hour acc",  # 1 hour accumulated
+            "tp": "wrfsfc::APCP::surface",
             "fg10m": "wrfsfc::GUST::surface",
             "tcwv": "wrfsfc::PWAT::entire atmosphere (considered as a single layer)",
-            "tcc": "wrfsfc::TCDC::entire atmosphere::anl",
-            "sd": "wrfsfc::WEASD::surface::anl",  # will be zero
-            "sde": "wrfsfc::SNOD::surface::anl",  # will be zero
-            "snowc": "wrfsfc::SNOWC::surface::anl",  # will be zero
+            "tcc": "wrfsfc::TCDC::entire atmosphere",
+            "sd": "wrfsfc::WEASD::surface",
```
gertln, line 204: not introduced here but base.py:77 declares sd as "snow depth (m of water equivalent) while WEASD is kg/m^3

gertln (approved):
LGTM. 
Fixing the unit of sd probably needs a follow up PR.

earth2studio/lexicon/hrrr.py
NickGeneva: ack! thanks for following up

PR #1095, opened 2026-08-25
Use NaN fill for unresolved GRIB fields

### LINKED ISSUE #1028: 🐛[BUG]: GRIB data sources return unwritten output slots as zeros or uninitialised memory
### Version

0.18.0a0

source

### Describe the issue

`GFS`, `GFS_FX`, `HRRR`, `HRRR_FX` and `GEFS_FX` pre-allocate their output array
and fill it task-by-task. When a variable cannot be resolved against the GRIB
index, `_create_tasks` logs a warning and `continue`s, no task is created, and
that variable's region of the array is returned to the caller
**exactly as allocated**.

Three of the five allocate with `np.zeros`, two with `np.empty`. Neither is reliably
detectable by the caller: `0.0` is a legitimate value for precipitation, snow
and cloud cover, and uninitialised memory may contain finite, plausibly-scaled
values. In either case the result can pass `np.isnan`, physical-range checks
and visual inspection. Uninitialised memory may equally contain NaN or
inf, which is why it cannot be relied on as a reproduction signal either.

`CFS_FX` already does this correctly.

| Class | Allocation | Skip path |
|---|---|---|
| `GFS` | [`np.zeros` L244](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/data/gfs.py#L244) | [L324-L327](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/data/gfs.py#L324-L327) |
| `GFS_FX` | [`np.zeros` L661](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/data/gfs.py#L661) | Uses `GFS._create_tasks` through inheritance |
| `HRRR` | [`np.zeros` L301](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/data/hrrr.py#L301) | [L423-L426](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/data/hrrr.py#L423-L426) |
| `HRRR_FX` | [`np.empty` L840](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/data/hrrr.py#L840) | Uses `HRRR._create_tasks` through inheritance |
| `GEFS_FX` | [`np.empty` L239](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/data/gefs.py#L239) | [L339-L342](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/data/gefs.py#L339-L342) |
| `CFS_FX` | [`np.full(nan)` L306](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/data/cfs.py#L306) ✅ | [L409](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/earth2studio/data/cfs.py#L409) |

`CFS_FX` even documents why:

> ```python
> # NaN-initialise so variables that are not resolved against the
> # grib index surface as detectable missing values instead of
> # arbitrary memory.
> ```

### Reproduction (deterministic, no network)

Save the following as `repro_test.py`:

```python
from datetime import datetime
from unittest.mock import patch

import numpy as np

def test_missing_variable_is_not_detectable(tmp_path, monkeypatch):
    monkeypatch.setenv("EARTH2STUDIO_CACHE", str(tmp_path))

    # Index has t2m only; z500 is absent on purpose.
    partial_index = {"1::TMP::2 m above ground": (0, 65300)}
    fake_grid = np.random.rand(721, 1440).astype(np.float32) + 1.0

    async def _fake_fetch_index(uri):
        return partial_index

    async def _fake_fetch_remote_file(*a, **k):
        return str(tmp_path / "ignored.grib2")

    ds = GFS(cache=True, verbose=False)
    with (
        patch.object(ds, "_fetch_index", side_effect=_fake_fetch_index),
        patch.object(ds, "_fetch_remote_file", side_effect=_fake_fetch_remote_file),
        patch("earth2studio.data.gfs._decode_gfs_grib", return_value=fake_grid),
    ):
        data = ds(datetime(2024, 1, 1), ["t2m", "z500"])

    # z500 was never fetched, yet nothing marks it as missing.
    assert not np.isnan(data.sel(variable="z500").values).any()   # passes on main
    assert (data.sel(variable="z500").values == 0.0).all()        # passes on main
```

Then run it from the repository root:

```console
$ uv run pytest repro_test.py -v
```

Both assertions pass today. A caller has no way to tell that region from real data.

Note that the repo's own
[`test_gfs_missing_variable_mock`](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/test/data/test_gfs.py#L347)
and
[`test_hrrr_missing_variable_mock`](https://github.com/NVIDIA/earth2studio/blob/166b2d2e09ecc4783c622fae0295ccf11d4d7c9f/test/data/test_hrrr.py#L389)
already assert `== 0.0`: they codify the current behaviour rather than specify
intended behaviour. Neither `HRRR_FX` nor `GEFS_FX` has a test verifying the
output when a requested variable is missing from the GRIB index.

### Reproduction against real data

`GFS.tp` at lead 0 is a clean case: the `f000` index contains no `APCP` record
at all (696 lines, none), so the skip path always fires.

Save as `repro.py`:

```python
import numpy as np
from datetime import datetime

d = GFS(cache=False)(datetime(2024, 1, 1), ["tp"])
print(np.isnan(d.values).any())
```

```console
$ uv run python repro.py
WARNING | earth2studio.data.gfs:_create_tasks:325 - Variable tp not found in
index file for time 2024-01-01 00:00:00 at 0:00:00, values will be unset
False
```

A global all-zero precipitation field, with nothing marking it as missing.

> ⚠️ This example stops demonstrating the bug if a planned GFS follow-up lands,
> since that change will reject `GFS.tp` outright rather than returning a fill
> value.

### Observed consequence for the `np.empty` classes

**This part is allocator- and platform-dependent and is not a reliable
reproduction step.** In one observed run on glibc/Linux (WSL2, CPython 3.13,
NumPy 2.4), the first array of a given size came back kernel-zeroed while
subsequent ones were served from already-dirtied memory, which is consistent with
glibc's mmap threshold being adjusted after a mapped chunk is freed, though we
did not verify that mechanism beyond this environment. In a loop over times,
the new array contained the **previous fetch's field, bit for bit**.

Allocating an array of the output shape, freeing it, and immediately allocating
another of the same shape:

```
filled 101325.0  -> next np.empty:   0.0000% of elements == 101325.0
filled 9.81      -> next np.empty: 100.0000% of elements == 9.81
```

The first allocation comes back zeroed; the second is served from the memory
just released.

A live `HRRR_FX` fetch of a variable that misses the index returned different
arrays on two identical calls in one process: means of `2.5e270` and `-97.5`
in one session, exact zeros in others. None of it is NaN, and none of it is
guaranteed.

The portable proof for these two classes is the source trace (`np.empty` plus
the `continue`), not a runtime demo.

### Nothing downstream catches it

Checked and ruled out: `gather_with_concurrency` never sees an expected-slot
count; `prep_data_array` and `fetch_data` contain no `isnan`/`isfinite`/
completeness logic; the lexicon modifier never runs for a skipped variable;
`xr.DataArray(data=np.empty(...))` does not copy the buffer
(`np.shares_memory` returns `True`); and `handshake_coords` validates the
coordinate arrays, never the data tensor, so an unwritten region passes.

## Expected behaviour

An unwritten output slot should be NaN, so that missing data is distinguishable
from real data.

## Suggested fix

Switch all five allocations to `np.full(..., np.nan)`, matching `CFS_FX`; update
the two existing tests to assert NaN; add missing-variable tests for the two
`np.empty` classes.

This fixes no lookup: it makes lookup failures detectable. It is a behaviour
change for callers relying on the zero fill and needs a CHANGELOG entry.

NickGeneva:
# Earth2Studio Pull Request

## Description

Initialize GFS, GFS_FX, HRRR, HRRR_FX, and GEFS_FX output arrays with NaN so variables that do not resolve against a GRIB index remain detectably missing. Successful fetch tasks continue to overwrite their output slots normally.

Existing analysis and forecast mock tests now exercise both successful and unresolved variables without adding duplicate test functions. The changelog documents the behavior change.

Closes #1028.

## Testing

- `uv run --extra data pytest test/data/test_gfs.py test/data/test_hrrr.py test/data/test_gefs.py -q` (43 passed, 31 slow tests skipped)
- `make lint`
- Live GFS request for `["t2m", "tp"]` at 2024-01-01: `t2m` fully finite and unresolved `tp` fully NaN

## Checklist

## Dependencies

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- GFS, HRRR, and GEFS GRIB sources now fill unresolved output regions with
+  `NaN` instead of zeros or uninitialized memory, making missing index records
+  detectable
 - `UFSObsConv` and `UFSObsSat` now decode diag files in parallel across a
   persistent spawn-based process pool (new `decode_workers` parameter,
   default `"auto"`); each file is decoded once per request window, frames are
--- a/earth2studio/data/gefs.py
+++ b/earth2studio/data/gefs.py
@@ -236,14 +236,15 @@ class GEFS_FX:
         # but this is much much cleaner to deal with, compared to something seen in the
         # NCAR data source.
         xr_array = xr.DataArray(
-            data=np.empty(
+            data=np.full(
                 (
                     len(time),
                     len(lead_time),
                     len(variable),
                     len(self.GEFS_LAT),
                     len(self.GEFS_LON),
-                )
+                ),
+                np.nan,
             ),
             dims=["time", "lead_time", "variable", "lat", "lon"],
             coords={
--- a/earth2studio/data/gfs.py
+++ b/earth2studio/data/gfs.py
@@ -241,8 +241,9 @@ class GFS:
         # but this is much much cleaner to deal with, compared to something seen in the
         # NCAR data source.
         xr_array = xr.DataArray(
-            data=np.zeros(
-                (len(time), 1, len(variable), len(self.GFS_LAT), len(self.GFS_LON))
+            data=np.full(
+                (len(time), 1, len(variable), len(self.GFS_LAT), len(self.GFS_LON)),
+                np.nan,
             ),
             dims=["time", "lead_time", "variable", "lat", "lon"],
             coords={
@@ -660,14 +661,15 @@ class GFS_FX(GFS):
         # but this is much much cleaner to deal with, compared to something seen in the
         # NCAR data source.
         xr_array = xr.DataArray(
-            data=np.zeros(
+            data=np.full(
                 (
                     len(time),
                     len(lead_time),
                     len(variable),
                     len(self.GFS_LAT),
                     len(self.GFS_LON),
-                )
+                ),
+                np.nan,
             ),
             dims=["time", "lead_time", "variable", "lat", "lon"],
             coords={
--- a/earth2studio/data/hrrr.py
+++ b/earth2studio/data/hrrr.py
@@ -298,14 +298,15 @@ class HRRR:
         # Note, this could be more memory efficient and avoid pre-allocation of the array
         # but this is much much cleaner to deal with
         xr_array = xr.DataArray(
-            data=np.zeros(
+            data=np.full(
                 (
                     len(time),
                     1,
                     len(variable),
                     len(self.HRRR_Y),
                     len(self.HRRR_X),
-                )
+                ),
+                np.nan,
             ),
             dims=["time", "lead_time", "variable", "hrrr_y", "hrrr_x"],
             coords={
@@ -839,14 +840,15 @@ class HRRR_FX(HRRR):
         # but this is much much cleaner to deal with, compared to something seen in the
         # NCAR data source.
         xr_array = xr.DataArray(
-            data=np.empty(
+            data=np.full(
                 (
                     len(time),
                     len(lead_time),
                     len(variable),
                     len(self.HRRR_Y),
                     len(self.HRRR_X),
-                )
+                ),
+                np.nan,
             ),
             dims=["time", "lead_time", "variable", "hrrr_y", "hrrr_x"],
             coords={
--- a/test/data/test_gefs.py
+++ b/test/data/test_gefs.py
@@ -346,15 +346,16 @@ def test_gefs_call_mock(tmp_path, monkeypatch):
     ):
         # Bypass real store by stubbing it to a truthy sentinel.
         ds.store = object()  # type: ignore[assignment]
-        data = ds(datetime(2024, 1, 1), timedelta(hours=3), ["t2m", "z500"])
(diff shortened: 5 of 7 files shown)


gertln (commented):
LGTM

PR #1097, opened 2026-08-25
Fix scorecard plots in versioned docs

NickGeneva:
## Summary

- Keep the existing Mike dependency and publishing behavior unchanged.
- Rewrite only iframes carrying the dedicated `data-e2s-scorecard-plot` marker while Mike builds a versioned site.
- Preserve Mike version prefixes such as `main/` and release versions without inline JavaScript.

## Testing

- Ran pre-commit on the changed generator and project configuration.
- Built and deployed a temporary versioned site with the existing PyPI Mike package.
- Confirmed the Mike-deployed Aurora page contains the static `../../../_static/scorecard/plot.html` URL and that the asset exists.
- Built the same generated page with Zensical and confirmed the static plot URL and asset remain valid.


docs/generate_scorecard.py
```diff
@@ -255,12 +255,20 @@ def _load_defaults() -> dict:
...
+     both MkDocs and Zensical builds. -->
 
-<iframe src="../../_static/scorecard/plot.html?model={model}&label={label_q}" title="{label} skill"
+<iframe title="{label} skill"
         style="width:100%;height:560px;border:1px solid rgba(128,128,128,.35);border-radius:10px;"
         loading="lazy"></iframe>
+<script>
+(() => {{
+  const frame = document.currentScript.previousElementSibling;
+  const plot = new URL("../../../_static/scorecard/plot.html", window.location.href);
```
NickGeneva: dynamic resolving on page access, i guess to handle the mike version?

Dibyajyoti-Chakraborty (approved):
LGTM

PR #1101, opened 2026-08-25
fix: synchronize FengWu ONNX I/O

### LINKED ISSUE #980: 🐛[BUG]: FengWu generates unreasonable TC tracks
### Version

0.9.0

pip

### Describe the issue

Hello,

I have been trying to get FengWu to produce reasonable TC tracks but instead they show zigzags or positions after 6 hours almost at the same place. I have it run for 36 hours and tried to use IFS and GFS as data sources in respective runs. I have sample data trying to track 06E (TS Fausto) in 2026 initalised at 19/7 06:00, which look like this:

Image
Image

I have also attached FuXi which should be what I expect (no problems for Aurora, SFNO and Pangu6 as well). Notice some rows have (almost) the same values. This happens to other TCs.

I have followed the same architecture as the Tropical Cyclone Tracking example to produce these results.

I have ruled out FengWu not iterating at 6-hour timesteps, FengWu not using 2 historical states and center positions of 06E saved in different track IDs (or multiple cyclones saved in the same track ID). I'm unsure what else could cause the center localization jitter that generates a track that is a lot slower than expected. Some pointers will be appreciated.

NickGeneva:
## Description

Closes #980.

FengWu binds PyTorch CUDA tensors directly to ONNX Runtime, but inference could
start before the input buffers were ready or return before the output buffers
were safe for PyTorch to consume. This CUDA stream race caused intermittent
repeated forecast timesteps.

Synchronize the bound inputs before inference and outputs afterward, following
the ONNX Runtime I/O binding contract.

## Testing

- `uv run --extra fengwu pytest test/models/px/test_fengwu.py -q` (11 passed,
  1 package test skipped)
- `make lint`
- Real FengWu checkpoint: 12-step CUDA rollout advanced at every step; before
  the fix, the third forecast step exactly repeated the preceding field

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -136,16 +136,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed intermittent repeated FengWu forecast timesteps by synchronizing ONNX
   Runtime I/O buffers with PyTorch.
-
 - Fixed HRRR GRIB index lookups for total precipitation at lead zero and for
   snow depth, snow cover, and total cloud cover at positive forecast lead
   times.
 - Fixed the March 22, 2021 GFS archive cutoff and historical `GFS_FX` total
   precipitation lookups.
-
 - Fixed `lat_weight` returning a small negative weight at the poles in
   float32, which could give NaN under `sqrt`. Weights are now clamped to be non-negative.
-
 - Fixed `SamudrACE` inference being non-reproducible run-to-run: toggling
   `torch.backends.cudnn.benchmark` on and off around each coupled cycle
   re-triggered cuDNN's GPU-timing-based algorithm search every cycle
@@ -153,7 +150,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   unsliced coords: `lead_time` kept `[-6h, 0h]` while the tensor held one
   step. Coords are now sliced the same way as FuXi, DLWP and FengWu, so the
   initial condition is yielded at `lead_time=[0h]`
-
 - Fixed `CFS_Reforecast_FX` and `CFS_Reforecast_FX_Flux` pointing at the retired
   NCEI archive path; the reforecast archive moved to
   `https://www.ncei.noaa.gov/oa/prod-cfs-reforecast` with renamed product subdirs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -134,6 +134,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed intermittent repeated FengWu forecast timesteps by synchronizing ONNX
+  Runtime I/O buffers with PyTorch.
+
 - Fixed HRRR GRIB index lookups for total precipitation at lead zero and for
   snow depth, snow cover, and total cloud cover at positive forecast lead
   times.
--- a/earth2studio/models/px/fengwu.py
+++ b/earth2studio/models/px/fengwu.py
@@ -308,7 +308,9 @@ class FengWu(torch.nn.Module, AutoModelMixin, PrognosticMixin):
         # Forward pass, fengwu onnx supports batched
         bind_input("input", x)
         output = bind_output("output", like=x)
+        binding.synchronize_inputs()
         ort_session.run_with_iobinding(binding)
+        binding.synchronize_outputs()
 
         # ONNX model outputs two time-steps, take the first
         output_tensor = output[:].contiguous()


NickGeneva:
I ran the fixed FengWu iterator against an ARCO ERA5 initial condition around
Hurricane Harvey. The full state changes at every six-hour step, including the
previously problematic +18 h to +24 h transition:

```text
lead=+06h mean_abs_change=36.901276 max_abs_change=3041.3281 identical=False
lead=+12h mean_abs_change=37.10936 max_abs_change=3251.5 identical=False
lead=+18h mean_abs_change=36.677441 max_abs_change=3123.4297 identical=False
lead=+24h mean_abs_change=35.340355 max_abs_change=2895.75 identical=False
```

The script saves `fengwu_iterator_no_repeat.jpg` in the repository root. The
JPEG is intentionally not attached here so it can be reviewed and uploaded
separately.

Run with:

```bash
uv run --extra fengwu --with matplotlib python fengwu_iterator_example.py
```

```python
from datetime import datetime
from pathlib import Path

import numpy as np
import torch

from earth2studio.data import ARCO, fetch_data
from earth2studio.models.px import FengWu
from earth2studio.utils.time import to_time_array

START_TIME = datetime(2017, 8, 25)
PLOT_STEPS = (1, 2, 3, 4)
OUTPUT = Path(__file__).with_name("fengwu_iterator_no_repeat.jpg")


def main() -> None:
    device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
    model = FengWu.load_model(FengWu.load_default_package()).to(device)
    x, coords = fetch_data(
        source=ARCO(),
        time=to_time_array([START_TIME]),
        variable=model.input_coords()["variable"],
        lead_time=model.input_coords()["lead_time"],
        device=device,
    )

    variables = list(coords["variable"])
    u10m, v10m = variables.index("u10m"), variables.index("v10m")
    previous = None
    panels = []

    for step, (state, step_coords) in enumerate(model.create_iterator(x, coords)):
        current = state[:, 0]
        if previous is not None:
            difference = (current - previous).abs()
            repeated = torch.equal(current, previous)
            lead_hours = int(step_coords["lead_time"][0] / np.timedelta64(1, "h"))
            print(
                f"lead=+{lead_hours:02d}h "
                f"mean_abs_change={difference.mean().item():.8g} "
                f"max_abs_change={difference.max().item():.8g} "
                f"identical={repeated}"
            )
            if step in PLOT_STEPS:
                wind = torch.hypot(current[0, u10m], current[0, v10m])
                old_wind = torch.hypot(previous[0, u10m], previous[0, v10m])
                panels.append(
                    (
                        lead_hours,
                        wind.cpu().numpy(),
                        (wind - old_wind).abs().cpu().numpy(),
                        repeated,
                    )
                )
        previous = current.clone()
        if step == max(PLOT_STEPS):
            break

    lat, lon = step_coords["lat"], step_coords["lon"]
    lat_mask = (lat >= 10) & (lat <= 40)
    lon_mask = (lon >= 255) & (lon <= 290)
    region = np.ix_(lat_mask, lon_mask)
    vmax = max(np.nanpercentile(wind[region], 99) for _, wind, _, _ in panels)
    dvmax = max(np.nanpercentile(delta[region], 99) for _, _, delta, _ in panels)
    fig, axes = plt.subplots(2, len(panels), figsize=(16, 7), constrained_layout=True)

    for column, (lead, wind, delta, repeated) in enumerate(panels):
        image = axes[0, column].pcolormesh(
            lon[lon_mask],
            lat[lat_mask],
            wind[region],
            cmap="viridis",
            vmin=0,
            vmax=vmax,
        )
        change = axes[1, column].pcolormesh(
            lon[lon_mask],
            lat[lat_mask],
            delta[region],
            cmap="magma",
            vmin=0,
            vmax=dvmax,
        )
        axes[0, column].set_title(f"+{lead} h wind speed\nidentical: {repeated}")
        axes[1, column].set_title("Absolute change from prior step")
        for axis in axes[:, column]:
            axis.set(xlabel="Longitude", ylabel="Latitude")

    fig.colorbar(image, ax=axes[0], label="10 m wind speed (m/s)", extend="max")
    fig.colorbar(
        change, ax=axes[1], label="Absolute wind-speed change (m/s)", extend="max"
    )
    fig.suptitle(f"FengWu iterator rollout from {START_TIME:%Y-%m-%d %H:%M} UTC")
    fig.savefig(OUTPUT, dpi=180, format="jpg", pil_kwargs={"quality": 90})
    plt.close(fig)
    print(f"Saved {OUTPUT}")


if __name__ == "__main__":
    main()
```


fengwu_iterator_no_repeat

Dibyajyoti-Chakraborty (approved):
Interesting find! The fix LGTM.

PR #1115, opened 2026-08-27
Fix docker build failure

shrek:
# Earth2Studio Pull Request

## Description

The serve dockerfile build was failing. Fix it with the following:

1) warp was getting pinned to < 1.12.0 in the Dockerfile because an older physicsnemo required it. The current physicsnemo version requires warp to >=1.14.0 breaking the pin.

2) The base image 26.01-py3 was shipping with pytorch 2.10.0a0+a361d3  which is a pre-release of torch 2.10.0. nvidia-physicsnemo==2.2.0a0 (needed for stormcast) requires torch>=2.10.0 was not satisfied.
The fix base image ie, 26.02-py3 ships 2.11.0a0, and 2.11.0a0 > 2.10.0 which satisfies the dependencies. 

The docker image was built with the above changes, and one workflow inference (deterministic workflow) was tested on lepton.
## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/earth2studio/blob/main/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.
- [ ] The [CHANGELOG.md](https://github.com/NVIDIA/earth2studio/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/earth2studio/issues) is linked to this pull request.
- [ ] Assess and address Greptile feedback (AI code review bot for guidance; use discretion, addressing all feedback is not required).

PR #1116, opened 2026-08-27
fix: backport serve Docker build fix to 0.18

NickGeneva:
## Description

Backports the two commits from #1115 to the 0.18.0 release branch.

- Update the NVIDIA PyTorch base image from 26.01 to 26.02.
- Remove the obsolete warp-lang<1.12.0 runtime pin.
- Preserve the source PR merge commit without importing unrelated main changes.

## Validation

- Verified the backport diff contains only serve/Dockerfile.
- Docker build validation resolved the NVIDIA 26.02 base image; the local check could not access ghcr.io/astral-sh/uv because the host lacks GHCR credentials.

### DIFF
--- a/serve/Dockerfile
+++ b/serve/Dockerfile
@@ -1,4 +1,4 @@
-FROM nvcr.io/nvidia/pytorch:26.01-py3
+FROM nvcr.io/nvidia/pytorch:26.02-py3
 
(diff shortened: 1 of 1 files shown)


NickGeneva:
Already approved and tested in #1115 this just points to rc

PR #1119, opened 2026-08-28
Pangu Lazy Initialization

Dibyajyoti-Chakraborty:
## Description

`Pangu6` and `Pangu3` rebuilt their extra ONNX `InferenceSession`s (24 h, plus 6 h for Pangu3) at the start of **every** `_default_generator` call — about 9 s per session — even for short rollouts that never execute the code using
them. This change builds each session on the first step that actually needs it and caches it on the model, so the common single-step case skips the cost entirely and long rollouts pay it once per model lifetime instead of once per
call. Rollout outputs are unchanged.

## Benchmark Test

```
import time

import numpy as np
import torch

from earth2studio.models.px import Pangu6

model = Pangu6.load_model(Pangu6.load_default_package()).to("cuda:0")
coords = OrderedDict(
    batch=np.array([0]),
    time=np.array([np.datetime64("2025-01-01T00:00:00")]),
)
coords.update({k: v for k, v in model.input_coords().items() if k != "batch"})
x = torch.zeros(*[len(v) for v in coords.values()], device="cuda:0")

for _ in range(4):
    t0 = time.perf_counter()
    it = model.create_iterator(x, coords)
    next(it)  # IC yield
    next(it)  # first 6h step
    torch.cuda.synchronize()
    print(f"{time.perf_counter() - t0:.2f}s")
```

## Results (H100, this branch vs `main`)

| Branch             | call 1 | call 2 | call 3 | call 4 |
| ------------------ | ------ | ------ | ------ | ------ |
| `main` (eager)     | 9.77 s | 9.62 s | 9.51 s | 9.70 s |
| `pangu_ops` (lazy) | 2.10 s | 0.40 s | 0.40 s | 0.40 s |

This is a major difference, especially in MoWE online datalaoder type cases where we run Pangu again and again for shorter lead times only.

## Checklist


pzharrington (approved):
I have one suggestion, take it or leave it: I suggest making it configurable to select for either of these two behaviors:

- Build all sessions at once, and cache
- Build lazily and cache

Basically, caching always helps, but better to provide both options. In the eager case, someone running longer rollouts per init would not pollute the step/iterator time within one rollout with session creation time.

Dibyajyoti-Chakraborty:
Added the option to have a configurable session option based on @pzharrington's comments.

PR #1121, opened 2026-08-28
Added regional split and new metrics to eval recipe

Dibyajyoti-Chakraborty:
### Description

- **Regional scoring**: `scoring.online.regions` accepts named latitude/longitude boxes (or lists of boxes, scored as their union) and adds a labeled `region` axis to `stats.zarr`/`scores.zarr`. Regions are applied as masked cosine-latitude weights inside the existing accumulation step, so regional scores cost one matrix product — no extra passes over the forecast. A null box means global.
- **New online metrics**: opt-in per-member MAE (`scoring.online.mae`) and log spectral distance (`scoring.online.lsd`, reusing `earth2studio.statistics.log_spectral_distance`), accumulated online alongside the existing RMSE/CRPS/spread statistics.
- **Fix**: clearing resume markers no longer races between ranks; concurrent removal of the same progress directory retries with backoff until it no longer exists.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `scoring.online.regions` takes named lat/lon boxes and adds a labeled
   `region` axis to `stats.zarr`/`scores.zarr`. New optional per-member
   metrics include MAE (`scoring.online.mae`) and log spectral distance
-  (`scoring.online.lsd`.
+  (`scoring.online.lsd`)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   respectively.
 - Added `eager_sessions` option to Pangu6 and Pangu3 to build the extra
   ONNX sessions at construction instead of on first use in a rollout
+- Added regional splits to evaluation recipe online scoring:
+  `scoring.online.regions` takes named lat/lon boxes and adds a labeled
+  `region` axis to `stats.zarr`/`scores.zarr`. New optional per-member
+  metrics include MAE (`scoring.online.mae`) and log spectral distance
+  (`scoring.online.lsd`.
 
 ### Changed
 
@@ -37,6 +42,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 ### Fixed
+
+- Evaluation recipe: clearing resume markers no longer races between
+  ranks. Concurrent removal of the same progress directory retries until
+  the directory no longer exists
 
 ### Security
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   source (`MeteosatLI`), providing per-flash, per-group and per-event
   detections from the LFL, LGR and LEF collections as a data frame
 - Added group- and flash-level variables to the `GOESGLM` data source
+- Added `eager_sessions` option to Pangu6 and Pangu3 to build the extra
+  ONNX sessions at construction instead of on first use in a rollout
+- Added regional splits to evaluation recipe scoring, in both the online
+  and store-then-score pathways: `scoring.regions` takes named coordinate
+  boxes and adds a labeled `region` axis to the score stores. New
+  optional per-member online metrics include MAE (`scoring.online.mae`)
+  and log spectral distance (`scoring.online.lsd`)
 
 ### Changed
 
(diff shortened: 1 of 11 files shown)


recipes/eval/cfg/default.yaml
```diff
@@ -276,6 +279,8 @@ scoring:
         # point CONUS-scale run from ~1.5GB to well under 250MB, at some
         # throughput cost).
         pairwise_member_tile: 8
+        # Named regional splits.
+        regions: null
```
pzharrington: Can you provide 1-2 commented out examples here of how the regions are actually configured?

recipes/eval/src/report/aggregation.py
```diff
@@ -111,7 +111,21 @@ def load_scores(cfg: DictConfig) -> xr.Dataset:
             "Run scoring (score.py) before generating a report."
         )
     logger.info(f"Opened score store: {store_path}")
-    return xr.open_zarr(store_path)
+    ds = xr.open_zarr(store_path)
+    if "region" in ds.dims:
+        labels = [str(r) for r in ds.region.values]
+        pick = str(
+            cfg.get("report", {}).get(
+                "region", "global" if "global" in labels else labels[0]
```
pzharrington, line 119: Defaulting to `global` makes sense for global models. I do think we'll need to revisit this choice for regional models, where `global` doesn't make as much sense...

recipes/eval/src/online.py
```diff
@@ -1709,7 +2120,13 @@ def add_stats_arrays(
         Backend from ``OutputManager.io``.
     array_groups : list[tuple[CoordSystem, list[str]]]
         Groups from :func:`stats_array_groups`.
+    region_names : Sequence[str] | None
+        Regional split names, recorded in the store's attributes so the
+        integer ``region`` axis stays interpretable (and so
+        :func:`finalize_stats` can reattach them as coordinate labels).
     """
+    if region_names is not None:
+        io.root.attrs["regions"] = list(region_names)
```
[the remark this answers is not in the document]
Dibyajyoti-Chakraborty: Updated

recipes/eval/src/report/aggregation.py
Dibyajyoti-Chakraborty: For regional models, users must select them manually because the regions can vary. But we need to think of something better eventually.

recipes/eval/cfg/default.yaml
Dibyajyoti-Chakraborty: Done.

recipes/eval/src/online.py
```diff
@@ -2246,37 +2679,102 @@ def _write_slab(self, time: np.datetime64) -> None:
...
 
-    lat_vals = np.asarray(spatial_coords["lat"])
-    w = lat_weight(torch.tensor(lat_vals, dtype=torch.float64))
-    shape[dims.index("lat")] = len(lat_vals)
-    return w.reshape(shape)
+    if "lat" not in dims or "lon" not in dims:
+        raise ValueError(
+            "scoring.online.regions needs 'lat' and 'lon' spatial "
+            f"dimensions on the scored grid; got {dims}."
+        )
```
pzharrington: Similar comment, we'll need to extend this for regional models. I think it won't be too bad if we require that regions always be rectangular boxes, then just the dim names and numbers change (e.g. box would be in HRRR grid indices for StormCast-CONUS) but the logic will be the same

recipes/eval/src/online.py
pzharrington, line 1: This appears to only work in the online scoring pathway. I think we should support it as well in the traditional store-then-score pathway (esp if someone comes up with new regions/metrics, they don't want to re-run inference). Can you add that? Config should be done similar to the online scoring (in fact maybe the whole `regions` config block can just be pulled out of `online` and live in `scoring.regions`, used by both pathways...)

Also, we should add documentation on this to the README (users and agents should be able to get going quickly with it). Then it should be good to go.

recipes/eval/src/online.py
Dibyajyoti-Chakraborty: Updated.

recipes/eval/src/online.py
Dibyajyoti-Chakraborty: Updated

CHANGELOG.md
```diff
@@ -11,21 +11,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-  `lightning_flash_footprint_pixels`. `GOESGLM` variable ids `flashe` and
-  `flashc` are renamed to `lightning_event_energy` and `lightning_event_count`
-  respectively.
 - Added `eager_sessions` option to Pangu6 and Pangu3 to build the extra
   ONNX sessions at construction instead of on first use in a rollout
+- Added regional splits to evaluation recipe online scoring:
+  `scoring.online.regions` takes named lat/lon boxes and adds a labeled
+  `region` axis to `stats.zarr`/`scores.zarr`. New optional per-member
+  metrics include MAE (`scoring.online.mae`) and log spectral distance
+  (`scoring.online.lsd`)
```
pzharrington: Not just online scoring anymore

Dibyajyoti-Chakraborty: Oh ya was updating it rn lol.

pzharrington (approved):
LGTM pending the changelog update!

CHANGELOG.md
Dibyajyoti-Chakraborty: Done.

PR #1122, opened 2026-08-28
Let the blog plugin order posts

NickGeneva:
## Summary

- remove the manually maintained Latest Updates block from the blog index
- let the Material blog plugin order posts from their front-matter dates
- keep the newer soil-health post ahead of the older MkDocs migration post

## Validation

- built the documentation with Zensical
- built the versioned documentation with MkDocs and Mike
- verified the newer post renders first and the manual section is absent


pzharrington (approved):
This is an ideal Friday afternoon PR

PR #981, opened 2026-07-20
StormScope MeteoSat Europe

jleinonen:
# Earth2Studio Pull Request

## Description

Adds the European StormScope MeteoSat model. 

1. Model wrapper: `earth2studio.models.px.stormscope_meteosat.StormScopeMeteosatEU`
2. Example in `examples/04_nowcasting/05_stormscope_meteosat_example.py`
3. Unit tests
4. Some fixes/improvements to the `MeteosatFCI` data source:
    - Fix `projection_extent` to account for FCI coordinate convention
    - `available_variables` to list the variables available at the selected resolution
    - Recycle opened data files (this speeds up data loading in the example by ~5x on my system)

Depends on https://github.com/NVIDIA/physicsnemo/pull/1848.

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/earth2studio/blob/main/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.
- [ ] The [CHANGELOG.md](https://github.com/NVIDIA/earth2studio/blob/main/CHANGELOG.md) is up to date with these changes.
- [ ] An [issue](https://github.com/NVIDIA/earth2studio/issues) is linked to this pull request.
- [x] Assess and address Greptile feedback (AI code review bot for guidance; use discretion, addressing all feedback is not required).

PR #1126, opened 2026-08-31
Fix StormScopeMeteosatEU default package pointing to stormcast-conus

negin513:
## Summary
- `StormScopeMeteosatEU.load_default_package()` (added in #981) referenced `hf://nvidia/stormcast-conus`, apparently copy-pasted from `stormcastconus.py`, instead of the actual EU Meteosat package.
- Points it at `hf://nvidia/stormscope-meteosat` instead. The pinned commit hash (`17148d712f7ef2c0b48355032dcad85173dd230e`) was already correct — it's the tip commit of `nvidia/stormscope-meteosat` and its tree contains `config.yaml`, both `.mdlus` checkpoints, and `metadata.nc`, matching what `load_model()` expects. Only the repo name and `cache_storage` label needed correcting.

## Test plan
- [x] Verified via the Hugging Face API/UI that `nvidia/stormscope-meteosat@17148d712f7ef2c0b48355032dcad85173dd230e` exists and contains the expected files.
- [ ] `load_default_package()` / `load_model()` successfully downloads and loads `StormScopeMeteosatEU` (not run here — no network/HF access in this environment).

### DIFF
--- a/earth2studio/models/px/stormscope_meteosat.py
+++ b/earth2studio/models/px/stormscope_meteosat.py
@@ -788,9 +788,9 @@ class StormScopeMeteosatEU(torch.nn.Module, AutoModelMixin, PrognosticMixin):
     def load_default_package(cls) -> Package:
         """Load prognostic package"""
         package = Package(
-            "hf://nvidia/stormcast-conus@17148d712f7ef2c0b48355032dcad85173dd230e",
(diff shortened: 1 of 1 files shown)


pzharrington:
Fixed in rc already

NickGeneva:
https://github.com/NVIDIA/earth2studio/pull/1127 Closing via rc rebase

PR #1134, opened 2026-09-02
Fix StormCast v1 mutating its input tensor in place

### LINKED ISSUE #1133: 🐛[BUG]: StormCast v1 mutates its input tensor in place (StormCastCONUS clones)
### Version

0.17.0 (still present on main as of 2026-08-31)

pip, source

### Describe the issue

StormCast.__call__ writes its result into the input tensor and returns the
same tensor object. The caller's initial condition is overwritten and no
error is raised.

stormcast.py L492-494 at 0.17.0 (L493-494 on main):

```python
x[i, j, k : k + 1] = self._forward(
    x[i, j, k : k + 1], conditioning[i, j, k : k + 1]
)
```

There is no clone on this path. StormCastCONUS has the same loop shape but
clones first, stormcastconus.py L598:

```python
x = x.clone()  # prevent editing of argument
```

So the two classes disagree, and v1 is the surprising one.

What it means for a caller:

- Any reference kept to the initial condition tensor holds forecast output
  after the first call. The shapes do not change, so nothing downstream
  notices.
- Under create_iterator, every yielded (x, coords) aliases the same storage.
  Keeping step N and reading it after step N+1 gives step N+1 values.

I found this while building a StormCast v1 run over the 14 March 2025
outbreak (HRRR initial conditions, GFS_FX conditioning), and worked around
it by cloning the IC before create_iterator.

### Minimum reproducible example

```python
model = StormCast.load_model(StormCast.load_default_package())
x, coords = fetch_data(
    source=HRRR(),
    time=time,
    variable=model.input_coords()["variable"],
    lead_time=model.input_coords()["lead_time"],
    device=device,
)
x, coords = map_coords(x, coords, model.input_coords())

before = x.clone()
out, _ = model(x, coords)

torch.equal(before, x)  # False, x now holds the +1h forecast
```

### Suggested fix

Mirror StormCastCONUS and clone at the top of __call__. It is one line and
it makes the two classes consistent. If the in-place write is intentional
for memory reasons, a note on __call__ would at least make it visible.

Kira-Ryan:
Fixes #1133.

`StormCast.__call__` writes each step's output back into the tensor the caller passed in (`x[i, j, k:k+1] = self._forward(...)`) without cloning first, so the initial condition handed to the model is overwritten by the forecast. `StormCastCONUS` already clones its input (`x = x.clone()  # prevent editing of argument`); this brings v1 in line with it.

Changes
- `earth2studio/models/px/stormcast.py`: clone the input before the batch/time/lead loops.
- `test/models/px/test_stormcast.py`: `test_stormcast_call_does_not_mutate_input` asserts the input tensor is unchanged after the call and that the output does not alias it. It fails on main and passes with the fix.
- `CHANGELOG.md`: entry under Fixed.

The cost is one extra copy of the input per call, the same as StormCastCONUS pays today.


NickGeneva (approved):
LGTM, thanks for catching this!

PR #1135, opened 2026-09-02
Warn and detect missing GOES data

### LINKED ISSUE #1118: 🐛[BUG]: StormScope 3km_10min intermittently returns all-NaN forecasts from finite inputs
### Version

Source installation at commit `076242b0d135374dbe5e1858d55335c55e946c9f`
(version reports `0.17.0a0`).

We have not yet reproduced this against the `0.17.0` release tag.

### Installation method

Source

### Describe the issue

We observed two StormScope `3km_10min` inference runs that completed
without raising an exception but returned entirely non-finite GOES and
MRMS forecast arrays.

Affected forecast initialization times:

- 2026-08-27T06:00:00Z
- 2026-08-28T06:00:00Z

Input validation immediately before inference showed:

- GOES channel finite fractions: approximately 0.98
- MRMS reflectivity, base reflectivity, and GLM finite fractions: 1.00

The completed forecast arrays contained:

- GOES: finite fraction 0.0 for all eight output channels
- MRMS: finite fraction 0.0 for reflectivity, base reflectivity, and GLM

Other initialization times have produced finite forecasts in the same
environment.

The coupled rollout follows the public StormScope example. Relevant
configuration:

- model: `3km_10min`
- batch size: 1
- `num_steps=25`
- `S_churn=10`
- `amp=True`
- `compile=False`
- no explicit RNG seed
- 36 autoregressive 10-minute forecast steps

Environment:

- Linux x86_64
- NVIDIA L40S
- Python 3.12
- PyTorch 2.10.0 / CUDA 12.8
- NATTEN 0.21.6 CUDA build

We did not instrument the original runs deeply enough to determine
whether the first non-finite value originated in GOES, MRMS,
conditioning construction, or the diffusion sampler.

Could you advise whether this is a known instability involving AMP,
stochastic sampling, or reduced sampler steps? Would you recommend
testing `amp=False`, the default 100 sampler steps, or a specific fixed
seed first?

It may also be useful for Earth2Studio to fail explicitly when
conditioning or model output becomes non-finite, rather than completing
the rollout with all-NaN arrays.

We can run a targeted configuration matrix and provide redacted tensor
statistics or checkpoint hashes if useful.

pzharrington:
# Earth2Studio Pull Request

## Description
Addresses #1118 . Detects and warns when data in from the GOES source contains invalid pixels not part of the expected region (outside the visible disk). Also will catch any non-sanitized NaNs in the StormScope model and error out if detected.

## Checklist


earth2studio/data/goes.py
```diff
@@ -410,6 +427,34 @@ async def fetch_array(
...
+            )
+            if dqf_name is not None and dqf_name in da:
+                navigated_mask = ~np.isnan(da[dqf_name].values)
+            else:
+                navigated_mask = self._on_disk_mask
+
+            nan_navigated = np.isnan(x[i]) & navigated_mask
+            nan_count = nan_navigated.sum()
+            if nan_count > 0:
+                nan_frac = nan_count / navigated_mask.sum()
```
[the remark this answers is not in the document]
pzharrington: Added

pzharrington:
@greptileai

sylvesterkaczmarek (commented):
On CUDA, the successful-path check `bool(torch.isnan(condition).any())` performs a full reduction and forces a device-to-host synchronization on every StormScope forecast, even when the tensor is valid. Could this validation be moved earlier in conditioning construction, or use a device-side check, so normal inference does not gain a synchronization barrier?

earth2studio/data/goes.py
```diff
@@ -410,6 +427,34 @@ async def fetch_array(
...
+            # Prefer the file's own per-band DQF variable to identify which
+            # pixels NOAA actually navigated/quality-assessed: DQF is NaN
+            # (fill) exactly where a pixel was never computed (off the Earth
+            # disk), and holds a real flag_values code (0-4) for every pixel
+            # that was. This is exact per-file ground truth, unlike our
+            # static ellipsoid-based on-disk mask, which can disagree with
+            # NOAA's navigation by a pixel or two right at the limb.
+            dqf_name = (
+                goes_name.replace("CMI_", "DQF_", 1)
+                if goes_name.startswith("CMI_")
```
NickGeneva: This is only applicable to CMI bands or something? Whats CMI here?

NickGeneva (approved):
Over all LGTM, just a question, approving to unblock.

NickGeneva:
One thing to note with this nan search, codex claims:

```
That is nearly 50% extra local decode/scan time. It may be acceptable for diagnostics, but it should be considered, especially for cached and multi-time requests.
```

Can you take a look to see if theres an easy perf improvement that could be done?

earth2studio/data/goes.py
```diff
@@ -410,6 +427,34 @@ async def fetch_array(
...
+                else None
+            )
+            if dqf_name is not None and dqf_name in da:
+                navigated_mask = ~np.isnan(da[dqf_name].values)
```
mordisdevone: I went and checked this against the actual files behind our failed 06Z runs, and I think the DQF mask drops the wedge it's meant to catch.

What I did: pulled the GOES-19 CONUS multi-band files from the public bucket, e.g. `s3://noaa-goes19/ABI-L2-MCMIPC/2026/244/05/` (2026-09-01, 05Z hour), opened them with xarray, and counted three things per band:

```
cmi_nan = isnan(CMI_C13)
on_disk = ABI fixed-grid discriminant test (same idea as self._on_disk_mask)
dqf_nan = isnan(DQF_C13)   # what navigated_mask keys off
```

For the 05:01 scan, C13:

```
NaN pixels on the disk:             815,106
...of which DQF is also fill (255): 814,722
what this warning would report:         384   (0.05%)
```

Same ratio on every scan and every date I checked (Aug 27 to Sep 3). Inside the wedge NOAA leaves DQF at `_FillValue` rather than writing a flag code, so "finite DQF = navigated" excludes exactly those pixels. The few hundred that do get counted are the ones carrying DQF=3 around the wedge edge.

The `_forward` check still catches it, so nothing unsafe gets through. It's just that the warning would print something that looks like limb noise for the case that started this.

Two thoughts: using `self._on_disk_mask` as the base (with DQF only trimming the limb, or the union of the two) gives the right count, and it drops the extra DQF read that the scan-time comment above is about. And the test would catch this if DQF were written as uint8 with `_FillValue` at the missing pixel, since that's how the real files decode. Happy to send the counting script or a patch for either if that's useful. Small one: `torch.isfinite` in the guard costs the same as `isnan` and covers inf too.

earth2studio/data/goes.py
pzharrington: Yes, "Cloud and Moisture Imagery" which are the 16 bands we serve via `abi01c`-`abi16c`

pzharrington: They each get their own quality flag `DQF_C##`

earth2studio/data/goes.py
pzharrington: Thanks, this should be addressed now.

pzharrington:
@NickGeneva performance concern should be alleviated too with the most recent update, I thought relying on DQF would help but it turns out just using the scan geometry constants is probably enough.

earth2studio/data/goes.py
mordisdevone: Confirmed, thanks Peter. Ran the new on-disk + limb logic over the same files: the 09-01 05:01 C13 scan now reports 815,105 (was 384), the clean 00Z scan stays quiet, and the one limb pixel lands in debug. Looks right to me, appreciate the quick turnaround.

PR #1136, opened 2026-09-03
fix(io): synchronize NetCDF4 writes

### LINKED ISSUE #546: 🐛[BUG]: NetCDF4 backend synchronization issue
### Version

Latest main

source

### Describe the issue

There seems to be a subtle issue with NetCDF4 files not being properly flushed to disk leading to issues trying to move/rename the file immediately after writing.

On my system, this works ok (`output_file` is on a Lustre file system):

```python
results = NetCDF4Backend(output_file)
run.deterministic([start_time], num_steps, model, source, results, device=device)

ds = xr.open_dataset(output_file)
print(float(ds["t2m"].max()))
```

outputs `304.91934`.

Meanwhile this leads to an output file that consists of fill values:

```python
results = NetCDF4Backend("/tmp/tmp_file.nc")
run.deterministic([start_time], num_steps, model, source, results, device=device)
shutil.move("/tmp/tmp_file.nc", output_file)

ds = xr.open_dataset(output_file)
print(float(ds["t2m"].max()))
```

outputs `9.969209968386869e+36`. Note that this involves a move from the local `/tmp` file system to Lustre.

Adding `.sync()` of the NetCDF4 `Dataset` fixes the issue:

```python
results = NetCDF4Backend("/tmp/tmp_file.nc")
run.deterministic([start_time], num_steps, model, source, results, device=device)
results.root.sync()  # <-- ADDED
shutil.move("/tmp/tmp_file.nc", output_file)

ds = xr.open_dataset(output_file)
print(float(ds["t2m"].max()))
```

outputs `305.44061` (the model is probabilistic, so it's normal to have a different result).

The latter is a quick fix but I don't think it's a great solution as it breaks the interchangeability of IO backends (`results.root.sync()` will fail if `results` is another backend that doesn't have this attribute). Would it be reasonable to add `self.root.sync()` at the end of the `write` method of `NetCDF4Backend`? I'm not sure if this would have adverse performance effects.

marco-trotta1:
# Earth2Studio Pull Request

## Description

Call `sync()` after each NetCDF4 array write. This makes written values visible to other readers before the backend closes.

Add a regression test that copies the active file and verifies its stored values with xarray.

Closes #546.

## Checklist

- [x] I am familiar with the Contributing Guidelines.
- [ ] The CHANGELOG is up to date with these changes.
- [x] An issue is linked to this pull request.
- [ ] Assess and address Greptile feedback.

## Dependencies

No new dependencies.

## Verification

- `uv run pytest test/io/test_netcdf4.py -q -k 'not cuda'`: 29 passed, 27 CUDA cases deselected.
- CPU I/O suite: 135 passed, 5 skipped, 91 CUDA cases deselected.
- Ruff checks passed on both changed files.


NickGeneva (approved):
LGTM, thanks for the fix!

PR #1152, opened 2026-09-10
[E2S 1.0] Time statistics spec

NickGeneva:
## Summary

- add temporal-statistic source planning and block-wise Xarray reductions
- support analysis times and forecast lead times for `mean`, `sum`, `min`, and `max` windows
- support subsecond through daily units, fixed 7-day weeks, and fixed 30-day months
- enforce exact half-open windows and reject missing or duplicate source coordinates
- add the time-statistics specification, example, and implementation research

## Intended `fetch_data` integration

Statistic-qualified variable labels such as `t2m:mean:1day` are the canonical model and output coordinate representation. This allows one model to request daily, weekly, and monthly variants of the same source variable without ambiguity. A mapping remains available as shorthand when each source variable appears at most once.

The fetch layer can then:

1. split qualified labels into source variables and modifiers
2. derive the source timestamps or lead times required for each modifier
3. group variables sharing a modifier into a single fetch and reduction
4. select and validate the exact half-open window
5. assemble the resulting DataArray with the requested qualified labels while preserving the NumPy, CuPy, or Dask backend

This PR provides the planning, reduction, registration, grouping, and metadata utilities. Wiring them into `fetch_data` is intentionally left to the coordinate/data-fetch integration work.

## Testing

- `uv run pytest -q test/utils/test_time_statistics.py test/grids` (53 passed)
- `uv run coverage report -m earth2studio/utils/time_statistics.py` (94%)
- `uv run pre-commit run --files earth2studio/utils/time_statistics.py test/utils/test_time_statistics.py dev/spec/TIME_STATISTICS_SPEC.md dev/examples/02_time_statistics.py dev/scratch/TIME_STATISTICS_RESEARCH.md`


earth2studio/utils/time_statistics.py
```diff
@@ -0,0 +1,302 @@
...
+import numpy as np
+import xarray as xr
+from numpy.typing import NDArray
+
+TimeReduction: TypeAlias = Callable[[xr.DataArray, Hashable], xr.DataArray]
+StatisticDeclaration: TypeAlias = Mapping[str, str] | str | None
+Duration: TypeAlias = str | np.timedelta64
+TemporalTarget: TypeAlias = np.datetime64 | np.timedelta64
+
+_DURATION_PATTERN = re.compile(r"^([+-]?)(\d+)(ns|us|ms|min|m|s|h|d)$")
```
pzharrington: We should also support weekly and monthly durations here for S2S/S2D.

dev/spec/TIME_STATISTICS_SPEC.md
```diff
@@ -0,0 +1,126 @@
...
+```python
+statistics = {
+    "u10m": "mean:24h",
+    "v10m": "mean:24h",
+    "t2m": "max:24h",
+}
+```
+
+A single modifier applies to every requested variable. Variables absent from a
+mapping remain instantaneous.
```
pzharrington: Spec should list out the different supported duration units (min, sec, hr, days, etc)

dev/spec/TIME_STATISTICS_SPEC.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Declaration
+
+Variables remain strings. A mapping adds a temporal modifier to selected variables:
+
+```python
+statistics = {
+    "u10m": "mean:24h",
+    "v10m": "mean:24h",
+    "t2m": "max:24h",
```
pzharrington: I know this is going beyond the scope of this PR, since the next one will address actual data fetch. But are we sure about this dict-based approach? I'm thinking about how this data will move through the package after arriving in a data source...we need to tie in the statistic tag on e.g. a returned `u10m` array/tensor that has had the 24h mean applied, so downstream consumers know what quantity it is. There are a number of ways to achieve that, which I think we can do without much trouble. But how would a model specify things in input/output coords? Imagine some evil user makes an S2S model that takes as input three versions of `t2m` with 3 different averages applied (e.g. daily, weekly, monthly)...how would that model specify that input?

pzharrington: A possible spicy alternative is to tie in the cadences into the lexicon. Forbid all base variable names from using the `:` character, then we could in theory allow a `u10m:mean:24h` kind of pattern as variable name and it would remain parsable by other parts of the package that need to know time reduction info. I'm aware that's a hideous variable name but just trying to get ideas going around possible extensions...

earth2studio/utils/time_statistics.py
NickGeneva: right good call

dev/spec/TIME_STATISTICS_SPEC.md
NickGeneva: Yeah I havent exaclty decided how that would work in the coord yet, your point about like having t2m min and max  or like difference averages is a possibility (I think the oribit model does this already)... aligning the variable names directly to have both the name and then also the stat string might be the most hardened implementation tbh.

pzharrington: This is important for coupling too so @negin513 should weigh in

earth2studio/utils/time_statistics.py
NickGeneva: added

dev/spec/TIME_STATISTICS_SPEC.md
NickGeneva: I think this is more applicable to the coordinate system update in the following PR than this one here which is largely just averaging utiltiy.

Ultimately its fetch_data implementation that drives most of the details.

pzharrington: Ok fair, we can punt

pzharrington (approved):

PR #1167, opened 2026-09-17
[E2S 1.0] Coordinate system update

NickGeneva:
## Summary
- add an allocation-free Xarray CoordinateSystem signature and coordinate helpers
- update FCN coordinate declarations and fetch_data to use the new signature where requested
- retain the existing CoordSystem contract for all other models and IO backends

## Testing
- uv run pre-commit run --all-files
- focused coordinate, Random regression, fetch-data, and FCN tests (167 passed, 15 skipped)


earth2studio/models/px/fcn.py
```diff
@@ -116,67 +148,25 @@ def __init__(
...
-        output_coords = output_coords.copy()
-        output_coords["batch"] = input_coords["batch"]
-        output_coords["lead_time"] = (
-            output_coords["lead_time"] + input_coords["lead_time"]
+    def output_coords(self, input_coords: CoordinateSystem) -> CoordinateSystem:
+        """Return the FCN coordinate signature after one forecast step."""
+        handshake_dataarray(input_coords, self.input_coords())
+        return input_coords.assign_coords(
+            lead_time=input_coords["lead_time"] + np.timedelta64(6, "h")
         )
```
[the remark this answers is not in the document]
NickGeneva: Thats okay

dev/spec/GRID_SPEC.md
```diff
@@ -88,6 +88,36 @@ Selected index values avoid generating geographic coordinates for the full grid:
...
+signature = coord_array(
+    ("batch", "variable", "lat", "lon"),
+    {"variable": ["t2m"]},
+    dynamic=("batch",),
+    grid="fcn1",
+)
+```
+
+StormCastCONUS uses the standard projected `y, x` dimensions directly, preserving
+native meter values and supporting `infer_grid()` without renaming.
```
pzharrington: StormCastCONUS mention here in spec is a bit random...why is this a special model?

dev/spec/GRID_SPEC.md
```diff
@@ -88,6 +88,36 @@ Selected index values avoid generating geographic coordinates for the full grid:
...
+
+Optional `grid_dims` maps standard grid axes to a model's
+dimension names, including grid-coordinate dimensions and the `dims` metadata.
+Explicit coordinates use the mapped names. This mapping does not rename the grid
+definition itself: rename model axes back before passing such an array to
+`infer_grid()` or grid selection methods. A cropped grid should use a definition
+constructed from the cropped axes, not the identifier of its full parent grid.
```
pzharrington: This description is a bit unclear to me

dev/spec/MODEL_CONTRACT_SPEC.md
```diff
@@ -2,38 +2,108 @@
...
-positionally against it.
+The new `CoordinateSystem` is an `xarray.DataArray` created by `coord_array()`.
+Its backing array stores only shape and dtype: it allocates no field values,
+including when every dimension has a concrete size. Read labels through `.coords`,
+order through `.dims`, and lengths through `.sizes`. Accessing field `.values` is
+unsupported. Dimension and auxiliary coordinate arrays themselves occupy memory.
+
+`input_coords()` is a *declaration*. Dynamic dimensions form an explicitly marked,
+zero-sized leading prefix (`earth2studio_dynamic_dims`), conventionally `batch`
+and, for regional models, `time`. The remaining dimensions, labels, auxiliary
```
pzharrington: Why only regional models have dynamic `time`?

dev/spec/MODEL_CONTRACT_SPEC.md
```diff
@@ -2,38 +2,108 @@
...
+coordinates, and declared grid/CRS/statistics metadata must match. Concrete inputs
+may use any leading batch dimensions or none; fixed trailing dimension order is
+authoritative. A zero-sized fixed dimension is not implicitly a wildcard.
+Resolve multiple dynamic dimensions together or from right to left: concretizing
+`time` in `(batch, time, ...)` retains a dynamic `batch`, but concretizing only
+`batch` would leave a non-leading wildcard and is rejected. Dimensions are not
+silently reordered or converted from wildcards to fixed zero-length axes.
```
pzharrington, line 33: A wrinkle here for the recent RoPE-based diffusion models (but any model able to take flexible input sizes). How do we manage `input_coords` and `output_coords` for that case? imo it's maybe a broader model contract question, how to standardize these more flexible models?

pzharrington: Similar, there may be some "foundation model" type things (or just obs-based models) that can take a flexible set of input variables (imagine e.g. stormscope that accepts satellite + lightning when lightning is available, or just satellite alone when not available)

dev/spec/MODEL_CONTRACT_SPEC.md
```diff
@@ -2,38 +2,108 @@
...
+validate geographic coordinates as well as axes.
+
+| Model | Grid declaration | Spatial dimensions | Output |
+| --- | --- | --- | --- |
+| `StormScopeGOES`, `StormScopeMRMS` | `CurvilinearGrid` from checkpoint geometry | `y, x` | Configured output offsets added to final input lead time |
+| `StormCastCONUS` | Cropped `ProjectedGrid` using registered HRRR CRS | `y, x` | Final input lead time plus one hour |
+| `PrecipitationAFNO` | Registered `fcn1` grid (720 × 1440) | `lat, lon` | `tp`, with `sum:6h` statistics |
+
+Regional validation subtracts the final input lead time before checking the
+declared history window. Before subtraction, `lead_time` must be an explicit,
```
pzharrington, line 81: "regional validation"?

dev/spec/GRID_SPEC.md
NickGeneva: Ehhhh probably just because it uses the HRRR grid, its using it as an example, can get it to talk more generally...

Previously we had like hrrr_y hrrr_x, the idea is now its just y, x, and the grid will create this for the user if the grid is known.

dev/spec/TIME_STATISTICS_SPEC.md
```diff
@@ -1,164 +1,103 @@
...
+| `mean:24h` | `[T - 24h, T)` |
+| `mean:-12h:+12h` | `[T - 12h, T + 12h)` |
 
-- `mean:24h` describes `T - 24h <= t < T`
-- `mean:-12h:+12h` describes `T - 12h <= t < T + 12h`
+Built-in methods are `mean`, `sum`, `min`, and `max`. Source cadence must divide
+the window exactly. Durations are integers with units `ns`, `us`, `ms`,
+`s/sec/second`, `m/min/minute`, `h/hr/hour`, `d/day`, `w/wk/week`, or
+`mo/mon/month`; plural names are accepted. Weeks are seven days, months fixed
+30-day windows—not calendar-relative months.
```
pzharrington: I may have led you slightly astray with this months suggestion. I was wondering if the 30-day window was an adopted convention, and I found that CF chooses:

- month defined exactly as year/12
- 30 day months only used in 360-day year calendars
- general warning on using years/months:
  - "We recommend that the unit year be used with caution. The Udunits package defines a year to be exactly 365.242198781 days (the interval between 2 successive passages of the sun through vernal equinox). It is not a calendar year. Udunits includes the following definitions for years: a common_year is 365 days, a leap_year is 366 days, a Julian_year is 365.25 days, and a Gregorian_year is 365.2425 days. For similar reasons the unit month, which is defined in `udunits.dat` to be exactly year/12, should also be used with caution.

Pandas 2.x outright refuses `M`, `Y`, and `y` timedeltas due to the unambiguitity...seems like months are ill-posed so maybe we drop them and just require user to choose some daily duration? Similar to how FuXi-S2S uses hours ranges to specify daily means

dev/spec/GRID_SPEC.md
pzharrington: Definitely support renaming to `y`, `x` instead of `hrrr_y`, `hrrr_x`

dev/examples/03_coordinate_signatures.py
```diff
@@ -0,0 +1,162 @@
...
+
+# %%
+atmosphere = coord_array(
+    ("batch", "variable", "lat", "lon"),
+    {"variable": ["u10m", "v10m"]},
+    dynamic=("batch",),
+    grid="fcn1",
+)
+precipitation = coord_array_like(
+    atmosphere, {"variable": ["tp"]}, statistics={"tp": "sum:6h"}
```
pzharrington: Is this is equivalent to doing `{"variable": ["tp:sum:6h"]}`?

dev/spec/GRID_SPEC.md
NickGeneva: I'll clear up the cropped grid spec a little yeah, this doesnt make any sense.

dev/spec/MODEL_CONTRACT_SPEC.md
NickGeneva: Ah I think its got confused since the two global models just have batch as the dynamic dim... this is incorrect.

dev/spec/MODEL_CONTRACT_SPEC.md
NickGeneva: Hmmm, you mean like models that could predict over any sort of input?

The best thing I can reasonably think of is bascally making that configured in the model... i.e. you need to manually tell the model your working in a given region to get the respective coords before starting the pipeline.

Without this theres no way for like fetch_data to know what to get because the model's domain coords becomes ambiguous and largely not informative right

Its why the input_coords and output_coords are not class methods

dev/spec/MODEL_CONTRACT_SPEC.md
NickGeneva: Yeah idk, good catch

dev/spec/TIME_STATISTICS_SPEC.md
NickGeneva: This is a good point i didnt think of, yeah lets drop it so we dont have to deal with this.

dev/examples/03_coordinate_signatures.py
NickGeneva: Yep! This was the original method I had implemented but maye i should remove it in favor of just the variable naming method so its just standard

NickGeneva: I'm going to remove this feature for now from te coords methods... no need to over complicate for something thats not needed. Triggering via the variable names is sufficient.

NickGeneva: removed

dev/spec/TIME_STATISTICS_SPEC.md
NickGeneva: Removed.

dev/spec/MODEL_CONTRACT_SPEC.md
NickGeneva: fixed

dev/spec/TIME_STATISTICS_SPEC.md
NickGeneva: Now just supports for seconds (plus sub seconds), minutes, hours, days and weeks. No month or year

dev/spec/MODEL_CONTRACT_SPEC.md
NickGeneva: corrected!

dev/spec/MODEL_CONTRACT_SPEC.md
pzharrington: Hmmm...yeah I think "configure then run" is the only pathway I can really see for this, i.e. a flexible multi-crop model has to be just pinned/set to specific crop coords ahead of time so `input_coords` is not changing mid-run (and so data fetch can happen etc). Probably same thing with variables. It's more of a question on what the model contract should add so that the `set_domain` or whatever method used to configure the coords in these kinds of models is standardized

earth2studio/models/px/fcn.py
```diff
@@ -116,102 +120,78 @@ def __init__(
...
-            {
-                "batch": np.empty(0),
-                "lead_time": np.array([np.timedelta64(6, "h")]),
-                "variable": np.array(VARIABLES),
-                "lat": np.linspace(90, -90, 720, endpoint=False),
-                "lon": np.linspace(0, 360, 1440, endpoint=False),
-            }
+            },
+            dynamic=("batch",),
+            grid="fcn1",
```
pzharrington: Total nit but I don't love the `fcn1` alias...e.g. also Aurora (and maybe some other models?) use this 720x1440 grid. Imo the base name is more self-documenting in that it's saying south pole excluded

earth2studio/utils/coords.py
```diff
@@ -14,22 +14,395 @@
...
+        handshake_dataarray(array, signature)
+
+
+def statistics_from_metadata(array: xr.DataArray | None) -> dict[str, str]:
+    """Return compact temporal-statistic modifiers from a signature."""
+    if array is None:
+        return {}
+    return {
+        variable: details["modifier"]
+        for variable, details in array.attrs.get(E2S_STATISTICS, {}).items()
```
pzharrington: Just noting it down since it may change pending next weeks discussions, but aren't we outright dropping the modifier dict approach in favor of putting this metadata in the variable name itself?

earth2studio/utils/coords.py
```diff
@@ -14,22 +14,395 @@
...
+            if dimension in sizes and sizes[dimension] != size:
+                raise ValueError(f"Coordinates disagree on size of '{dimension}'")
+            sizes[dimension] = int(size)
+    return sizes
+
+
+def coord_array(
+    dims: Sequence[Hashable],
+    coords: Mapping[Hashable, Any] | None = None,
+    *,
```
pzharrington: Variable number of args?

earth2studio/utils/coords.py
NickGeneva: Yeah this can get removed, its not used any where

earth2studio/utils/coords.py
```diff
@@ -14,22 +14,395 @@
...
+    dynamic: Sequence[Hashable] = (),
+    sizes: Mapping[Hashable, int] | None = None,
+    grid: str | GridDefinition | None = None,
+    grid_dims: Mapping[str, str] | None = None,
+    dtype: DTypeLike = np.float32,
+    name: Hashable | None = None,
+    attrs: Mapping[Hashable, Any] | None = None,
+) -> CoordinateSystem:
+    """Create an allocation-free Earth2Studio coordinate signature.
```
pzharrington: "allocation-free" is maybe a bit over-succinct here. For this docstring I'd just do something like "Create a coordinate signature from input dimensions and other optional information. The returned `CoordinateSystem`'s backing array stores only shape and dtype but does not allocate memory for field values." or similar, 2nd sentence only being needed if you want to explicitly note the allocation-free-ness

earth2studio/utils/coords.py
NickGeneva: Nope, should be removed

earth2studio/models/px/fcn.py
NickGeneva: Yeah good point, better name thats like about south pole excluding would be better

earth2studio/utils/coords.py
NickGeneva: Makes sense, good call

pzharrington (approved):

PR #1178, opened 2026-09-21
eval: score one online rollout against several truth sources

### LINKED ISSUE #1177: 🚀[FEA]: score one online rollout against several truth sources
## Is this a new feature, an improvement, or a change to existing functionality?

New Feature

## How would you describe the priority of this feature request

Medium

## Please provide a clear description of problem you would like to solve.

Comparing a forecast against more than one truth dataset is a routine ask — our
own ERA5 archive, IFS HRES, and self-verification against our own analysis — and
today the eval recipe gives no good way to do it.

Online scoring reduces each rollout to sufficient statistics as it runs, against
exactly one verification source. So to score against N truths you either:

- **run the rollout N times**, paying N x the model cost for forecasts that are
  bit-identical, or
- **set `output.retain=all`** and score offline N times, which means
  materializing `forecast.zarr`.

Neither works at campaign scale. A 16-member, 118-IC, 104-channel run keeps
`output.retain=none` precisely because the raw forecast is enormous, and the
model cost is the whole budget — doubling it to add a second reference is not a
trade anyone would make.

The forecast is already in memory at every lead step. Nothing about scoring it
against a second truth field requires re-running the model.

### What we would like

A `scoring.online.verification_sources` mapping of reference name -> DataSource,
each name getting its own stats/scores store (`stats__era5.zarr`,
`scores__ifs.zarr`), with the store schema otherwise unchanged. Left unset, a
campaign resolves to a single unnamed reference and writes exactly the filenames
it writes today.

The cost should be N truth-field reads and N passes through the existing
per-step accumulator — never a second rollout. Every statistic is built from the
residual `f - y`, so none of the per-step collectives can be shared between
references: N scales the ensemble-moment and rank-count reduces, one field each,
and the fair-CRPS member exchange, which ships M fields and therefore dominates.

There is an obvious-looking optimisation here that does not work, and it is worth
recording so nobody tries it. The per-reference moments could in principle be
derived from `sum(f)` and `sum(f^2)`, which contain no truth field and so could
be reduced once and shared by every reference. The problem is precision: for
z500 the values are ~5.4e4 and the ensemble variance is ~1e4, so recovering the
variance means subtracting two numbers near 4.7e10 that differ by only ~1e4. In
float32 the rounding step at that magnitude is already the size of the answer —
measured 7% error, against ~1e-7 for the residual form used today, which works
with numbers ~100 instead of ~54000.

aayushg55:
# Earth2Studio Pull Request

## Description

Adds `scoring.online.verification_sources`, a mapping of reference name ->
`DataSource`, so one online rollout can be scored against several truth datasets
in a single pass. Previously the only ways to compare against N truths were to
run the rollout N times or to retain `forecast.zarr` and score offline, neither
of which is affordable at campaign scale where `output.retain=none`.

Each named reference gets its own physical store, the name inserted before the
extension: `stats.zarr` -> `stats__era5.zarr`, `scores.zarr` ->
`scores__ifs.zarr`. That is a reference-per-store split rather than a new axis on
the existing schema, so `stats_array_groups` and `add_stats_arrays` are
unmodified.

Left unset the knob resolves to a single unnamed reference, keyed `None`, which
maps back to the plain filenames, so an existing campaign is unchanged.
`named_store_filename` is the only place filenames are derived — there is no
separate single-reference branch to drift. Keying the unnamed case `None` rather
than a sentinel string means it cannot collide with a user-chosen name: YAML
mappings require string keys, so the unnamed and named cases are mutually
exclusive by construction.

Reference names are validated against `^[A-Za-z0-9][A-Za-z0-9_-]*$` before use.
They are interpolated into a store filename that `OutputManager` joins onto
`output.path`, so an unvalidated name such as `../../x` would write outside the
output tree.

`MultiRefScorer` is a drop-in for `OnlineScorer`: the pipeline only ever calls
`begin_item` / `update` / `finish_item`, so the rollout loop and every per-step
model call are untouched. Each child keeps its own `FieldCache`, its own
accumulator buffers and its own store, and the forecast chunk is handed to every
child unchanged rather than copied per reference.

Cost is N truth reads and N accumulator passes per lead step, never a second
rollout. Every statistic is built from the residual `f - y`, so no per-step
collective is shareable between references: N scales the ensemble-moment and
rank-count reduces (one field each) and the fair-CRPS member exchange (M fields,
so it dominates). Reducing reference-independent raw-forecast moments would be
shareable but loses ~7% of the variance in float32, since `sum(f^2)` dwarfs the
spread it is differenced from. The config block documents this so the cost model
is predictable and the unsafe optimisation is not attempted later.

### Also in here

`resolve_verification_sources` carried an unreachable `raise ValueError` for an
empty `verification_sources`. The earlier `if not named:` caught `None` and `{}`
alike and returned the default, so a config with `verification_sources: {}` —
exactly the mistake the message described — silently scored a single reference.
Unset and set-but-empty (`{}`) are now distinguished, and the dead branch is
gone.

closes #1177 

## Checklist


recipes/eval/cfg/default.yaml
```diff
@@ -242,9 +242,53 @@ scoring:
...
+        # verification.zarr, so any additional reference is yours to stage.
+        #
+        # verification_sources:
+        #     era5:
+        #         _target_: src.data.PredownloadedSource
+        #         path: ${output.path}/verification.zarr
+        #     ifs:
+        #         _target_: src.data.PredownloadedSource
+        #         path: /data/ifs_analysis.zarr   # staged by you, see below
+        #     self_analysis: null   # reuse the pipeline's default resolution
```
Dibyajyoti-Chakraborty, line 289: `PredownloadedSource` takes `store_path`. Uncommenting this block will fail.

recipes/eval/test/test_online.py
```diff
@@ -1062,6 +1065,92 @@ def test_crps_disabled_and_deterministic_runs_skip_it(self, tmp_path):
...
+        self.finished = []
+
+    def begin_item(self, item):
+        self.begun.append(item)
+
+    def update(self, x, coords):
+        self.updates.append(x)
+
+    def finish_item(self, item):
+        self.finished.append(item)
```
Dibyajyoti-Chakraborty, line 1151: One issue might be that the marker is written by the first child before the others have flushed. A crash between the individual child jobs leaves later references NaN for an IC that resumes, then skips it. One solution I can think of is to update this MultiRefScorer.finish_item to account for all children jobs to complete.

Dibyajyoti-Chakraborty (commented):
This is a needed feature. Overall the PR looks feasible. Only some issues related to job failure cases, and resuming, needed to be thought of. I put some comments inline too. Also update the changelog

